PR #7, opened 2019-05-03
Nominations for agent and collector

SergeyKanzhelev:
@pjanotti nominating you as maintainer for agent/collector. Let me know if there are problems with this.

Note, this is nomination, not a complete list.

### DIFF
--- a/community-members.md
+++ b/community-members.md
@@ -68,7 +68,8 @@ Repo: TBD
 
 Maintainers:
 
-- [Paulo Janotti](https://github.com/pjanotti)
+- [Bogdan Drutu](https://github.com/BogdanDrutu), Google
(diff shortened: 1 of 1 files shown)


community-members.md
```diff
@@ -64,6 +64,12 @@ Maintainers:
 
 ## Agent and Collector
 
+Repo: TBD
+
+Maintainers:
+
+- [Paulo Janotti](https://github.com/pjanotti)
```
bogdandrutu: I am also very involved in this, so you can add me as well for this. Also we are going to use this inside Google (very soon) so it is good to have me as maintainer.

pjanotti:
np at all, thanks @bogdandrutu and @SergeyKanzhelev

PR #32, opened 2019-05-10
Dotnet sig sync up proposal.

SergeyKanzhelev:
Let's see if we can do 9 AM Tuesdays. We have couple people from MS joining as well

CC: @gregkalapos @austinlparker

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ We organize the community into Special Interest Groups (SIGs) in order to improv
 
 ### .NET SDK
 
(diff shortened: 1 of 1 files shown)


austinlparker:
Works for me!

gregkalapos:
9 AM PST is ok to me in general, but next week it doesn't work unfortunately (whole week company event). Is there any other option? - don't wanna push, just asking, since this would be the first one, which I'd especially like to attend.

SergeyKanzhelev:
@gregkalapos I'll make sure to publish detailed notes. Also happy to chat any time

PR #77, opened 2019-05-29
Add Python SIG info to README

c24t:
Add the new python SIG calendar. cc @tedsuo

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,9 @@ Conference ID: 214 680 28#
 
 ### Python SDK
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -47,6 +47,9 @@ Conference ID: 214 680 28#
 
 ### Python SDK
 
+The Python SIG meets weekly on [Thursday 9 AM PDT](https://calendar.google.com/calendar/embed?src=google.com_346fplg7bemb44783pege5amj4%40group.calendar.google.com&ctz=America%2FLos_Angeles). This meeting is subject to change depending on contributors' availability, check the calendar for specific dates.
```
c24t: I followed @SergeyKanzhelev's example here and created a separate public calendar for the python client, but we may just want to use the OT public calendar for everything.

SergeyKanzhelev: separate calendars may be easier for people who are only interested in couple of SIGs. Combining calendars is easy for individual.

I wonder if Google Calendar allows meta-calendars which simply combine a set of relevant calendars together

PR #79, opened 2019-05-29
Added bai as approver for Ruby

SergeyKanzhelev:
@bai, what's your company affiliation?


bai:
That'd be Shopify, thanks.

PR #81, opened 2019-05-30
Specify Proto maintainers and approvers

songy23:


community-members.md
```diff
@@ -141,6 +141,14 @@ Maintainers:
 Maintainers:
 - [Francis Bogsanyi](https://github.com/fbogsany), Shopify
 
+## Proto
```
SergeyKanzhelev: I'd suggest we keep the same group for proto as we do for specs. 

@bogdandrutu would you agree?

songy23:
SG, filed https://github.com/open-telemetry/community/issues/82.

PR #85, opened 2019-05-31
governance charter document

SergeyKanzhelev:
Authored and reviewed by @bhs @bogdandrutu @SarahNovotny @SergeyKanzhelev @yurishkuro

### DIFF
--- a/governance-charter.md
+++ b/governance-charter.md
@@ -1,0 +1,234 @@
+# OpenTelemetry Governance Committee Charter
+
+Last update: 5/31/2019
+
+## Overview
+
+This document describes the bootstrapping process for the OpenTelemetry
+Governance Committee. It describes the formation of the initial governance
+committee and its core responsibilities. This includes its governance processes
+and processes to reform itself as necessary.
+
+## Goals
+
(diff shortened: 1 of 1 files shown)


governance-charter.md
```diff
@@ -0,0 +1,234 @@
...
+[CIVS](http://civs.cs.cornell.edu/). The top vote getters will be elected to the
+respective positions. 
+
+### Maximal representation
+
+To encourage diversity there will be a maximum of one-third representation on
+the Steering Committee from any one company at any time. If the outcomes of an
+election result in greater than 1/3 representation (or maximum of two, whichever
+is greater), the lowest vote getters from any particular company will be removed
+until representation on the committee is less than one-third.
```
tigrannajaryan: It is also worth saying who gets the empty seat after removal (I am guessing the person with the next highest vote that meets maximal representation rule, but would be good to mention that explicitly).

SergeyKanzhelev: there may be a need for a special election.

SergeyKanzhelev: yes, I think this is assumed

PR #88, opened 2019-06-03
Just mention couple of related efforts

SergeyKanzhelev:


SergeyKanzhelev:
@tsloughter is the group still in proposal? Do you have a link to your meetings (if any) - may be useful to include it here?

tsloughter:
@SergeyKanzhelev the group is active. I'll create a calendar for the meetings in a bit and that can be listed.

Since there is a SIG per language though it might make sense to have an Erlang/Elixir one. Would make it easier to have listed a couple contributors like is done today for us on the website.

PR #90, opened 2019-06-04
Added service tests repository

SergeyKanzhelev:
See discussion here: https://github.com/open-telemetry/opentelemetry-service/issues/8

Keeping testbed separate from service as it will have it's own milestones and potentially can be used for other SDKs as well in the future


tigrannajaryan:
@bogdandrutu please review / approve if you are ok.

bogdandrutu:
I think we decided to not go this path.

PR #149, opened 2019-07-11
Link to shared calendar for js SIG

mayurkale22:
Same as https://github.com/open-telemetry/community/pull/147


SergeyKanzhelev:
@mayurkale22 should SIG be renamed as "JS" instead of "node" in public calendar?

mayurkale22:
> @mayurkale22 should SIG be renamed as "JS" instead of "node" in public calendar?

Sure, I will do that.

PR #151, opened 2019-07-18
Change Service SIG to 9am PT

flands:


tigrannajaryan:
@bhs yes, this was agreed in yesterday's SIG meeting.

PR #156, opened 2019-08-07
Add Technical Committee Charter

yurishkuro:


tech-committee-charter.md
```diff
@@ -0,0 +1,88 @@
...
+
+The TC will establish and maintain a development process for OpenTelemetry. The development process will establish guidelines for how the developers and community will operate.
+
+The TC is additionally responsible for organizing the project structure, including possibly the creation and alignment of Special Interest Groups (SIGs), and/or sub-projects. Each SIG or sub-project must have a well-defined scope and must work within that scope.
+
+The development process will include a process for the TC to oversee and approve changes in the lifecycle of a Project, which will include consideration of the following criteria:
+
+* Cleanliness of code base
+* Ample and diverse Members, Approvers and Maintainers to assure the vitality of the project
+Stability (e.g. presence of test suites, stable APIs and use of an appropriate source-code control system)
```
SergeyKanzhelev: missing star for the list 
```suggestion
* Stability (e.g. presence of test suites, stable APIs and use of an appropriate source-code control system)
```

yurishkuro: thx, fixed

tech-committee-charter.md
```diff
@@ -0,0 +1,88 @@
...
+
+This OpenTelemetry Technical Committee Charter reflects the scope and expectations of the Technical Committee ("TC") with the maintainers, the consumers, and the ecosystem of the project. The charter will not be perfect, and so has a simple amendment process – a TC member proposes changes to be discussed and voted on by the full TC.
+
+## Establishment of the Technical Committee
+
+TC memberships are not time-limited. There is no maximum size of the TC. The size is expected to vary in order to ensure adequate coverage of important areas of expertise, balanced with the ability to make decisions efficiently. The TC must have at least four members and is recommended to have an odd number of members for tie-breaking purposes.
```
mtwo: Pasting discussion from the monthly governance meeting: we should consider making these time limited

yurishkuro: are there arguments in favor fo time limited?  The downside is that then TC would need to have formal election process similar to GC

tedsuo:
Summarizing the proposal, please let me know if this is accurate. 

* The TC has final authority and responsibility for the specification and technical roadmap. 
* Decisions are made by consensus voting mechanisms chosen by the TC. 
* The TC may be any size greater than three. 
* Existing TC members choose new members. 
* No more than 1/4 of the TC may be employed by the same company. 

There are more details in the proposal; are there any top level bullet points which I missed?

SergeyKanzhelev:
@tedsuo, from this angle, one additional item here is also a note about SIG maintainers and their relationship to TC.

bhs:
One thing that occurred to me: in the current formulation, it will be awkward to add a single TC member since that would move the total TC headcount from odd to even. We could either assume the convention of adding pairs of people at a time?

I've thought more about having "real" elections for the TC. There is a clear tradeoff between...
1. project velocity (both in terms of the overhead of running elections, and also the potential hit to velocity if/when new TC members arrive with their different technical vision and direction)
1. checks and balances on the integrity and quality of the TC

One idea I wanted to float: IMO we really want to bias for _stability of vision_ for at least the rest of 2019 and probably much of 2020. How would we feel about leaving formal TC elections out of the TC charter for now, but formally adding something at the GC level that the issue _must_ be revisited in 2020 Q1? At that point, OpenTelemetry should be GA'd in most languages and the pendulum could swing back towards "checks and balances".

My two cents.

tech-committee-charter.md
bhs: The governance committee met today and discussed this topic (among others).

Briefly, we believe that there is a tradeoff between the stability and velocity of the project and robust checks and balances on TC members (including term limits). **Given the current project stage** (read: "very early and highly dynamic / chaotic"), we believe that OpenTelemetry's goals will be best-served by leaving term limits out of the TC charter *for now*, but we also recognize that this may very well shift in the future as the technical landscape around OpenTelemetry stabilizes.

As such, our resolution is to leave this section as-is for now (i.e., "no term limits"), **but** to intentionally reconsider this topic no later than **July 2020**; if the balance has shifted towards "checks and balances" and away from "stability" at that time, we may very well introduce term limits, a formal TC election process, and so forth.

bhs:
Well, in my mind, then, this is ready to merge; @bogdandrutu, please object soon if the odd-numbered consideration feels important.

PR #161, opened 2019-08-20
Fix some un-updated terminology

bhs:
Namely, "Steering Committee" is the "Governance Committee" and the "TSC" is the "TC" in OpenTelemetry.


bhs:
(Just some housecleaning here)

bhs:
(Merging as this is presumably uncontroversial)

PR #163, opened 2019-08-21
Propose COMMUNICATION.md

bhs:
From the document itself:

>OpenTelemetry has a Code of Conduct which primarily pertains to "bannable" offenses: things like harassment, personal attacks, and so on. Certainly OpenTelemetry does not tolerate any such behavior.
>
>And while we all deserve to collaborate without fear of overt harassment and personal attacks, that is a very low bar! This document is more aspirational and has fewer "teeth" (i.e., it's not about "banning" people), but it's also very important: it describes how we wish to communicate with each other and how we aspire to create an environment that cultivates trust, even in the face of inevitable technical and organizational disagreement.

.....

Personally, I think the code of conduct is both super important and also pretty useless on a day-to-day level. I want something separate that we can reference in our conversations to promote optimal behavior and, potentially, to call out "suboptimal" behavior as needed.

Curious for other thoughts (on whether this doc should even exist, and what it should contain if we like the idea).

### DIFF
--- a/COMMUNICATION.md
+++ b/COMMUNICATION.md
@@ -1,0 +1,21 @@
+# Communication within the OpenTelemetry Project
+
+## Document purpose
+
+OpenTelemetry has a [Code of Conduct](https://github.com/open-telemetry/community/blob/master/code-of-conduct.md) which primarily pertains to "bannable" offenses: things like harassment, personal attacks, and so on. Certainly OpenTelemetry does not tolerate any such behavior.
+
+And while we all deserve to collaborate without fear of overt harassment and personal attacks, that is a very low bar! This document is more aspirational and has fewer "teeth" (i.e., it's not about "banning" people), but it's also very important: it describes **how we wish to communicate with each other** and **how we aspire to create an environment that cultivates trust**, even in the face of inevitable technical and organizational disagreement.
+
+Finally, this is **a living document** and should be adapted over time as the OpenTelemetry project evolves. At any time, there are certainly errors of omissions or even simple clarifications we can make to the wording here, and any and all are welcome to propose such changes.
+
+Now, without further ado:
+
+## Communication guidelines
+
+What follows are an unordered list of principles that guide our communication with each other – both active community members as well as those who may just be "passing through" to file an issue or suggest an improvement.
+
+* **Critique ideas, not people:** It's fine and expected to disagree about technology and project direction. Even when those disagreements are significant or frustrating, though, it's not fine to criticize the participants. (In fact, it's usually a sign that there's no strong technical argument being made)
+* **Be welcoming:** By nature, OpenTelemetry touches a lot of other projects and will thus have a larger-than-normal ratio of casual committers and passers-by. For this reason, it's especially important to be welcoming to newcomers. This means thanking them for their interest and contributions, and being patient with their initial misunderstandings. In fact, their points of confusion are a valuable signal for documentation gaps.
+* **Try to keep conversations in the open:** Where possible, try to have conversations in async formats like GitHub issues, PRs, and emails. If resolution requires a call or videoconference, try to make a recording and notes available to keep our community accessible to other timezones.
+* **Strive for consensus, but don't require it:** Of course it's great when we can get everyone to agree. And certainly everyone should be – and should feel – heard. But once debate has died down and a decision must be made, it's best for the project and community to commit to that decision and make the best of it.
+


COMMUNICATION.md
```diff
@@ -0,0 +1,21 @@
...
+* **Be welcoming:** By nature, OpenTelemetry touches a lot of other projects and will thus have a larger-than-normal ratio of casual committers and passers-by. For this reason, it's especially important to be welcoming to newcomers. This means thanking them for their interest and contributions, and being patient with their initial misunderstandings. In fact, their points of confusion are a valuable signal for documentation gaps.
+* **Try to keep conversations in the open:** Where possible, try to have conversations in async formats like GitHub issues, PRs, and emails. If resolution requires a call or videoconference, try to make a recording and notes available to keep our community accessible to other timezones.
+* **Strive for consensus, but don't require it:** Of course it's great when we can get everyone to agree. And certainly everyone should be – and should feel – heard. But once debate has died down and a decision must be made, it's best for the project and community to commit to that decision and make the best of it.
```
bogdandrutu: To be honest this is very subjective. If there are comments/concerns not answered it is hard to say that the conversation died. By intentionally ignoring someone's comments does not mean that you strive for consensus or the communication died.

As an example my RFC about sampling, Yuri had important concerns/comments which I cannot ignore for 2 days and say the conversation died, even though half of the approvers approved that PR which I can use in my favor to say that I strive for consensus.

This probably needs a bit more clarification, also probably does not belong in the communication document, more like contribution guideline.

COMMUNICATION.md
```diff
@@ -0,0 +1,21 @@
...
+
+Finally, this is **a living document** and should be adapted over time as the OpenTelemetry project evolves. At any time, there are certainly errors of omissions or even simple clarifications we can make to the wording here, and any and all are welcome to propose such changes.
+
+Now, without further ado:
+
+## Communication guidelines
+
+What follows are an unordered list of principles that guide our communication with each other – both active community members as well as those who may just be "passing through" to file an issue or suggest an improvement.
+
+* **Critique ideas, not people:** It's fine and expected to disagree about technology and project direction. Even when those disagreements are significant or frustrating, though, it's not fine to criticize the participants. (In fact, it's usually a sign that there's no strong technical argument being made)
```
jpkrohling: I feel like this is a good place to reference this blog post: https://developers.redhat.com/blog/2019/07/08/10-tips-for-reviewing-code-you-dont-like/ . I know your point isn't about code reviews, but most of the arguments in that blog post apply to general communication as well.

COMMUNICATION.md
bhs: That's fair. I agree this is too murky as worded.

I'm looking for "a sign on the wall" that can both help (a) when conversations have stalled and a decision simply has to be made, and also (b) when a single individual has a valid concern that should be resolved, even when everyone else seems fine with the topic/proposal/issue.

I'll think about other ways to frame this, but if there are suggestions from others reading this thread, I'm totally open.

bhs: (I have fleshed things out a bit further)

COMMUNICATION.md
bhs: Good point, I have added a link. I looked around for other articles, but couldn't find one that's better.

COMMUNICATION.md
yurishkuro: I think there are two different aspects here. One is about communication style / conduct, which is the purpose of this document, where we want to listen to dissenting opinions, understand and address their concerns, rather than ignoring people or trying to shut them up with dismissive statements. And a separate question of what to do when there are unresolvable disagreements. The latter is not a question of communication style, it's a matter of policy and process, namely that if consensus cannot be achieved through discussion it will be resolved by the TC format vote.

bhs: @yurishkuro that's a great point. I have reworded this again.

COMMUNICATION.md
```diff
@@ -0,0 +1,23 @@
...
+
+Finally, this is **a living document** and should be adapted over time as the OpenTelemetry project evolves. At any time, there are certainly errors of omissions or even simple clarifications we can make to the wording here, and any and all are welcome to propose such changes.
+
+Now, without further ado:
+
+## Communication guidelines
+
+What follows are an unordered list of principles that guide our communication with each other – both active community members as well as those who may just be "passing through" to file an issue or suggest an improvement.
+
+* **Critique ideas, not people:** It's fine and expected to disagree about technology and project direction. Even when those disagreements are significant or frustrating, though, it's not fine to criticize the participants. In fact, it's usually a sign that there's no strong technical argument being made. (More specifically, see [this post](https://developers.redhat.com/blog/2019/07/08/10-tips-for-reviewing-code-you-dont-like/) for tips about code reviews in particular)
```
ccaraman: Would a stronger statement like `it isn't acceptable to ...` communicate our philosophy more clearly?

COMMUNICATION.md
```diff
@@ -0,0 +1,23 @@
...
+* **Be welcoming:** By nature, OpenTelemetry touches a lot of other projects and will thus have a larger-than-normal ratio of casual committers and passers-by. For this reason, it's especially important to be welcoming to newcomers. This means thanking them for their interest and contributions, and being patient with their initial misunderstandings. In fact, their points of confusion are a valuable signal for documentation gaps.
```
ccaraman: nit: for -> of

COMMUNICATION.md
```diff
@@ -0,0 +1,23 @@
...
+* **Try to keep conversations in the open:** Where possible, try to have conversations in async formats like GitHub issues, PRs, and emails. If resolution requires a call or videoconference, try to make a recording and notes available and discoverable to keep our community as accessible as possible across distant timezones.
+* **Strive for consensus, but don't require it:** Of course it's great when we can get everyone to agree. And certainly everyone should be – and should feel – heard. As a rule of thumb, if there is a lone dissenter (or three) in any given discussion, their points should have been clearly addressed: not just with a simplistic "Sorry, I disagree," but with a principled counterargument. (Note that the way we ultimately make decisions in spite of consensus is not just a communication issue, but a matter of policy and process; as such, that is out of scope for this document)
```
tedsuo:
```suggestion
* **Strive for consensus, but don't require it:** Of course it's great when we can get everyone to agree. And certainly everyone should be – and should feel – heard. As a rule of thumb, if there is a lone dissenter (or three) in any given discussion, their points should be clearly addressed: not just with a simplistic "Sorry, I disagree," but with a principled counterargument. (Note that the way we ultimately make decisions in spite of consensus is not just a communication issue, but a matter of policy and process; as such, that is out of scope for this document)
```

tedsuo: The parenthetical here feels a little contorted. Is the intent to say "Note that project governance and the rules for final decision making are covered elsewhere"?

tedsuo: "should have been clearly addressed" -> "should be clearly addressed"

tedsuo:
Thanks @bhs. I feel this document describes the conversational tone and approach to discussion which has helped this observability community come together.

COMMUNICATION.md
bhs: great suggestion, done

COMMUNICATION.md
bhs: I did something different, but agreed it was bad before :)

COMMUNICATION.md
bhs: done

bhs: I did something...

bhs:
Any further comments or concerns? If not, I will merge sometime next week...

COMMUNICATION.md
tedsuo: LGTM

thomashchan1:
This LGTM as well. Thanks for drafting and working this.

PR #175, opened 2019-09-07
add Fred Hebert as an approver for Erlang lib

tsloughter:
The opentelemetry-erlang repo has only in the past month started having PRs for review, so the 3 months of reviewing requirement can not be met. But @ferd has given substantial reviews on our most recent PRs that are defining the structure of the project https://github.com/open-telemetry/opentelemetry-erlang/pulls

Fred in general is a good fit as an approver due to his experience with Erlang, both developing and teaching, plus writing 3 books on the subject:

- https://learnyousomeerlang.com/
- http://www.erlang-in-anger.com/
- http://propertesting.com/

### DIFF
--- a/community-members.md
+++ b/community-members.md
@@ -147,6 +147,7 @@ Maintainers:
 Repo: [open-telemetry/opentelemetry-erlang](https://github.com/open-telemetry/opentelemetry-erlang)
 
 Approvers:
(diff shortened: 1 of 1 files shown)


bogdandrutu:
While we are still bootstrapping the erlang sig I think it is fine to bring in members with experience like this. I support this request.

tsloughter:
Anything else we need to do to get this through?

PR #192, opened 2019-09-26
Removed timezone from the link

SergeyKanzhelev:
@trask , if you want to re-use the same zoom room for the meeting - please make it half an hour so it will not be back to back to the next one also using the same room. Alternatively, perhaps @tedsuo can change zoom room to another one for GoLang SIG

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,7 +148,7 @@ You can also join us on [gitter](https://gitter.im/open-telemetry/opentelemetry-
 
 #### Java Auto-Instrumentation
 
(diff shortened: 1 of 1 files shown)


trask:
Can we switch rooms so we don't have to switch times?

Also, may want to remove time zone from all the other links on this page?

Thanks!

SergeyKanzhelev:
> Can we switch rooms so we don't have to switch times?

I switched you to blueberry https://github.com/open-telemetry/community/blob/master/docs/how-to-handle-public-calendar.md

> Also, may want to remove time zone from all the other links on this page?

Yes, probably. I didn't check.

trask:
Thanks

PR #197, opened 2019-09-26
Add Krzesimir Nowak as Go approver

jmacd:
Krzesimir Nowak has made several substantial contributions to the OTel-Go repo, including the OpenTracing bridge. 

https://github.com/open-telemetry/opentelemetry-go/pull/98
https://github.com/opentracing/opentracing-go/pull/220

@krnowak

### DIFF
--- a/community-members.md
+++ b/community-members.md
@@ -78,6 +78,7 @@ Approvers:
 - [Paulo Janotti](https://github.com/pjanotti), Omnition
 - [Steven Karis](https://github.com/sjkaris), Omnition
 - [Ted Young](https://github.com/tedsuo), LightStep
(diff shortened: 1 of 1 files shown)


bhs:
I think by the "letter of the law" (per [these requirements](https://github.com/open-telemetry/community/blob/master/community-membership.md#requirements-1)), Krzesimir hasn't made enough contributions to be an "Approver"... is the idea to make an exception? If so, why? Mainly because we simply need more Approvers in Go? (Otherwise I guess we start with "Maintainer", yes?)

To be clear: I am really impressed with Krzesimir's work, and these sound like substantial contributions – just trying to respect the processes we have, etc.

PR #202, opened 2019-10-01
Update Link & Phone numbers for Spec SIG

Oberon00:
We aren't using MS Teams anymore.


Oberon00:
Although maybe instead of updating, we should remove the Zoom link altogether because they are already on the calendar and on the meeting notes. It seems that with these three places, one is always behind (often we got split meetings when the link changes)

bhs:
(Given @Oberon00's latest comment, I'll hold off on merging... I like the idea of just referring to the cal event for meeting instructions, btw)

SergeyKanzhelev:
@bhs, this can be done as a separate PR. Merging =)

PR #218, opened 2019-10-14
Remove iredelmeier as a JS approver

### LINKED ISSUE #217: REQUEST: Remove me as a JS approver

iredelmeier:
[Closes #217]

### DIFF
--- a/community-members.md
+++ b/community-members.md
@@ -91,7 +91,6 @@ Repo: [open-telemetry/opentelemetry-js](https://github.com/open-telemetry/opente
 Approvers:
 
 - [Yuri Shkuro](https://github.com/yurishkuro), Uber
-- [Isobel Redelmeier](https://github.com/iredelmeier), LightStep
 - [Roch Devost](https://github.com/rochdev), DataDog
 - [Brandon Gonzalez](https://github.com/bg451), LightStep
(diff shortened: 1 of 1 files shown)


bogdandrutu:
Can you please elaborate a bit on the reasoning? Is this just because you no longer want to be involved in that project? Is because you don't feel you have enough time? Or because you cannot work with people from that SIG?

Also something like "personal decision" is fine (if you don't want to share details), but I try to understand if we can do something to help.

iredelmeier:
@bogdandrutu @SergeyKanzhelev I got grandfathered in (or something like that) when really I'm not actively contributing to the JS SDK. Definitely nothing to do with personal problems or other problems with the people!

bhs:
I agree that it would be great to learn anything @iredelmeier would like to share about how this SIG, this repo, and/or OTel in general can improve, but in the meantime I don't think there should be any requirement for explanation before merging a self-request to leave a SIG; so I will merge this.

(In parallel, @iredelmeier has provided her rationale, so I'm feeling even further emboldened ;)

bogdandrutu:
@bhs sorry if it was miss understand that it was a requirement, I was asking to learn how we can improve the experience for developers if there was a problem.

Thanks @iredelmeier for the clarification.

PR #277, opened 2019-12-19
Sync CODEOWNERS to governance committeee

lizthegrey:


CODEOWNERS
```diff
@@ -12,4 +12,4 @@
 #  https://help.github.com/en/articles/about-code-owners
 #
 
-*  @bhs @bogdandrutu @sarahnovotny @SergeyKanzhelev @yurishkuro
+*  @opentelemetry/admins
```
yurishkuro: is the admins list actually visible publicly?

image

lizthegrey: We should make it publicly visible IMO.

lizthegrey: @yurishkuro you have access to update the visibility settings.

yurishkuro: changed. I assume it's not case sensitive.

lizthegrey:
Hm, now the shield icon is gone from everyone. This doesn't seem right :(

PR #279, opened 2019-12-21
Fix EEF Observability group link

GregMefford:
I wanted to fix the link since it was broken, but figured it would also be nice to add a few words of description.


GregMefford:
Signed CLA after pushing the commit, so let's see if this comment updates it. 😅

GregMefford:
CLA signed.

PR #280, opened 2019-12-21
Fix role name inconsistency

GregMefford:
There isn't a definition for the `owner` role, but looking at the detailed requirements below, it looks like the intention was for this to be `approvers` instead.

### DIFF
--- a/community-membership.md
+++ b/community-membership.md
@@ -12,7 +12,7 @@ delegated. They can be reached via e-mail cncf-opentelemetry-governance@lists.cn
 
 | **Role**   | **Responsibilities**                                  | **Requirements**                                             | **Defined by**                                               |
 | ---------- | ----------------------------------------------------- | ------------------------------------------------------------ | ------------------------------------------------------------ |
-| member     | active contributor in the community.  reviewer of PRs | sponsored by 2 approvers. multiple contributions to the project. | OpenTelemetry GitHub org member.                             |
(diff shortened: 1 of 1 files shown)


lizthegrey:
Approvers or maintainers.

community-membership.md
```diff
@@ -12,7 +12,7 @@ delegated. They can be reached via e-mail cncf-opentelemetry-governance@lists.cn
 
 | **Role**   | **Responsibilities**                                  | **Requirements**                                             | **Defined by**                                               |
 | ---------- | ----------------------------------------------------- | ------------------------------------------------------------ | ------------------------------------------------------------ |
-| member     | active contributor in the community.  reviewer of PRs | sponsored by 2 owners. multiple contributions to the project. | OpenTelemetry GitHub org member.                             |
+| member     | active contributor in the community.  reviewer of PRs | sponsored by 2 approvers or maintainers. multiple contributions to the project. | OpenTelemetry GitHub org member.                             |
```
yurishkuro: Maintainers across many repos is already a sizeable group. Is there some friction why we need to expand it more, to approvers?

lizthegrey:
That's two approvals, merging.

community-membership.md
lizthegrey: Oh, race conditioned. Note that it's tidying up what the criteria _already_ said in https://github.com/GregMefford/community/blob/master/community-membership.md#member below, and summarizing it correctly, rather than changing the criteria.

PR #290, opened 2020-01-26
Incorporate an OTel-wide maturity matrix

bhs:
Per the header comment for the file under review here:

>   OpenTelemetry is a complex project with many moving parts. Thankfully, we strive to keep those parts loosely coupled where we can (this has been a goal since the very announcement of OpenTelemetry, per this blog post: https://medium.com/opentracing/merging-opentracing-and-opencensus-f0fe9c7ca6f0)
> 
>   This file describes the structure of the project and its sub-components, definitions of API and stability maturity levels, and finally records the current maturity of each component. We maintain this file as YAML in order to make it easier to automate scripts (e.g., for the website) that present the data to OpenTelemetry end-users.

This PR stems from the most recent technical *and* governance committee meetings. cc @mtwo who is doing a survey about blockers to a beta release in 5 languages for Kubecon EMEA.

All comments welcome, but especially those about the actual schema (formatting, repo location, naming, etc will be easy to change and I'm very flexible on those fronts!).

### DIFF
--- a/maturity-matrix.yaml
+++ b/maturity-matrix.yaml
@@ -39,9 +39,11 @@ productionMaturityLevels:
   # security issues.
   - unstable
   # "beta" components have been used successfully in production workloads at
-  # scale, but either due to the uncertainty introduced by active development
-  # or a small number of production deployments, they should still be used
-  # cautiously in new deployments. They do not have known security issues.
+  # scale: that is, the component is used in support of well-known product
+  # functionality at publicly-held companies. Nevertheless, either due to
+  # the uncertainty introduced by active development or a small number of
+  # production environments, they should still be used cautiously in new
+  # deployments. They do not have known security issues.
   - beta
   # "stable" components are, to the best of the author's knowledge, safe for
   # typical production use cases.
@@ -90,7 +92,7 @@ perLanguageComponents:
 # end-user programming language.
 standaloneComponents:
   collector:
-    api: beta
+    api: alpha
     production: beta
   operator:
     api: alpha
@@ -102,10 +104,12 @@ standaloneComponents:
 # repository). E.g., something like this:
 #
 # languages:
-#   logo:
-#     url: "https://raw.githubusercontent.com/open-telemetry/opentelemetry-logo/master/maturity.yaml"
+#   pascal:
+#     repoUrl: "https://github.com/open-telemetry/opentelemetry-pascal"
+#     maturityUrl: "https://raw.githubusercontent.com/open-telemetry/opentelemetry-pascal/master/maturity.yaml"
 languages:
   java:
+    repoUrl: "https://github.com/open-telemetry/opentelemetry-java"
     traceAPI:
       api: alpha
       production: beta
--- a/maturity-matrix.yaml
+++ b/maturity-matrix.yaml
@@ -1,0 +1,138 @@
+# OpenTelemetry is a complex project with many moving parts. Thankfully, we
+# strive to keep those parts loosely coupled where we can (this has been a goal
+# since the very announcement of OpenTelemetry, per this blog post:
+# https://medium.com/opentracing/merging-opentracing-and-opencensus-f0fe9c7ca6f0)
+#
+# This file describes the structure of the project and its sub-components,
+# definitions of API and stability maturity levels, and finally records the
+# current maturity of each component. We maintain this file as YAML in order to
+# make it easier to automate scripts (e.g., for the website) that present the
+# data to OpenTelemetry end-users.
+
+
+###############################################################################
+# MATURITY LEVELS
+#
+# OpenTelemetry components advertise their own maturity along two axes:
+# - PRODUCTION MATURITY: How reliable is this component from a
+#   production-readiness standpoint?
+# - API MATURITY: How stable is the component's API from a backwards- and
+#   forwards-compatibility standpoint)?
+#
+# In some cases, for instance, OpenTelemetry componants may be immature from an
+# API maturity standpoint, yet be nearly 100% safe from a production standpoint
+# – and vice versa. Similarly, end-users may have more tolerance for lost
+# development time or software stability depending on their situation, so we
+# strive to separate self-reported maturity along these two axes.
+#
+# Needless to say, representations here are best-effort, and there is no
+# substitute for a mature release process. I.e., don't blindly deploy "mature"
+# OpenTelemetry components (or any software, for that matter!).
+productionMaturityLevels:
+  # "unknown" maturity is just what it purports to be. It's reasonable to
+  # assume "unstable", but we separate "unknown" and "unstable" to distinguish
+  # between a lack of evidence about production maturity and actual evidence of
+  # immaturity.
+  - unknown
+  # "unstable" components are not recommended for production workloads. They
+  # may crash the process, introduce performance artifacts, or have known
+  # security issues.
+  - unstable
+  # "beta" components have been used successfully in production workloads at
+  # scale, but either due to the uncertainty introduced by active development
+  # or a small number of production deployments, they should still be used
+  # cautiously in new deployments. They do not have known security issues.
+  - beta
+  # "stable" components are, to the best of the author's knowledge, safe for
+  # typical production use cases.
+  - stable
+
+# All OpenTelemetry APIs follow semver conventions (i.e.,
+# backwards-incompatibly changes should bump the major version number).
+apiMaturityLevels:
+  # "alpha" maturity APIs can change in incompatible ways at any time.
+  - alpha
+  # "beta" maturity APIs should not introduce backwards-incompatible changes
+  # more than once every three months; and when those changes are introduced, the
+  # authors will make a best effort to provide compatibility bridges.
+  - beta
+  # "stable" maturity APIs should not introduce backwards-incompatible changes
+  # more than once every twelve months, and will make every effort to provide
+  # compatibility bridges if at all possible.
+  - stable
+
+
+###############################################################################
+# OPENTELEMETRY COMPONENTS
+#
+# Some OpenTelemetry components run within end-user applications and thus have
+# a footprint across many languages. Naturally, the production and API maturity
+# of these components varies from language to language. Other components run as
+# infrastructure in their own containers or VMs and thus have a single maturity
+# level regardless of the end-user's programming language.
+#
+# The maturity – or even the existence – of the "perLanguageComponents" varies
+# from language to language. These are components that integrate into the
+# actual end-user application runtime in production. See "languages" for the
+# actual maturity levels.
+perLanguageComponents:
+  - traceAPI
+  - metricsAPI
+  - contextAPI
+  - traceSDK
+  - metricsSDK
+  - contextSDK
+  - automaticInstrumentation
+
+# The maturity of the "standaloneComponents" is independent of any particular
+# end-user programming language.
+standaloneComponents:
+  collector:
+    api: beta
+    production: beta
+  operator:
+    api: alpha
+    production: beta
+
+# For each language, we either specify the perLanguageComponents maturity
+# inline, or we specify a single "url" that points to another YAML file with a
+# top-level perLanguageComponents key (presumably in the OpenTelemetry language
+# repository). E.g., something like this:
+#
+# languages:
+#   logo:
+#     url: "https://raw.githubusercontent.com/open-telemetry/opentelemetry-logo/master/maturity.yaml"
+languages:
+  java:
+    traceAPI:
+      api: alpha
+      production: beta
+    metricsAPI:
+      api: alpha
+      production: beta
+    contextAPI:
+      api: alpha
+      production: beta
+    traceSDK:
+      api: alpha
+      production: unstable
+    metricsSDK:
+      api: alpha
+      production: unstable
+    contextSDK:
+      api: alpha
+      production: unstable
+    automaticInstrumentation:
+      api: beta
+      production: unstable
+
+  # TODO once we've converged on the schema for all of this
+  js: ""
+  dotnet: ""
+  ruby: ""
+  erlang: ""
+  python: ""
+  go: ""
+  php: ""
+  cpp: ""
+  rust: ""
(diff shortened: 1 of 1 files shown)


lizthegrey:
Fixed some typoes, updated golang status, added "unimplemented" as a state in addition to "alpha".

we currently do _not_ release major versions on breaking changes while we're in alpha for golang, as a FYI. our goal is to release major/minor versions corresponding to spec major/minor versions rather than incrementing in unpredictable ways that slew from the spec major/minor versions.

maturity-matrix.yaml
```diff
@@ -47,9 +47,11 @@ productionMaturityLevels:
   # typical production use cases.
   - stable
 
-# All OpenTelemetry APIs follow semver conventions (i.e.,
-# backwards-incompatibly changes should bump the major version number).
+# All OpenTelemetry APIs follow semver conventions (i.e., after v1.x,
```
bhs: yes, thank you for this! that's actually what I'd meant to write, so it's wonderful to see your telepathy skills on display... :)

maturity-matrix.yaml
```diff
@@ -47,9 +47,11 @@ productionMaturityLevels:
...
+# backwards-incompatible changes should bump the major version number).
 apiMaturityLevels:
+  # "unimplemented" APIs do not exist yet for the component/language.
```
bhs: @lizthegrey WDYT about (also) adding something like "unplanned" or "undesired" or "notApplicable"? E.g., auto-instrumentation in some future `otel-c99` support.

tedsuo:
CC @austinlparker 

We’ve been trying to track this via github milestones, and automatically scrape that data to update the status page on the website. Would be great if we could continue to have an automated process for keeping the website up to date.

bhs:
@tedsuo @austinlparker yes, agree – I like the idea of using repo metadata (milestones, tags, whatever) to track status. And also agree about the website. My goal with this PR is primarily to debate "the schema" for what we're telling end-users about maturity. If we figure that out, I am personally flexible about the mechanics as long as they allow for automation and don't require some crazy amount of upfront work.

austinlparker:
Once we've got the schema sorted out then it should be pretty easy to add some tags or whatever to auto-populate the website.

lizthegrey:
Agree, the main change here is moving off of one-dimensional v0.1->v0.4 versioning to tagging each maturity stage per category and project.

maturity-matrix.yaml
```diff
@@ -0,0 +1,161 @@
...
+# actual end-user application runtime in production. See "languages" for the
+# actual maturity levels.
+perLanguageComponents:
+  - traceAPI
+  - metricsAPI
+  - contextAPI
+  - traceSDK
+  - metricsSDK
+  - contextSDK
+  - automaticInstrumentation
```
SergeyKanzhelev: should the integrations and exporters be also tracked? Zipking and Jaeger exporters, integrations like SpringBoot, etc.?

maturity-matrix.yaml
```diff
@@ -0,0 +1,161 @@
...
+
+# The maturity of the "standaloneComponents" is independent of any particular
+# end-user programming language.
+standaloneComponents:
+  collector:
+    api: beta
```
SergeyKanzhelev: there is always a "latest stable" and "current" versions. Should we make this spearation?

maturity-matrix.yaml
```diff
@@ -0,0 +1,161 @@
...
+  - beta
+  # "stable" components are, to the best of the author's knowledge, safe for
+  # typical production use cases.
+  - stable
+
+# All OpenTelemetry APIs follow semver conventions (i.e., after v1.x,
+# backwards-incompatible changes should bump the major version number).
+apiMaturityLevels:
+  # "unimplemented" APIs do not exist yet for the component/language.
+  - unimplemented
```
SergeyKanzhelev: should there also be "N/A". Do we expect auto-instrumentation for the GoLang for instance?

maturity-matrix.yaml
```diff
@@ -0,0 +1,161 @@
...
+      api: beta
+      production: beta
+    metricsSDK:
+      api: alpha
+      production: unstable
+    contextSDK:
+      api: unimplemented
+      production: unknown
+    automaticInstrumentation:
+      api: unimplemented
```
SergeyKanzhelev: if it's not planned - should it be simply ommitted?

maturity-matrix.yaml
austinlparker: IMO we should focus solely on SDK/API for trace/metric/context here and allow exporters and integrations to simply reference their stability or maturity against the lower level components.

maturity-matrix.yaml
```diff
@@ -0,0 +1,161 @@
...
+  # TODO once we've converged on the schema for all of this
+  js: ""
+  dotnet: ""
+  ruby: ""
+  erlang: ""
+  python: ""
+  golang:
+    traceAPI:
+      api: beta
+      production: beta
```
SergeyKanzhelev: I'd suggest to include the link to the issues query on GitHub showing what's left for this area. It may be an optional attribute

maturity-matrix.yaml
lizthegrey: Yes, auto-instrumentation for golang includes wrappers for gorillamux, sqlx, etc. -- so it's not _fully_ automatic but it's O(1) lines of code to add.

SergeyKanzhelev: Yeah. But than it falls under "integrations" bucket and we may want more than overall status? Perhaps those integrations can also contribute to the registry that @austinlparker is working on?

maturity-matrix.yaml
SergeyKanzhelev: some integrations and exporters are the baseline for the stability as you cannot use API/SDK without ability to export it. Similar for integrations. I'm not sure how automaticInstrumentation here is different than integrations?

maturity-matrix.yaml
austinlparker: Yeah, I think auto-instrumentation is just a type of integration in the registry.

maturity-matrix.yaml
austinlparker: Hm, that's a good point, although I don't know how to square it. Automatic instrumentation does seem to require integrations.

maturity-matrix.yaml
```diff
@@ -0,0 +1,161 @@
...
+  - unknown
+  # "unstable" components are not recommended for production workloads. They
+  # may crash the process, introduce performance artifacts, or have known
+  # security issues.
+  - unstable
+  # "beta" components have been used successfully in production workloads at
+  # scale, but either due to the uncertainty introduced by active development
+  # or a small number of production deployments, they should still be used
+  # cautiously in new deployments. They do not have known security issues.
+  - beta
```
austinlparker: I don't know if it's possible, but it might be nice to try and lock down what 'production workloads at scale' means exactly.

maturity-matrix.yaml
mtwo: +1, I think that end-users expect at least some of these (OT exporter, Jaeger exporter, Prometheus exporter, basic HTTP integrations) to be tracked

maturity-matrix.yaml
mtwo: In the long-term, definitely. However I don't think that we need to until after the first release.

maturity-matrix.yaml
bhs: I do _not_ want to suggest that we try to maintain a full compatibility matrix with framework-level things like Spring. The "registry" seems like the right place to track that sort of thing.

By "automaticInstrumentation", I meant specifically the idea of "no-code" (literally) installation of the OpenTelemetry instrumentation. This is an important requirement for many brownfield services that are no longer actively maintained but nevertheless participate in transactions. As such, I would not like to extend the definition to "1-line code changes" that nevertheless require recompilation.

@SergeyKanzhelev re your comment:

>some integrations and exporters are the baseline for the stability 

I disagree – they may be the baseline for "usefulness", but not for stability. 

Given the possibly large diversity of exporters in the future, I would personally prefer that we:

1. Adopt the API and production maturity levels for exporters, and
2. Track exporter maturity via the registry

Remember that we're not debating how this information is displayed on a website at the moment, we're just debating what should be considered "internal" to OpenTelemetry rather than a plugin/etc. The website can absolutely bring in the maturity levels for popular exporters in designated languages.

maturity-matrix.yaml
bhs: I agree with @mtwo here... for now, I think we're basically implicitly tracking "latest stable" which probably won't be very stable for a few more months. :)

maturity-matrix.yaml
bhs: @lizthegrey re

>auto-instrumentation for golang includes wrappers for gorillamux, sqlx, etc. -- so it's not fully automatic but it's O(1) lines of code to add.

I'm sympathetic to the fact that it's hard to do better than this ^^ in Go, but I (unfortunately) don't think 1-line changes "count" for brownfield auto-instrumentation in many environments. This would support some sort of "n/a" designation per @SergeyKanzhelev's comment.

An alternative would be to define different levels of auto-instrumentation: perhaps "unsupported", "smallCode", and "noCode". Or, alternatively, to define auto-instrumentation criteria differently for different languages (with their different expectations).

maturity-matrix.yaml
bhs: I like the idea of being explicit here, or at least allowing for the option.

maturity-matrix.yaml
bhs: I'm a little hesitant to do this since in the early stages, that list will have lots of errors of omission and could lead to a false sense of security...

maturity-matrix.yaml
bhs: lol, true. This was lazy on my part.

How about `... production workloads at scale: that is, the component is used in support of well-known product functionality at publicly-held companies.`

I don't love this, but am hoping that it's difficult to become publicly-held without achieving some version of "scale"... admittedly not a perfect match for the kind of scale we're seeking here, but I am loath to add a Spans/sec threshold here.

SergeyKanzhelev: good enough I think

maturity-matrix.yaml
SergeyKanzhelev: can we declare it as optional so I can use it for .NET?

maturity-matrix.yaml
SergeyKanzhelev: Either n/a or just move it to the registry

maturity-matrix.yaml
SergeyKanzhelev: We need to avoid situation when OT is only usable with the specific vendor exporter. This is why I think we should declare some baseline exporters here. It may be minimum functionality and not scalable, but must be reliable and stable.

maturity-matrix.yaml
austinlparker: sgtm

maturity-matrix.yaml
austinlparker: I wonder about the 'diversity of exporters' -- Yuri could probably weigh in here, but I suspect that many "vendors", both OSS and commercial, would like to simply accept OTel data format natively and thus not require an exporter at all.

With that in mind, this feels like a transient issue; Yes, there'll be maturity levels for exporters but over time those exporters simply won't matter since your backend will accept the data with no translation.

maturity-matrix.yaml
tigrannajaryan: Please change Collector `api` maturity to "alpha". We expect minor breaking changes to public interfaces in the Collector codebase in the near future. Once that is done we will declare it "beta".

For `production` it matches the suggested definition of "beta" (run on production at a public company) so we can keep it.

maturity-matrix.yaml
SergeyKanzhelev: > I suspect that many "vendors", both OSS and commercial, would like to simply accept OTel data format natively and thus not require an exporter at all.

Not immediately. It may take a while

> With that in mind, this feels like a transient issue; Yes, there'll be maturity levels for exporters but over time those exporters simply won't matter since your backend will accept the data with no translation.

Yes, but if we intent to track progress towards 1.0 I'd suggest to include tracking of this transient issue

maturity-matrix.yaml
austinlparker: IMO it'd be good to have a URL field in the schema that points to the repo.

maturity-matrix.yaml
bhs: @SergeyKanzhelev I'm going to continue to push back on this... if I want to use Jaeger's exporter and it's the only thing that's mature for my given language, IMO that's fine. Now, whether the SIGs would feel comfortable marking their APIs as "stable" (or even "beta") with zero or one exporter implemented is a different question.

maturity-matrix.yaml
bhs: @tigrannajaryan done 👍

bhs:
Hi folks... I've updated the PR here in two ways:
* I removed the auto-instrumentation stuff until it's more fully-baked (it can go in the brand-new registry in the meantime)
* I added a blurb about "beta" APIs requiring multiple impls, with at least one being a major OSS project

Can reviewers either approve or request further changes? I can then fill in the gaps in other languages, potentially as follow-on PRs. cc @lizthegrey @SergeyKanzhelev @austinlparker @mtwo @tedsuo (who've all provided comments earlier – thanks).

maturity-matrix.yaml
lizthegrey: agree we should introduce n/a.

bhs: done

bhs:
One more ping on this – still looking for an approval from an approver!

bhs:
@lizthegrey thanks.

All: I'll merge this tomorrow or Tuesday if there are no further concerns.

As mentioned on the most recent Gov Committee call, I'll also make a very crude sketch of how we could represent this on the marketing site and submit an issue to track that. (I don't think I'll have the cycles to actually implement the marketing site piece, though)

bhs:
Thanks, all... Happy to take PRs on this file as we bring it into sync with reality.

PR #323, opened 2020-04-07
fix java calendar link

zeitlinger:
fix java calendar link

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,7 +181,7 @@ You can also join us on [gitter](https://gitter.im/open-telemetry/opentelemetry-
 
 ### Java SDK
 
-The Java SIG meets weekly on [Friday at 8 AM PDT](https://calendar.google.com/calendar/embed?src=lightstep.com_uuegcsvip16hv2qq31sj08sbkk%40group.calendar.google.com).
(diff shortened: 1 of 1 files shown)


SergeyKanzhelev:
@zeitlinger can you please sign CLA

zeitlinger:
> Alternatively you can use this link: https://calendar.google.com/event?action=TEMPLATE&tmeid=aGlqamsyYTV0dDh0ZDM1YmpkZWRjN2wyNjhfMjAyMDA0MTBUMTUwMDAwWiBnb29nbGUuY29tX2I3OWUzZTkwajdiYnNhMm4ycDVhbjVsZjYwQGc&tmsrc=google.com_b79e3e90j7bbsa2n2p5an5lf60%40group.calendar.google.com&scp=ALL Small problem is that it will require auth

yes, that is one click faster.
Should we change all calendar links to this format?

SergeyKanzhelev:
@zeitlinger I remember complains that links must not require auth with Google account. So perhaps this may be an alternative link. If you want to send an additional PR - it would be great. Merging this one for now.

zeitlinger:
> @zeitlinger I remember complains that links must not require auth with Google account. So perhaps this may be an alternative link. If you want to send an additional PR - it would be great. Merging this one for now.

2 links could also be confusing, so I'd leave it at that.
All you have to do is click on "more details" to add it to your calendar.

PR #354, opened 2020-05-26
moved candidates pictures into archive

SergeyKanzhelev:


SergeyKanzhelev:
@yurishkuro nope, those were broken with the archiving of the original file

PR #366, opened 2020-06-01
Update Java Auto-Instrumentation SIG scope

trask:
In agreement with both the Java SIG and the Java Auto-Instrumentation SIG, we have expanded the scope of the Java Auto-Instrumentation SIG to also include (manual) library instrumentation (and we have renamed the opentelemetry-auto-instr-java repo to opentelemetry-java-instrumentation).

This should help us to better re-use code and tests across both auto-instrumentation and (manual) library instrumentation, and provide a consistent experience for users regardless of which approach they prefer.


SergeyKanzhelev:
merging as it seems to be approved by both Java SIGs

PR #373, opened 2020-06-05
switch to easybridge cla

### LINKED ISSUE #309: Adopt new Communitybridge/Linux Foundation CLA bot
https://lfcla.com/ is the new Linux Foundation CLA bot, and we should switch to using it as it's less cumbersome than the existing Linux Foundation Account requirement.

lizthegrey:
Fixes #309

### DIFF
--- a/docs/how-to-configure-new-repository.md
+++ b/docs/how-to-configure-new-repository.md
@@ -46,11 +46,9 @@ policies:
    `Require review from Code Owners` is checked.
 2. The default setting for `Required approving reviews` is `1` approval. Individual 
    repositores may opt-in for bigger number of required reviews.
-3. Every repo MUST enforce `cla/linuxfoundation` check for `Require status
-   checks to pass before merging` section.
-   During the transition period, the CNCF `EasyCLA` check MAY be enabled.
-   After the transition period, it MUST be enabled, and `cla/linuxfoundation`
-   MUST NOT be enabled.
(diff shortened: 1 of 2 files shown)


docs/how-to-configure-new-repository.md
```diff
@@ -48,6 +48,9 @@ policies:
    repositores may opt-in for bigger number of required reviews.
 3. Every repo MUST enforce `cla/linuxfoundation` check for `Require status
    checks to pass before merging` section.
+   During the transition period, the CNCF `EasyCLA` check MAY be enabled.
```
yurishkuro, line 51: should the wording above "MUST enforce `cla/linuxfoundation` check" still remain?

lizthegrey: I can say `MAY be enabled instead of cla/linuxfoundation`, does that work?

yurishkuro: Since this is talking about configuring new repos, do we even need to mention the old CLA?

lizthegrey: No, because I think it also reflects our policy for existing repos, de facto.

yurishkuro: oh... the file is `docs/how-to-configure-new-repository.md`

in this case I would phrase it "at least one of these CLA checks MUST be required"

lizthegrey: PTAL

PR #394, opened 2020-06-18
Move Collector adopters to project-wide file

mtwo:
Coincides with https://github.com/open-telemetry/opentelemetry-collector/pull/1141


mtwo:
> 
> 
> let's not use `**_` formatting, it's extra maintenance and the words already stand out as hyperlinks

Fixing. I had originally copied the formatting from another CNCF project (Envoy I think?).

PR #430, opened 2020-07-29
Add OTLP maturity levels

tigrannajaryan:
I am not sure if the yaml structure we have here is formally defined here.
I invented an "otlp" section here assuming that the file has informal
structure. If this is not so please correct me.

This is dependent on https://github.com/open-telemetry/opentelemetry-proto/pull/185

### DIFF
--- a/maturity-matrix.yaml
+++ b/maturity-matrix.yaml
@@ -173,3 +173,8 @@ languages:
   php: ""
   cpp: ""
   rust: ""
+
+otlp:
+  traces: stable
+  metrics: alpha
+  common: stable


mtwo:
Is there value in splitting this up per signal type (traces, metrics, logs)?

tigrannajaryan:
> Is there value in splitting this up per signal type (traces, metrics, logs)?

Yes, I believe there is. We already clearly have this splited in the OTLP repository. We need the flexibility to assign different maturity levels to signals because that's the reality.

mtwo:
Nevermind, I'd glanced at the PR in a rush and missed that you distinguished between these already!

PR #441, opened 2020-08-10
No need for Admins group

SergeyKanzhelev:
Remove the requirement to make Admins group an admin of every repository. Org owners has enough permissions without this so this will remove an extra synchronization requirements.

### DIFF
--- a/docs/how-to-configure-new-repository.md
+++ b/docs/how-to-configure-new-repository.md
@@ -29,8 +29,7 @@ for the repository.
    repository.
 6. Root-level `CODEOWNERS` file on the repository should include superset of
    people from both `foo-approvers` and `foo-maintainers`.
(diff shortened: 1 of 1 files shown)


lizthegrey:
iirc the GC is in admins group but not org owners group?

bogdandrutu:
@lizthegrey the problem with having this Admins with GC members is that any GC review counts as "approver" which should not be the case. Maybe making GC members owners of the org should be discussed separately.

SergeyKanzhelev:
@lizthegrey I think the best thing to do is to create a special group for GC that has control over community repo. There is nothing in GC charter that requires admin priviledges on repositories, I think this Admin group is a leftover from the time we decided to remove GC members from GitHub org owners list.

PR #482, opened 2020-09-17
Added notice about the file headers

### LINKED ISSUE #113: Can we remove the requirement for license headers?
The license should already be defined at the root of each project, does each file really need such documentation as well?

Removing the headers would also make it much less hassle to update the license and copyright date when needed, since it would only need to be done in a single place.

SergeyKanzhelev:
Fixes #113


tigrannajaryan (commented):
This is a useful addition. I wonder if we want to consolidate this info with the "Copyright Notices" section that is in CONTRIBUTING.md? I think it would be good if it is all in one place, either here or in CONTRIBUTING.md - both approaches are fine with me.

bogdandrutu (approved):

PR #484, opened 2020-09-17
Azure Pipelines are not used any longer. Switched to GitHub actions

SergeyKanzhelev:


SergeyKanzhelev:
hm... still seems to be used by somebody. Will clear this out first

PR #596, opened 2021-01-11
Clarify maintains' access to public calendar

reyang:
This was discussed during the Jan. 11th, 2021 maintainers meeting.

1. Clarified that maintainers can schedule/organize SIG meetings.
2. Clarified the process on how to get write access to public calendar. (not sure if @SergeyKanzhelev will replace @mtwo)

@SergeyKanzhelev @mtwo please confirm.


SergeyKanzhelev (approved):
I'm also trying to set up a google group with the access to the calendar and we can have more admins for the group so it wouldn't be a single person

PR #606, opened 2021-01-14
update public calendar policies

SergeyKanzhelev:


assets.md
```diff
@@ -28,7 +28,14 @@ This file is intended to list all the assets controlled by OpenTelemetry.
...
 - [Calendar Google Group](https://groups.google.com/forum/#!forum/opentelemetry-calendar)
-    - Owner: Morgan McLean @mtwo
+    - Owner: @SergeyKanzhelev
+    - Write access controled via: opentelemetry-calendar-contributors@googlegroups.com
+
+- [opentelemetry-calendar-contributors Google Group](opentelemetry-calendar-contributors@googlegroups.com)
+    - Owners: current GC members.
+
+- [OpenTelemetry Calendar Invites Google Group](https://groups.google.com/g/opentelemetry-calendar)
+    - Owners: @mtwo
```
SergeyKanzhelev, line 38: @mtwo perhaps you can make opentelemetry-calendar-contributors@googlegroups.com an admin for the Invites group? And make GC members as owners

mtwo: Sure, doing it now

mtwo: Looks like Google Groups doesn't like that:
```
Error! A group may not be the owner or manager of another group
```

assets.md
```diff
@@ -28,7 +28,14 @@ This file is intended to list all the assets controlled by OpenTelemetry.
 ## Communication channels
 
 - [Calendar Google Group](https://groups.google.com/forum/#!forum/opentelemetry-calendar)
-    - Owner: Morgan McLean @mtwo
+    - Owner: @SergeyKanzhelev
+    - Write access controled via: opentelemetry-calendar-contributors@googlegroups.com
+
+- [opentelemetry-calendar-contributors Google Group](opentelemetry-calendar-contributors@googlegroups.com)
```
SergeyKanzhelev, line 34:
```suggestion
- [opentelemetry-calendar-contributors Google Group](https://groups.google.com/g/opentelemetry-calendar-contributors)
```

assets.md
```diff
@@ -28,7 +28,14 @@ This file is intended to list all the assets controlled by OpenTelemetry.
...
+    - Owners: current GC members.
+
```
SergeyKanzhelev, line 36:
```suggestion
    - Owners: current GC members.
    - Members: SIG maintainers and individuals appointed by maintainers

```

assets.md
SergeyKanzhelev: thinking about it I personally don't want to be a part of this group as I only import invites I need. Maybe you can keep yourself as a sole owner or maybe we need somebody from CNCF or from the current members as a backup

mtwo: Do we have a CNCF admin email address that we can use for things like that? I think that we do (for the Twitter account, etc.) but I don't know the details (?). Alternatively I can just add all of the GC and / or TC members as admins?

SergeyKanzhelev: @amye do you want to be a group owner?

@mtwo my problem owning this group is that it will flood my calendar with invites =).

alolita (approved):
+1

SergeyKanzhelev:
Let's keep it this way for now. We can fix the ownership of invites google group later

PR #607, opened 2021-01-15
Added Alolita as a channel manager

SergeyKanzhelev:
Also morgan's google corp e-mail is not actual any longer. @mtwo, pls request access any time if needed for other account


SergeyKanzhelev:
CC: @alolita

bhs (approved):
I think this LGTM... apologies if I'm missing something elsewhere, but was the removal of @mtwo intentional? If so, 👍 to merge...

PR #623, opened 2021-01-25
Replace community Gitter with GitHub discussions

mtwo:


sarahnovotny (approved):
looks right as long as the only SIGs moving today are Java SIGs.

SergeyKanzhelev (approved):

mtwo:
> 
> 
> looks right as long as the only SIGs moving today are Java SIGs.

And the community discussions. Some SIGs are still switching over and they'll update this page once they do.

PR #633, opened 2021-02-03
Update Adopters layout

flands:
In addition to listing adopters, list what components of OpenTelemetry
they have adopted. This is based on confirmed data -- others to be submitted in future PR(s) once confirmed.


yurishkuro:
TBH, I prefer the old bullet list format. It is easier to edit, and the exact dimensions of adoption aren't that important imo (can always be mentioned in free form text).

SergeyKanzhelev:
I think it is coming from this comment: https://docs.google.com/document/d/1f4WQR6K84giyOCKjCsqcoVA8YUIzBhWZDVo3CwM1wuk/edit?disco=AAAAHvn1Leo

yurishkuro:
Thanks for the reference. Doesn't change my opinion though. What's even more useful than the list of adopters are the case studies. For example, Jaeger's adopters page (which is quite stale as people using the OSS projects often don't bother to make it known) includes links to blogs and talks https://github.com/jaegertracing/jaeger/blob/master/ADOPTERS.md. The table format does not lend itself to this kind of information.

flands:
People interested in Otel may be interested in who is adopting what (I know I have been asked this many times). Agreed that the table format is harder to maintain. Another approach could be to have three columns: company, components, notes and just list pieces adopted under components. Let me update the PR and see how it looks.

flands:
@yurishkuro how about now?

yurishkuro:
Since components are now in free form, I think bullet list makes even more sense, eg

* company
  * components: ...
  * blog post: title...
  * talk: title...

weyert:
Personally, it's nice to see what's being adopted.

flands:
Sorry for the delay -- I tried the layout proposal, but at least to me it did not look as good as current PR. I checked and found similar examples like: https://github.com/openebs/openebs/blob/master/ADOPTERS.md. I would like to suggest keeping in its current form.

mtwo:
Agreed with @flands, I prefer the table layout similar to what's linked on the openebs repo

mtwo (approved):

SergeyKanzhelev:
for me it really depends on the goal. If page is just a collection of links - table is better. If we want to allow "statements of adoption" on the page with nice wording, quotes, and links, list would be better.

flands:
Today it appears like quick links -- would advise statements to be more like CNCF case studies.

SergeyKanzhelev (approved):
We can always change formatting if we will decide to make this page more like a case studies and statements

PR #736, opened 2021-05-13
update Python Zoom link

codeboten:


mtwo (approved):
Was there a mismatch between this and the calendar?

mtwo:
> 
> 
> Was there a mismatch between this and the calendar?

Nevermind, looked at the calendar and there was. Thanks for updating!

PR #742, opened 2021-05-27
Add a process for SIGs to resolve tech conflicts

bhs:
Technical consensus is wonderful when we can get there... and when we can't, this process should help us "disagree and commit."

All comments welcome!

cc @open-telemetry/governance-committee @open-telemetry/technical-committee

### DIFF
--- a/community-membership.md
+++ b/community-membership.md
@@ -244,7 +244,7 @@ that issue.
 4. If the vote is split, the OpenTelemetry TC should be brought into the
 discussion, and the TC itself gets a (single) tiebreaking vote.
 
-While inevitably these sorts of decisions will be "disappointing for somebody,"
+While inevitably these sorts of decisions will be disappointing for somebody,
 it's incredibly important for the project to maintain velocity and recognize
 that we are all coming to these sorts of technical issues with the best of
 intentions and remain aligned about the overall goals of the OpenTelemetry
--- a/community-membership.md
+++ b/community-membership.md
@@ -219,3 +219,33 @@ The following apply to the subproject for which one would be an owner.
 - Ensure a healthy process for discussion and decision making is in place.
 - Work with other maintainers to maintain the project's overall health and
   success holistically.
+
+### Resolving technical conflicts within a SIG
+
+From time to time, the Maintainers for a given OpenTelemetry SIG may be unable
+to reach consensus on a technical issue. While it's healthy and appropriate to
+make a sincere attempt to understand all points of view and consider the
+tradeoffs, it's also healthy to occasionally "disagree and commit."
+
+Within OpenTelemetry, SIG Maintainers are chosen specifically _because_ they
+are domain experts, so we would like to keep as much of the decision-making
+authority with the SIG Maintainers rather than immediately "escalating" to the
+overall OpenTelemetry-wide Technical Committee. As such, this is
+OpenTelemetry's process for resolving technical issues where Maintainers cannot
+reach consensus:
+
+1. The SIG Maintainers should succintly document the options under
+consideration as a GitHub issue within the SIG's respective repo (note that it
+is *not* required to document the complete framing and pros/cons – just the
+actual go-forward options themselves).
+2. Each SIG Maintainer must formally vote for their choice by commenting on
+that issue.
+3. The option receiving the most votes wins.
+4. If the vote is split, the OpenTelemetry TC should be brought into the
+discussion, and the TC itself gets a (single) tiebreaking vote.
+
+While inevitably these sorts of decisions will be "disappointing for somebody,"
+it's incredibly important for the project to maintain velocity and recognize
+that we are all coming to these sorts of technical issues with the best of
+intentions and remain aligned about the overall goals of the OpenTelemetry
+project.
--- a/community-membership.md
+++ b/community-membership.md
@@ -1,6 +1,6 @@
 # Community membership
 
-Last updated: 2020-03-04
+Last updated: 2021-05-27
 
 This doc outlines the various responsibilities of contributor roles in
 OpenTelemetry. The OpenTelemetry project is subdivided into subprojects under
--- a/community-membership.md
+++ b/community-membership.md
@@ -241,7 +241,7 @@ actual go-forward options themselves).
 2. Each SIG Maintainer must formally vote for their choice by commenting on
 that issue.
 3. The option receiving the most votes wins.
-4. If the vote is split, the OpenTelemetry TC should be brought into the
+4. If the vote is a tie, the OpenTelemetry TC should be brought into the
 discussion, and the TC itself gets a (single) tiebreaking vote.
 
 While inevitably these sorts of decisions will be disappointing for somebody,


lizthegrey (approved):

yurishkuro (approved):

community-membership.md
```diff
@@ -219,3 +219,33 @@ The following apply to the subproject for which one would be an owner.
...
+reach consensus:
+
+1. The SIG Maintainers should succintly document the options under
+consideration as a GitHub issue within the SIG's respective repo (note that it
+is *not* required to document the complete framing and pros/cons – just the
+actual go-forward options themselves).
+2. Each SIG Maintainer must formally vote for their choice by commenting on
+that issue.
+3. The option receiving the most votes wins.
+4. If the vote is split, the OpenTelemetry TC should be brought into the
```
dyladan, line 244: The word "split" implies a 1-2 vote would still require TC input.

```suggestion
4. If the vote is a tie, the OpenTelemetry TC should be brought into the
```

community-membership.md
```diff
@@ -219,3 +219,33 @@ The following apply to the subproject for which one would be an owner.
...
+discussion, and the TC itself gets a (single) tiebreaking vote.
+
+While inevitably these sorts of decisions will be "disappointing for somebody,"
```
dyladan, line 247: No need for quotes here. Quotes could be interpreted to imply that disappointment is invalid.

```suggestion
While inevitably these sorts of decisions will be disappointing for somebody,
```

dyladan (approved):

community-membership.md
```diff
@@ -219,3 +219,33 @@ The following apply to the subproject for which one would be an owner.
 - Ensure a healthy process for discussion and decision making is in place.
 - Work with other maintainers to maintain the project's overall health and
   success holistically.
+
+### Resolving technical conflicts within a SIG
```
SergeyKanzhelev, line 223: please change "Last updated: 2020-03-04" on the top.

SergeyKanzhelev, line 223: should it be h4 (subsection of maintainer's "Responsibilities and privileges")? Or this applied wider, also to approvers?

SergeyKanzhelev (changes requested):
text looks good, I have question on the placement

community-membership.md
bhs: I think this applies specifically to Maintainers (the next-highest H2), and I don't think of this as a Maintainer responsibility as much as "something relevant to maintainers"... I don't feel strongly and can change this if you do. :)

SergeyKanzhelev (approved):

community-membership.md
SergeyKanzhelev: ok, this will do than

community-membership.md
```diff
@@ -219,3 +219,33 @@ The following apply to the subproject for which one would be an owner.
...
+OpenTelemetry's process for resolving technical issues where Maintainers cannot
+reach consensus:
+
+1. The SIG Maintainers should succintly document the options under
+consideration as a GitHub issue within the SIG's respective repo (note that it
+is *not* required to document the complete framing and pros/cons – just the
+actual go-forward options themselves).
+2. Each SIG Maintainer must formally vote for their choice by commenting on
+that issue.
+3. The option receiving the most votes wins.
```
tigrannajaryan, line 243: If the issue can be solved by voting does it need to follow a formal process like this? The voting can happen in the SIG and I am not sure how making the voting formal helps unless for some reason we think we need to have a voting record.
I am not opposed to voting but would like to understand how we think formalizing it and recording as an issue helps.

community-membership.md
```diff
@@ -219,3 +219,33 @@ The following apply to the subproject for which one would be an owner.
...
+4. If the vote is a tie, the OpenTelemetry TC should be brought into the
```
tigrannajaryan, line 244: If it reached the point when maintainers cannot come to a consensus the TCs involvement may be necessary even when the voting is not a tie. If a minority of maintainers feel so strongly opposed to idea that they are willing to escalate I think we should allow that escalation to happen. I do not feel that saying "the majority of maintainers always win" is the right approach.

So, as an alternate to this PR's proposal how about we simply say: if maintainers cannot reach a consensus within the SIG they can submit a formal arbitration request as outlined in items 1-2 above and the TC will be responsible to make decision. No voting is necessary and we only resort to a formal process when it is clear the regular process in the SIG is not working.

Again, I am not opposed to the proposal in this PR, but I wonder if a simpler alternate may be more attractive.

community-membership.md
```diff
@@ -219,3 +219,33 @@ The following apply to the subproject for which one would be an owner.
...
+overall OpenTelemetry-wide Technical Committee. As such, this is
+OpenTelemetry's process for resolving technical issues where Maintainers cannot
+reach consensus:
+
+1. The SIG Maintainers should succintly document the options under
+consideration as a GitHub issue within the SIG's respective repo (note that it
+is *not* required to document the complete framing and pros/cons – just the
+actual go-forward options themselves).
+2. Each SIG Maintainer must formally vote for their choice by commenting on
+that issue.
```
tigrannajaryan, line 242: I like this approach.

community-membership.md
yurishkuro: Because voting *is* a formal process where someone is bound to lose, so it is better to have it documented. If the SIG was able to reach the consensus without it they wouldn't need to reach for this mechanism.

community-membership.md
bhs: @tigrannajaryan FYI, this PR is in response to an actual and outstanding technical deadlock happening in a SIG where there is an explicit desire for "an official process."

I share your desire for a simple approach but thought that filing a (short) issue that only outlines the options under consideration would be less complexity – and, frankly, drama – than escalating to the TC, agreeing on the framing document, etc.

Anyway, this is also just my two cents. I could also add something in here about how "the TC is always glad to hear both sides of a technical dispute and attempt to arbitrate or help find consensus"... thoughts?

yurishkuro: The issue is that escalating to TC could mean moving the decision from people on the SIG who know the language best to people on TC who may not know the language at all, which could be even less desirable outcome than a majority vote.

I do agree though that an attempt at mediation by TC could come before calling for a formal SIG vote. In such case TC can potentially bring a different, broader perspective.

yurishkuro:
> My only remaining questions is the following: do we want to additionally allow non-tie situations to be escalated?

Like appealing the SIG decision to TC - I think it's reasonable.

carlosalberto (approved):

mtwo (approved):

arminru (approved):
Thank you for also adding good reasoning and explanations which are important for sensitive topics like these.

community-membership.md
```diff
@@ -1,6 +1,6 @@
 # Community membership
 
-Last updated: 2020-03-04
+Last updated: 2021-05-27
```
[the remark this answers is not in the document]
bhs: There are similar things in other docs... I'll leave this in for now (removing them seems like it should be a cross-cutting PR).

bhs:
@tigrannajaryan re this comment:

> My only remaining questions is the following: do we want to additionally allow non-tie situations to be escalated? For example in a SIG with 3 maintainers in a situation of 2 against 1 do we feel that it is good enough to say that the majority is always right and there is no need for TC involvement?

Nothing prevents any SIG (or OTel community member) from reaching out to the TC at any time... my expectation is that a pattern of poor technical decisions *would* eventually get escalated, for instance. But if we allow non-tie escalations, there's not much point in voting in the first place, right?

PR #845, opened 2021-09-21
Remove monday instrumentation meeting

tedsuo:
We no longer need this meeting, it has been removed from the calendar.


alolita:
+1 Thanks @tedsuo for removing the calendar invite.

PR #856, opened 2021-09-28
update voting method

lizthegrey:
we are no longer using CIVS due to lack of github integration (since we don't know all participants' emails), and because CIVS requires email opt-in, having been (ab)used by RMS supporters to spam people

### DIFF
--- a/governance-charter.md
+++ b/governance-charter.md
@@ -153,9 +153,9 @@ members.
 
 ### Election process
 
-Elections will be held using time-limited Condorcet ranking on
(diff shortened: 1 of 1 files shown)


yurishkuro (approved):

SergeyKanzhelev:
curious, is there material difference between Condorcet ranking and Approval voting?

lizthegrey:
https://rangevoting.org/AppCW.html -- not in practice, only in an extreme adversarial tactical voting case.

PR #874, opened 2021-10-12
Self-nomination: Bogdan Drutu for Governance Committee

bogdandrutu:
Signed-off-by: Bogdan Drutu <bogdandrutu@gmail.com>


lizthegrey:
that's people working for three distinct employers, so it goes through.

PR #880, opened 2021-10-14
Self-nomination: Csuzhang Cheung for Governance Committee

hanyuancheung:
Please consider my nomination for the Governance Committee election. Thx~


lizthegrey:
This self-nomination is timely, but needs to receive 2 additional ratifications by tonight in order to be accepted as per the published election timeline.

lizthegrey:
@open-telemetry/otel-elections, I need a judgment call as to whether we accept this self-nomination that was _technically_ 7 hours late but could be excused under "timezone mix-up"

PR #884, opened 2021-10-14
Nominating Ilan Rabinovitch for GC

mikezvi:


dyladan:
Accepted and ratified

PR #913, opened 2021-11-18
Reflect calendar / Zoom updates

mtwo:
Updated our readme and documentation to reflect the recent improvements made to our calendar and Zoom links, which are now meeting-specific and are tied to a single CNCF Zoom account. Summary / justification:

I've updated all of our Zoom links in the calendar to use unique IDs. This has a few advantages:
- No more clashing meetings
- No more meetings from one SIG that unexpectedly flow into another SIG's call
- Meeting recordings should have titles now

There are some disadvantages:
- Meetings IDs and passwords are no longer predictable, meaning that meetings must be joined from the community calendar; I've removed static meeting links from the agenda docs and will remove them from GitHub
- Creating new meeting series will require access to the CNCF Zoom accounts

I made all of the edits through our new OpenTelemetry Google account, which is tied to the governance committee mailing list.

### DIFF
--- a/docs/how-to-handle-public-calendar.md
+++ b/docs/how-to-handle-public-calendar.md
@@ -1,24 +1,15 @@
 # How to create and configure meetings
 
-In OpenTelemetry we are using CNCF zoom account. We currently own four
-identical zoom accounts. The default room from every single account is
-used and a room can only host one meeting at a time. These are the
-account rooms:
-
-- https://zoom.us/j/6729396170
-- https://zoom.us/j/8203130519
-- https://zoom.us/j/8287234601
-- https://zoom.us/j/5227112777
-
-Meetings with the same room must not be held at the same time or be adjacent.
+In OpenTelemetry we are using the CNCF's Zoom accounts. We currently own four
+identical zoom accounts, though most meetings are created by the first account.
 
 All rooms are configured the same way. Some notes on meetings configuration:
 
-- **Host is not required** to join the meeting. Anybody can join the meeting.
+- **The host is not required** to join the meeting. Anybody can join the meeting.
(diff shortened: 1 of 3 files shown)


jpkrohling (approved):
LGTM. Should I have access to any credentials to manage any of those accounts? I'm especially curious about the "Project Google account" one.

docs/how-to-handle-public-calendar.md
```diff
@@ -34,11 +25,13 @@ meetings can join the group.
...
+to either link one of the OpenTelemetry Zoom accounts to your public calendar, or you can use the credentials for the
+cncf-opentelemetry-governance@lists.cncf.io  Google account, which is already linked to one of the CNCF Zoom accounts.
 
 Please keep the membership of this group up to date and accurate.
 
 ## Zoom bombing prevention
-All meetings now require a passcode to join: _77777_. This passcode is embedded into the calendar meeting invites and meeting notes documents.
+All meetings are created by Zoom with randomized passcodes, which are embedded into the shared calendar links.
 All members of opentelemetry-calendar-contributors@googlegroups.com have access to [this document](https://docs.google.com/document/d/1gt9ctxKGPrM_XTINqLgkSxYypdrczHkt2znjwgBU4UU/edit#)
 listing the host keys for our meetings and explaining how to deal with inappropriate behavior in Zoom.
```
arminru, line 37: Will these still work? We didn't have any Zoom bombings recently but people who joined and accidentally had their mic open while they left their desk with a lot of noise at home or in the office, so we had to mute them using the host key.

mtwo: Yes, the host accounts remain the same. The fundamental difference is that we're using scheduled meetings (from the same host) rather than Zoom rooms.

Issue #982, opened 2022-02-16
go-maintainers missing permissions


Aneurysm9:
The @open-telemetry/go-maintainers team is missing `admin` permissions on the [opentelemetry-go repo](https://github.com/open-telemetry/opentelemetry-go).  We have had these permissions in the past and require them to appropriately maintain our repository.

Can we have the `admin` permissions restored and an explanation of who removed them, when, and for what reason?

tigrannajaryan:
I am not sure about the "who removed" part, but I think the policy currently says that maintainers do not normally have admin permissions: https://github.com/open-telemetry/community/blob/main/docs/how-to-configure-new-repository.md#permissions

I think this was also discussed once in the past (granting admin permissions to maintainers) but I can't remember the outcome of that decision.

tigrannajaryan:
Here, found the discussion: https://github.com/open-telemetry/community/issues/616

bogdandrutu:
Based on the audit log this was done in Dec 2021.

<img width="925" alt="Screen Shot 2022-02-16 at 10 45 58 AM" src="https://user-images.githubusercontent.com/1373887/154334535-4bf0edd0-e1d5-4316-b1a2-0adb1460a766.png">

The reason is very simple, since as TC members we are responsible to respect the rules and maintain the organization, and the rules are [here](https://github.com/open-telemetry/community/blob/main/docs/how-to-configure-new-repository.md):
> The team foo-maintainers has Maintain permissions for the repository.

There is no exception/request documented that go maintainers need special set of permissions.

bogdandrutu:
I also updated lots of other places in the past, unfortunately the audit log is deleted after 6months, or at least I could not find it.

<img width="932" alt="Screen Shot 2022-02-16 at 10 50 12 AM" src="https://user-images.githubusercontent.com/1373887/154335190-6277846c-45fc-4c27-add1-798dd03471fc.png">

bogdandrutu:
@dyladan we discussed this in the GC, and I told you folks (2w ago if I remember correctly the time) that github allows us to define new "roles" with custom permissions, so we can define a "super-maintainer" role with more permissions if we identify that is needed. Also we can then update the community rules to give "super-maintainer" role to our maintainers.

[...]

PR #184, opened 2019-09-18
maintainers should be maintainers

SergeyKanzhelev:
Fixes #49 #84

Issue #84, opened 2019-05-31
Should Maintain (beta) permission level be set for maintainer's teams


SergeyKanzhelev:
There is a beta permission level setting:
https://help.github.com/en/articles/repository-permission-levels-for-an-organization

Perhaps it can be used for repo-maintainers team instead of "Write".

It will give some additional permissions like setting up CI hooks

SergeyKanzhelev:
this was fixed via #184

Issue #49, opened 2019-05-15
Repo configuration: repo-approvers team maintainers


SergeyKanzhelev:
I think `repo-maintainers` team members should be set as maintainers of `repo-approvers` team. Need to document it

PR #983, opened 2022-02-17
Allow maintainers admin permissions on a case-by-case basis

dyladan:
In some cases, maintainers have admin permissions already. This officially allows that on a case-by-case basis.

The decision was made during the governance committee meeting on 2/17/2022 to temporarily grant admin permission to maintainers while the permissions differences between admin and maintain roles can be assessed. Currently, the rule is already inconsistently applied and many maintainers already have this permission and depend on it as part of their workflows. If it is determined that the maintain role is sufficient for maintainers, or that a custom role can be created, then it should be documented on an issue/PR and this can be updated then.

Related issues and PRs:
- Increase permission from `Write` to `Maintain` permission
	- Issue #84
	- PR #184
- Related discussion in #982

I would like approval from both @open-telemetry/governance-committee and @open-telemetry/technical-committee members for this PR, and would also appreciate the input of any maintainers.

### DIFF
--- a/docs/how-to-configure-new-repository.md
+++ b/docs/how-to-configure-new-repository.md
@@ -30,8 +30,10 @@ for the repository.
    PRs. See the [Policies](#policies) section that suggest to ensure that `Require
    review from Code Owners` is checked.
 4. The team `foo-approvers` has `Write` permissions for the repository.
-5. The team `foo-maintainers` has `Admin` permissions for the
+5. The team `foo-maintainers` has `Maintain` permissions for the
    repository.
+   In individual cases, for example, if this is required to allow maintainers to publish releases independently,
+   the team `foo-maintainers` can be granted `Admin` permissions instead.
 6. Root-level `CODEOWNERS` file on the repository should include superset of
    people from both `foo-approvers` and `foo-maintainers`.
 
--- a/docs/how-to-configure-new-repository.md
+++ b/docs/how-to-configure-new-repository.md
@@ -30,13 +30,10 @@ for the repository.
    PRs. See the [Policies](#policies) section that suggest to ensure that `Require
    review from Code Owners` is checked.
 4. The team `foo-approvers` has `Write` permissions for the repository.
-5. The team `foo-maintainers` has `Maintain` permissions for the
+5. The team `foo-maintainers` has `Admin` permissions for the
    repository.
 6. Root-level `CODEOWNERS` file on the repository should include superset of
    people from both `foo-approvers` and `foo-maintainers`.
-7. Some repositories may include more individuals with `Admin` permissions.
-   Typically to help set up repository, CI, web hooks or other administrative
-   work.
 
 ![image](https://user-images.githubusercontent.com/9950081/57563719-d7b6b300-7355-11e9-9ebb-3c4f549336bc.png)
 


docs/how-to-configure-new-repository.md
```diff
@@ -30,13 +30,10 @@ for the repository.
    PRs. See the [Policies](#policies) section that suggest to ensure that `Require
    review from Code Owners` is checked.
 4. The team `foo-approvers` has `Write` permissions for the repository.
-5. The team `foo-maintainers` has `Maintain` permissions for the
+5. The team `foo-maintainers` has `Admin` permissions for the
    repository.
 6. Root-level `CODEOWNERS` file on the repository should include superset of
    people from both `foo-approvers` and `foo-maintainers`.
-7. Some repositories may include more individuals with `Admin` permissions.
```
bhs, line 37: are there examples where this privilege is presently in use / necessary?

bhs (commented):
The Maintain->Admin change LGTM. I do have a quick question/clarification about the removed bullet.

docs/how-to-configure-new-repository.md
dyladan: Giving maintainers admin privileges removes the need for this doesn't it?

bogdandrutu:
> Reverts a change made in https://github.com/open-telemetry/community/pull/184 which restricted maintainers from admin to maintain permission.

This is wrong, it was never Admin. So please correct and add the motivation to increase the scope.

bogdandrutu (changes requested):
Blocking since this is not a revert, it is an extra permission. So not sure we need "time" to analyze. We do have that time now.

bhs:
@bogdandrutu we discussed this for 25m at the GC meeting today – it's hard to quickly summarize such a long conversation, but basic argument that stuck with me was (a) maintainers already have the rights to modify the actual source code which means they are already highly trusted by definition, (b) the lack of Admin privs is causing friction in day-to-day work, and (c) GitHub does not allow for sufficiently granular permissions to do something custom / in-between here.

bogdandrutu:
@bhs the solution may be the good one, I have problems with the motivation written in the PR.

trask (approved):

SergeyKanzhelev (approved):
Definitely the right step that will help maintainers and TC members. PR description as @bogdandrutu said is not accurate, we always wanted more permissions for the maintainers, not less.

Has @open-telemetry/governance-committee decided on pursuing efforts to help keep repositories in compliant state? I'd suggest:

- get some github backup solution to back up code, issues, and metadata to protect from accidental deletion
- automation to restore EasyCLA enforcement on repositories
- audit of permissions granted in various repositories for compliance

jpkrohling:
@bogdandrutu: 
> This is wrong, it was never Admin.

There is a mismatch between the written rules and reality. As evidenced by the issue originating this discussion, some maintainers used to have admin privileges before on the repositories they maintain.

> Have not seen this, but I can believe it may be the case. Need some concrete examples.

The case that originated this issue is one example: maintainers need to change the required checks from time to time. While @dyladan provided a workaround, the same pattern adopted by the otel-go team is adopted by the otel-operator repository: whenever there's a new Kubernetes minor version available, we adjust the e2e tests to use that new version, making the check required while making the N-3 minor version not required anymore.

Another example that happened a few months ago involved the otelcol-releases repo, where I needed a secret to be set in the repository and couldn't get an otelcol maintainer to do it for about a week. I'm pretty sure I had a similar situation happen before that with the operator repository.

> It does see [here](https://docs.github.com/en/enterprise-cloud@latest/organizations/managing-peoples-access-to-your-organization-with-roles/managing-custom-repository-roles-for-an-organization)

I understand @SergeyKanzhelev [tried](https://github.com/open-telemetry/community/issues/982#issuecomment-1042602691) that already as part of #982.

jpkrohling:
> get some github backup solution to back up code, issues, and metadata to protect from accidental deletion

While I think that having a backup isn't a bad idea, I'm having a hard time understanding how an accidental deletion of a GitHub repository may occur: deleting a repository involves going to a page that is well "hidden" in the admin options, scrolling down to the last option, clicking a red button, typing the repository name and clicking another button.

> audit of permissions granted in various repositories for compliance

To be honest, no matter what we decide here, this is needed anyway.

dyladan:
@bogdandrutu updated

dyladan:
For what its worth, I was advocating for keeping maintain role when we were in the gov meeting. The argument was made that because of the reality of how the rule was inconsistently applied, some maintainers have come to depend on that role. Go is actually the only repo that I know of that depends on this. It may be sufficient to apply "maintain" privileges to all maintainers except go. There is a bullet point in the existing documentation that allows for other people not on the TC to have admin privileges if they need it.

As far as the CNCF `**/**` rule you are correct it cannot be removed. It can be worked around in some cases. For example (with permission from CNCF) there is a rule in otel js contrib which overrides the cncf rule for only branches with a particular name in order to allow our release automation to make release branches. The CNCF allowed this because we were able to do it for only a particular branch name and guarantee there would be no new intellectual property committed without a CLA. I'm not sure which usecase go removed the rule for so I can't comment on a particular workaround for them. Without knowing who removed it or why, we can't know if they depend on the rule being removed or if adding it back will break something for them.

If we do grant admin permissions, then I would argue at the very least we need to educate all admins which things they are allowed to change without approval from CNCF or TC and what they are not. CNCF's primary motivation in all of this is to make sure all commits have a CLA.

We could possibly make an automation that ensures the proper branch rules are in place and repo settings are correct, but it would be a lot of work and we might miss something. Maybe instead we can have a way to report admin changes. They should happen rarely, so it shouldn't be a large burden to tell the TC/GC any time a change is made which requires admin privileges. If the change violates some rule, then the maintainer can be educated of their mistake and it can be corrected.

docs/how-to-configure-new-repository.md
Aneurysm9: I would think that retaining this flexibility, even if rarely used, might be good.  Does removing this mean that it MUST NOT be done in the future, or does it become undefined whether such exceptions can be made?

bogdandrutu:
> If the permission was removed from the Go maintainers in December because it was thought that one of us had improperly removed a branch protection rule, why was nothing said to any of us at the time?

Who said that? As explained in the issue, the permission was removed to enforce the written rules. I did not see the removal of the default entry until 2 days ago when I asked about that.

dyladan:
I did say "that I know of" for the record. JS also has tests across runtimes and the CI checks are simply not required for us.

> I'm not sure that Go is the only group that requires the ability to manage required CI checks or secrets. It looks like Java, Python, and .NET also have matrix builds across runtime and OS combinations that need to be periodically maintained. Has anyone asked them how they're doing that? Has anyone actually talked to any maintainers about this topic?

I also noticed that . Most likely they are using their admin permissions to do this. That's what this PR was authored to address. Maintainers have de-facto admin permissions and simply removing it across the board is likely to be quite disruptive.

> As for not noticing that the permission had been removed, it is infrequently required, but that doesn't change the fact that it is required. You could also remove our permission to create releases without telling us and we may not notice for some time, but eventually we will notice and it will be an adverse impact on our ability to fulfill our role.

I agree with @Aneurysm9 here that "infrequently required" is not the same as "not required"

> If the permission was removed from the Go maintainers in December because it was thought that one of us had improperly removed a branch protection rule, why was nothing said to any of us at the time? Why were the audit logs not checked then? Have they been checked since? Why was a question only asked about it two days ago?

The permission was not removed because of the missing branch protection rule. In https://github.com/open-telemetry/community/issues/982#issuecomment-1042022085 @bogdandrutu states the permission was fixed simply because he noticed it was non-compliant with policy. I have no reason to doubt that is the reason. As far as I know, @bogdandrutu only noticed the missing rule two days ago and he did immediately ask about it in the slack channel. @bogdandrutu please correct me if that is wrong.

dyladan:
There seems to be quite a bit of confusion stemming from incorrect understanding of history and the timeline so I am going to write it down here. If anyone sees anything wrong here, please let me know and i'll update the comment.

1. Prior to #184 maintainers had `Write` permissions per the [community docs](https://github.com/open-telemetry/community/blob/d1aa602e0b38a7838c926bafc82eabbdb293bd97/docs/how-to-configure-new-repository.md?plain=1#L21)
2. In May of 2019 @SergeyKanzhelev opened the issues #49 and #84 to suggest we increase the permissions of maintainers from `Write` to `Maintain (Beta)`
3. In september of 2019, @SergeyKanzhelev opened the PR #184 to resolve the above issues which was merged with 2 approvals
4. In December of 2021, @bogdandrutu noticed the permissions were not as documented on the go repository and set the permission to the maintain permission which is correct as documented
5. Two days ago @Aneurysm9 opened the issue #982 asking to have admin permission restored, and asking who made the change.
6. During the following conversation, @bogdandrutu noticed the missing branch protection rule and asked in the maintainers channel why it was removed and if it was approved by CNCF

It is not clear when the following things happened
- The `**/**` rule is removed from the go repository
- Many maintainers (including go maintainers) are granted admin permissions. Some of those maintainers now depend on a particular admin permission (Change CI Required Checks) as a part of their workflows. It is currently unknown if other admin privileges are being used by any maintainers for regular tasks.

Aneurysm9:
> 5. Two days ago @Aneurysm9 opened the issue
https://github.com/open-telemetry/community/issues/982 asking to have admin permission restored, and asking who made the change. Presumably, the permission change was discovered by the go maintainers when they tried to update the required CI checks to reflect the [release](https://github.com/open-telemetry/opentelemetry-go/releases/tag/v1.4.1) they cut that morning (@Aneurysm9 please confirm this point for me).
> 6. During the following conversation, @bogdandrutu noticed the missing branch protection rule and asked in the maintainers channel why it was removed and if it was approved by CNCF

These two events are inverted.  Bogdan asked why the branch protection rule was removed and as a result, in the process of looking into that, I discovered that I had significantly reduced permissions in the Go repo.

dyladan:
> > 5. Two days ago @Aneurysm9 opened the issue
> >    [go-maintainers missing permissions #982](https://github.com/open-telemetry/community/issues/982) asking to have admin permission restored, and asking who made the change. Presumably, the permission change was discovered by the go maintainers when they tried to update the required CI checks to reflect the [release](https://github.com/open-telemetry/opentelemetry-go/releases/tag/v1.4.1) they cut that morning (@Aneurysm9 please confirm this point for me).
> > 6. During the following conversation, @bogdandrutu noticed the missing branch protection rule and asked in the maintainers channel why it was removed and if it was approved by CNCF
> 
> These two events are inverted. Bogdan asked why the branch protection rule was removed and as a result, in the process of looking into that, I discovered that I had significantly reduced permissions in the Go repo.

fixed

dyladan:
Given @bogdandrutu and @Aneurysm9 agreement on the above timeline, I would suggest we move forward with this change.

At the next maintainers meeting, we should inform the maintainers that we (GC/TC) are considering applying `maintain` permission to all SIG maintainers, and that this means removing `admin` permissions that many of them currently have. I suggest we present a deadline of 3 weeks for maintainers to identify parts of their workflow which require admin permissions, and let the TC know which admin permissions they require. After the 3 week deadline, the TC and GC should review the findings and decide on a permanent policy regarding permissions moving forward.

bogdandrutu:
@dyladan thanks for the timeline, that is very useful.

I am a bit confused by your comment:

>  I would suggest we move forward with this change.

This means we document that permission should be Admin, then in the next paragraph you say to set everyone to `Maintain`:

> At the next maintainers meeting, we should inform the maintainers that we (GC/TC) are considering applying maintain permission to all SIG maintainers, and that this means removing admin permissions that many of them currently have.

bogdandrutu (commented):
The change looks good to me, still I am not 100% sure about the issue references in the description since for example https://github.com/open-telemetry/community/issues/49 - refers to the hierarchy of the teams; https://github.com/open-telemetry/community/issues/84 - increases the permission from write -> maintain; https://github.com/open-telemetry/community/pull/184 - adds maintain permission;

Also https://github.com/open-telemetry/community/issues/982 does not seem to ask for this, but for other informations which was given to the issue author.

I want to be clear that, what TC members have been doing in the past was applying the rules, and this is a change of the rules and not a "reaction" to the TC members which were not doing the right thing.

dyladan:
@bogdandrutu I updated the description to clarify #84 and #184 increase from write to maintain and dropped #49 since it is irrelevant. I think #982 still has related discussion even if it is not targeting the exact same thing.

bhs (approved):

jpkrohling (approved):

tigrannajaryan (changes requested):
We have discussed this in the TC meeting and have some reservations about the change. 

It appears that the 2 situations where the Admin permissions are necessary are the following:
1. Secrets management. The TC thinks that requiring maintainers to manage secrets on each repository individually may not be the best approach. Github allows org-level secret management. We think that this may be a better approach to store secrets, which does not require Admin permission on repos. We likely need to discuss this separately to understand how exactly this will work and what the policies are going to be, how we rotate secrets, etc.
1. Branch protection rules. It is not exactly clear how often it is necessary to make this changes, but it doesn't seem to be necessary on a daily basis. We think that the TC can continue to be responsible for this. The maintainers will need to ask the TC for such changes. The TC commits to being responsive and make changes as necessary. If we find that this happens much more frequently than we imagine or the TC is not responsive enough and blocks the maintainers then we can reconsider this approach.

Other than the above 2 use cases are the other situations when Admin permissions are needed? If there are please describe them. I am blocking this for now on behalf of the TC until we have better understanding of the other use cases.

dyladan:
> We have discussed this in the TC meeting and have some reservations about the change.
> 
> It appears that the 2 situations where the Admin permissions are necessary are the following:
> 
> 1. Secrets management. The TC thinks that requiring maintainers to manage secrets on each repository individually may not be the best approach. Github allows org-level secret management. We think that this may be a better approach to store secrets, which does not require Admin permission on repos. We likely need to discuss this separately to understand how exactly this will work and what the policies are going to be, how we rotate secrets, etc.
> 2. Branch protection rules. It is not exactly clear how often it is necessary to make this changes, but it doesn't seem to be necessary on a daily basis. We think that the TC can continue to be responsible for this. The maintainers will need to ask the TC for such changes. The TC commits to being responsive and make changes as necessary. If we find that this happens much more frequently than we imagine or the TC is not responsive enough and blocks the maintainers then we can reconsider this approach.
> 
> Other than the above 2 use cases are the other situations when Admin permissions are needed? If there are please describe them. I am blocking this for now on behalf of the TC until we have better understanding of the other use cases.
> 
> To be clear: this is not about TC not trusting maintainers. It is simply to reduce the possibility of accidents and mistakes.

In addition to those usecases I can say there are multiple repositories (at least go, java, and python) who change which CI checks are required each time a new version is released. This also requires admin permission.

trask:
> In addition to those usecases I can say there are multiple repositories (at least go, java, and python) who change which CI checks are required each time a new version is released. This also requires admin permission.

one option for this use case might be creating an `accept-pr` job that can be used as the CI check, this has worked well in the java-isntrumentation repo:

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/bd7bf94d71ddd520c2dff383a2fac9cc51749c22/.github/workflows/pr.yml#L301-L312

dyladan:
To the best of my knowledge, nobody is claiming that giving admin permissions to maintainers is the correct long-term solution. That being the case, many maintainers already have admin privileges and there are clear examples where they are depending on that permission for their workflows. The documentation in the community repository does not match with this reality. The proposal in the PR is to update the community documentation to match the real world to give the TC and GC time to create a real maintainer permission strategy. From what I can tell, there is no such existing strategy.

Blocking this PR after a week of public iteration by requesting changes with no suggested resolution is extremely disappointing to me. At least tell me what you would expect to see. The current situation where the documented policy and the real world don't match is obviously not acceptable, but enforcing the existing policy by simply stripping the admin privileges from those maintainers that have it may be very disruptive to those maintainers who are currently depending on it.

In an above comment https://github.com/open-telemetry/community/pull/983#issuecomment-1045070745, I clearly stated these things and got +1's from multiple GC and TC members, which led me to assume it was going to be acceptable.

> > In addition to those usecases I can say there are multiple repositories (at least go, java, and python) who change which CI checks are required each time a new version is released. This also requires admin permission.
> 
> one option for this use case might be creating an `accept-pr` job that can be used as the CI check, this has worked well in the java-isntrumentation repo:
> 
> [open-telemetry/opentelemetry-java-instrumentation@`bd7bf94`/.github/workflows/pr.yml#L301-L312](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/bd7bf94d71ddd520c2dff383a2fac9cc51749c22/.github/workflows/pr.yml#L301-L312)

This is exactly the type of workaround that I hope will allow us to restrict the maintainer permissions in the future, but it doesn't remove the need for a short term fix to give us time to come up with a long term plan.

tigrannajaryan:
>... with no suggested resolution is extremely disappointing to me. At least tell me what you would expect to see. 

Sorry, I was not clear. The TC's suggestion right now is:
- Remove Admin rights in all repos to match the current documented policy.
- The TC will make any changes that require Admin permissions. The maintainers will need to ask a TC member to help them.
- Monitor for maintainers' requests for a while to understand if this is a manageable situation (how often is the help needed, how quickly TC member respond) and if not grant Maintainers Admin rights.

Also, not right now, but as soon as we have a good understanding on how:
- Move secret management to org level.

I'll let other TC members to comment in case I have missed anything, I tried to summarize our today's discussion as best as I could.

tigrannajaryan:
Also, if I am not strongly opposed to granting Admin rights to Maintainers if we do no believe the [proposed approach](https://github.com/open-telemetry/community/pull/983#issuecomment-1049173989) is good enough for now. We can grant the permissions and work on a better solution in parallel.

tigrannajaryan:
> In addition to those usecases I can say there are multiple repositories (at least go, java, and python) who change which CI checks are required each time a new version is released. This also requires admin permission.

How often do the releases happen? Presumably monthly? With a need to change something monthly in every repo I am on the fence of making it a TC responsibility (it is not very frequent, but nor is it so rare to say that it is a no-brainer).

trask:
> We can grant the permissions and work on a better solution in parallel.

This would be my preference, giving us time to plan and transition. I can help work on this from the maintainer side.

tigrannajaryan:
@open-telemetry/technical-committee please chime in.

dyladan:
> > ... with no suggested resolution is extremely disappointing to me. At least tell me what you would expect to see.
> 
> Sorry, I was not clear. The TC's suggestion right now is:
> 
> * Remove Admin rights in all repos to match the current documented policy.
> * The TC will make any changes that require Admin permissions. The maintainers will need to ask a TC member to help them.
> * Monitor for maintainers' requests for a while to understand if this is a manageable situation (how often is the help needed, how quickly TC member respond) and if not grant Maintainers Admin rights.

This is fine as long as the TC can commit to being responsive on these requests (<24 hours). My preference would be to go the opposite direction, grant the permission, actively seek feedback from maintainers on which permissions they actually use, and propose work-arounds that remove maintainer dependence on those permissions.

> Also, not right now, but as soon as we have a good understanding on how:
> 
> * Move secret management to org level.

I don't really have any preference on this as I think changing secrets is something that happens very rarely outside of regular key rotations.

> > We can grant the permissions and work on a better solution in parallel.
> 
> This would be my preference, giving us time to plan and transition. I can help work on this from the maintainer side.

This is essentially what I am suggesting.

> > In addition to those usecases I can say there are multiple repositories (at least go, java, and python) who change which CI checks are required each time a new version is released. This also requires admin permission.
> 
> How often do the releases happen? Presumably monthly? With a need to change something monthly in every repo I am on the fence of making it a TC responsibility (it is not very frequent, but nor is it so rare to say that it is a no-brainer).

It probably varies SIG to SIG but there are possibly workarounds that don't require the checks to be modified every time. For example, you could create a single job named `all checks passed` which depends on other checks. If the `all checks passed` check is required, then the ones it depends on are de facto required as well even if they aren't explicitly required. This is essentially what java instrumentation is doing.

tigrannajaryan:
>This is fine as long as the TC can commit to being responsive on these requests (<24 hours).

I think this is doable.

However, given that there is a use case which we were not aware of I am going to revoke my block. We explicitly discussed in TC meeting that we would like to block until we hear of a use case which we do not know about.

If any other TC member has a strong opinion please chime in.

yurishkuro:
I keep seeing "grant now, sort out the needs later", but what is the current emergency that requires granting admin rights to all maintainers right now? Our desired state is not to grant admin rights, so if there is an emergency, could we not sort it our on case by case basis?

> This is fine as long as the TC can commit to being responsive on these requests (<24 hours).

What is the specific use case that requires this kind of SLO?

docs/how-to-configure-new-repository.md
```diff
@@ -30,13 +30,10 @@ for the repository.
    PRs. See the [Policies](#policies) section that suggest to ensure that `Require
    review from Code Owners` is checked.
 4. The team `foo-approvers` has `Write` permissions for the repository.
-5. The team `foo-maintainers` has `Maintain` permissions for the
+5. The team `foo-maintainers` has `Admin` permissions for the
    repository.
```
arminru, line 34: How about this?

```suggestion
5. The team `foo-maintainers` has `Maintain` permissions for the
   repository.
   In individual cases, for example, if this is required to allow maintainers to publish releases independently,
   the team `foo-maintainers` can be granted `Admin` permissions instead.
```
This is similar to the previous wording but would reflect the status quo where some maintainer _teams_ rather than individuals have admin access and would thus fix the current state for now without making `Admin` the default for _all_ (new) repos.

If solutions can be found that suit all maintainers (like the suggestion provided by @trask in https://github.com/open-telemetry/community/pull/983#issuecomment-1049155557), we can remove this clause again in future.

dyladan: @tigrannajaryan does this work for you?

dyladan:
> I keep seeing "grant now, sort out the needs later", but what is the current emergency that requires granting admin rights to all maintainers right now? Our desired state is not to grant admin rights, so if there is an emergency, could we not sort it our on case by case basis?
> 
> > This is fine as long as the TC can commit to being responsive on these requests (<24 hours).
> 
> What is the specific use case that requires this kind of SLO?

It isn't that it needs to be done extremely quickly as much as things that should be simple now have an outside dependency, so that dependency should have as little impact as possible. The specific number isn't important to me and we don't have to codify an SLO, but I wanted to just make clear that if the TC wants to go with the more restrictive approach to resolving this (which would be fine with me by the way) they need to minimize the impact of that.

docs/how-to-configure-new-repository.md
tigrannajaryan: Yes, works for me.

arminru (approved):

yurishkuro (approved):

bogdandrutu (approved):

docs/how-to-configure-new-repository.md
SergeyKanzhelev: I'm totally ok removing it now. Going forward if we will start developing more automation, this or similar item may be needed back. I think ultimate goal is to make maintainers life easier and if it will require "external" help, this should be fine.

alolita (approved):

tigrannajaryan (approved):

PR #1014, opened 2022-03-29
Add 1Password to assets

dyladan:


codeboten (approved):
👍

assets.md
```diff
@@ -2,6 +2,13 @@
 
 This file is intended to list all the assets controlled by OpenTelemetry.
 
+## Credential Storage
+
+- [1Password Teams](https://1password.com/)
+  - https://github.com/1Password/1password-teams-open-source#opentelemetry
```
arminru, line 8:
```suggestion
  - Team account: https://opentelemetry.1password.com/
  - https://github.com/1Password/1password-teams-open-source#opentelemetry
```

arminru (approved):

assets.md
```diff
@@ -2,6 +2,14 @@
 
 This file is intended to list all the assets controlled by OpenTelemetry.
 
+## Credential Storage
+
+- [1Password Teams](https://1password.com/)
+  - Team account: https://opentelemetry.1password.com/
+  - https://github.com/1Password/1password-teams-open-source#opentelemetry
+  - PR: https://github.com/1Password/1password-teams-open-source/pull/364
+  - Admin: Alex Boten @codeboten
```
yurishkuro, line 11: Unlike other entries in this file, this looks like a shared utility. Could we add some description of how to use it?

yurishkuro: The latter.
* Are we going to share passwords across all maintainers of all SIGs, or each SIG should have its own 1password sub-account (don't know if possible), etc.
* Do we want to enforce the model where published artifacts are always produced by bot accounts (whose password is shared), instead of adding personal accounts of maintainers to the "owners team" in each respective artifactory?

dyladan: Seems like those are questions for the GC/TC to decide on. For now, I thought it was important just to get this in this list so that people know who the admin is. The resource already exists so it should be documented that it exists.

yurishkuro (approved):

assets.md
dyladan: Created https://github.com/open-telemetry/community/issues/1063 for these questions. I haven't used the teams features of 1Password but a quick browse of their website seems to show that access to resources can be granted to only specific people or subteams. This seems like a reasonable way for the GC/TC to disseminate access to resources to maintainers without broadcasting secrets on an insecure channel.

trask (approved):

dyladan:
@open-telemetry/governance-committee can we get this approved and merged? There is a related issue #1063 to determine how it should be best used but for now it is an asset that is owned and should be listed as such.

jpkrohling (approved):

PR #1106, opened 2022-07-08
Replace donation template with a form

yurishkuro:
cc @bhs 

Add a new required section "reason for donation"


bhs (approved):
Works for me. Thanks for the YML change, too.

Issue #1147, opened 2022-08-29
SIG documents should be owned by an OpenTelemetry account


dyladan:
Related #1146

All SIG meeting note documents should be owned by an OpenTelemetry or CNCF google account. This will prevent access issues like we have seen before in the related issue. This has also happened in the past when OTel members have changed employers.

There should also be a defined ownership/administrative permission structure in order to handle instances of abuse, spam, or other issues with the docs. For example, it would be desirable for maintainers to have sufficient permission to change a document to a read-only state or to roll back to a previous history version.

dyladan:
The following documents, unless checked off, are owned by contributors and not owned by the opentelemetry account:

- [ ] [Specification SIG doc](https://docs.google.com/document/d/1pdvPeKjA8v8w_fGKAN68JjWBmVJtPCpqdi9IZrd6eEo/edit) owned by @carlosalberto 
- [ ] [Log SIG doc](https://docs.google.com/document/d/1cX5fWXyWqVVzYHSFUymYUfWxUK5hT97gc23w595LmdM/edit) owned by @tigrannajaryan 
- [ ] [Sampling SIG doc](https://docs.google.com/document/d/1gASMhmxNt9qCa8czEMheGlUW2xpORiYoD7dBD7aNtbQ) owned by @tedsuo 
- [ ] [Instrumentation SIG doc](https://docs.google.com/document/d/1dWHhyXnfVife-cQ2DW5-d5Ldp1Lq8Rre2UsHpyo8cEE) owned by @tedsuo 
- [ ] [Maintainers meeting doc](https://docs.google.com/document/d/14lBtLAdzsV7M8xQNdhTJ0d2PxM1R7FByw-GttKJzVVw/edit) owned by @carlosalberto 
- [ ] [Collector Doc](https://docs.google.com/document/d/1r2JC5MB7GupCE7N32EwGEXs9V_YIsPgoFiLP4VWVMkE/edit?usp=sharing) owned by @flands
- [ ] [C/C++: SDK Doc](https://docs.google.com/document/d/1i1E4-_y4uJ083lCutKGDhkpi3n4_e774SBLi9hPLocw/edit?usp=sharing) owned by @reyang
- [ ] [.NET: Automatic Instrumentation Doc](https://docs.google.com/document/d/1XedN2D8_PH4YLej-maT8sp4RKogfuhFpccRi3QpUcoI/edit?usp=sharing) owned by @pjanotti
[...]

PR #1159, opened 2022-09-06
Update Agent Management doc to be owned by GC

dyladan:
Related to #1147


tigrannajaryan (approved):

tigrannajaryan:
Linked from old to new doc and old doc is now readonly.

tigrannajaryan:
Updated calendar.

dyladan:
@tigrannajaryan do you think updating the calendar before this PR merges is a good idea?

tigrannajaryan:
> @tigrannajaryan do you think updating the calendar before this PR merges is a good idea?

I made the old doc readonly and added an announcement in it linking to the new one, so there is no way for one to use the wrong doc even if they find it through here, so it should be fine.

arminru (approved):

tigrannajaryan:
@open-telemetry/technical-committee we need more approvals to merge this.

Issue #1209, opened 2022-09-29
Nomination: @cartermp for Governance Committee


lizthegrey:
As per https://github.com/open-telemetry/community/blob/main/governance-charter.md#eligibility-for-candidacy, anyone may nominate themself or another person for GC. I nominate @cartermp and am a Member of Standing, so count as one of the three Members of Standing required. Carter will also need to accept if he so wishes once he is back from PTO.

sharrmander:
I support this nomination.

cartermp:
I accept! Will formally run once back from PTO.

lizthegrey:
Three members of standing have ratified including self-ratification.

svrnm:
I support this nomination :-)

Just a formality:
> Anyone may nominate either themselves or someone else to be a candidate in the election. To be ratified as a candidate, the nominee must accept the nomination and three Members of Standing (including the nominator, if she/he has standing) **from three different employers**, must endorse the nomination.

@lizthegrey & @cartermp you're both with honeycomb, right? So I fixed that.

lizthegrey:
oh, oops, yes, good call. Thank you for fixing!

PR #1231, opened 2022-10-06
Add Phillip Carter to GC nominations list

cartermp:
Closes https://github.com/open-telemetry/community/issues/1209 and adds myself to the list!


lizthegrey (approved):
as per ratifications in #1209 we're good.

PR #1237, opened 2022-10-07
Add note about ranked voting and alternative voting systems

jpkrohling:
Based on today's GC meeting, I added the summary of the discussion about the evaluated voting systems, along with our requirements.


yurishkuro (approved):

sharrmander (approved):
I support the election committee evaluating other voting options which allow for ranked choice voting

bhs:
Having personally run a CIVS election, the actual *voting* part seemed great... BUT the registration situation was a minor nightmare since it's email-only.

If there was a way to do GitHub-authenticated ranked-choice voting, that would certainly be ideal!

lizthegrey:
> Having personally run a CIVS election, the actual _voting_ part seemed great... BUT the registration situation was a minor nightmare since it's email-only.
> 
> If there was a way to do GitHub-authenticated ranked-choice voting, that would certainly be ideal!

K8s has it configured but we don't.

https://github.com/kubernetes/community/blob/master/elections/teams/README.md

jpkrohling:
> K8s has Elekto configured but we don't.

I can commit to checking with the CNCF if we can get their help in running Elekto next year.

yurishkuro (approved):

PR #1259, opened 2022-10-13
Clarify that administrators must be subject to branch protection rules

tigrannajaryan:
We still allow one exception: in an emergency, when no-one else is available to approve the change.

If we agree with this change we need to audit all repos to make sure the "Do not allow bypassing the above settings" is enabled everywhere.

### DIFF
--- a/docs/how-to-configure-new-repository.md
+++ b/docs/how-to-configure-new-repository.md
@@ -41,19 +41,24 @@ for the repository.
 
 ## Policies
 
-Typically `main` branch is set as default. And protected with the following
-policies:
+Typically `main` branch is set as default.
+
+The default branch MUST be protected with the following policies:
 
 1. Enable `Require pull request reviews before merging` setting. Make sure
    `Require review from Code Owners` is checked.
 2. The default setting for `Required approving reviews` is `1` approval. Individual 
-   repositores may opt-in for bigger number of required reviews.
+   repositories may opt-in for bigger number of required reviews.
 3. Every repo MUST enforce the CNCF `EasyCLA` check under the
    `Require status checks to pass before merging` section. (if the check isn't
    showing, have a CLA manager [enable the check for the
    repo](https://project.lfcla.com/#/project/a0941000002wBz4AAE/cla)). Check
    with the OpenTelemetry's [EasyCLA admins](../assets.md#user-content-opentelemetry-managed-assets:~:text=Easy%20CLA) 
-4. It is a good practice to check the `Include administrators` setting.
+4. Enable "Do not allow bypassing the above settings" so that administrators are subject
+   to the same checks. In case of emergency and if they are unable to obtain the required
+   approvals from other approvers the administrators may **temporarily** bypass this
+   check and immediately revert it back as soon as the emergency change is done.
+   The fact of bypassing MUST be actively communicated. 
 5. Repository MUST enforce `Restrict who can push to matching branches` setting
    to only allow the members of `foo-maintainers` to push to the `main`
    branch.


yurishkuro (approved):

docs/how-to-configure-new-repository.md
```diff
@@ -41,19 +41,24 @@ for the repository.
...
 2. The default setting for `Required approving reviews` is `1` approval. Individual 
-   repositores may opt-in for bigger number of required reviews.
+   repositories may opt-in for bigger number of required reviews.
 3. Every repo MUST enforce the CNCF `EasyCLA` check under the
    `Require status checks to pass before merging` section. (if the check isn't
    showing, have a CLA manager [enable the check for the
    repo](https://project.lfcla.com/#/project/a0941000002wBz4AAE/cla)). Check
    with the OpenTelemetry's [EasyCLA admins](../assets.md#user-content-opentelemetry-managed-assets:~:text=Easy%20CLA) 
-4. It is a good practice to check the `Include administrators` setting.
+4. Enable "Do not allow bypassing the above settings" so that administrators are subject
```
cartermp, line 57: Unless there's an audit log of some sort, I don't think this accomplishes much other than requiring an admin to spend another minute enabling/disabling things. Is the thinking that this prevents unintentional damage by an admin merging something without approvals by accident?

Instead, I'd prefer just making it a requirement that all admins must communicate to the other maintainers when they are doing something like this. I don't know how to enforce it, but that's just part of the trust we place in people with admin permissions.

tigrannajaryan: There is an audit log that shows who and when changed the repo settings.

I think the increased friction is desirable in this case to both prevent accidents and also to remind the person that they doing something exceptional.

I don't think there is a need to strictly enforce this. The only way to do that would be to disallow Admin permission to maintainers and I think that goes too far.

jpkrohling (approved):

docs/how-to-configure-new-repository.md
cartermp: Makes sense!

docs/how-to-configure-new-repository.md
```diff
@@ -41,19 +41,24 @@ for the repository.
...
+   to the same checks. In case of emergency and if they are unable to obtain the required
+   approvals from other approvers the administrators may **temporarily** bypass this
```
arminru, line 59:
```suggestion
   approvals from other approvers, the administrators may **temporarily** bypass this
```

arminru (approved):

docs/how-to-configure-new-repository.md
```diff
@@ -41,19 +41,24 @@ for the repository.
...
-   repositores may opt-in for bigger number of required reviews.
+   repositories may opt-in for bigger number of required reviews.
 3. Every repo MUST enforce the CNCF `EasyCLA` check under the
    `Require status checks to pass before merging` section. (if the check isn't
    showing, have a CLA manager [enable the check for the
    repo](https://project.lfcla.com/#/project/a0941000002wBz4AAE/cla)). Check
    with the OpenTelemetry's [EasyCLA admins](../assets.md#user-content-opentelemetry-managed-assets:~:text=Easy%20CLA) 
-4. It is a good practice to check the `Include administrators` setting.
+4. Enable "Do not allow bypassing the above settings" so that administrators are subject
+   to the same checks. In case of emergency and if they are unable to obtain the required
```
arminru, line 58: Do you think there would be any good way to define what constitutes an emergency?
Maybe we could just give some representative examples.

melbgirl:
Do we want to apply this to every repo uniformly (or is this just for the community repo?)? Also, to the point that others have raised up-thread and in [1851](https://github.com/open-telemetry/opentelemetry.io/pull/1851), what happens in the case of emergencies? Are there exceptions in the case of critical or urgent issues?

docs/how-to-configure-new-repository.md
trask: TIL today on a non-otel repo:

Without "Do not allow bypassing the above settings", admins can git push directly to main (oops 😅).

trask:
I think this is resolved now (by #1311), 

https://github.com/open-telemetry/community/blob/main/docs/how-to-configure-new-repository.md#branch-protection-rule-main

> * Do not allow bypassing the above settings: ✔️
>
> ...
>
> **Important:** the only ones of these rules which may be changed are
>
> ...
>
> * Do not allow bypassing the above settings
>   * Maintainers may temporarily disable this

I think the one remaining open issue from the discussion in this PR is whether we want to mention examples of when it is ok to temporarily disable it.

PR #1282, opened 2022-11-01
Corrected Comms SIG meeting time

melbgirl:
Was 10:30, has been 10:15 for a while.


trask (approved):

dyladan:
@austinlparker when is the meeting time changing? probably makes sense to try to keep the repo as current as possible but maybe we can open a draft that reflects the new time

svrnm:
From now on we will meet on every other Monday 10-11am US PT, so we can update it to that

svrnm:
@melbgirl can you update the PR to say that we now meet on monday 10-11am PST?

dyladan:
@svrnm I was going to update it myself but it wasn't clear from your message, do you mean every monday or is it still every other monday?

dyladan:
Nevermind I see

dyladan (approved):

Issue #954, opened 2022-01-20
Add opentelemetry-java-bot to EasyCLA allowlist


trask:
Based on the discussion at https://github.com/open-telemetry/community/issues/809#issuecomment-1016522419, can we get @opentelemetry-java-bot added to the EasyCLA allowlist? we would like to use it to push some automatically-generated, non-copyrightable code as part of release automation.

(and thanks to @bobstrecansky and @austinlparker who alerted me to the cool OTel language icons)

bogdandrutu:
@trask  https://github.com/opentelemetry-java-bot is an org, you need to setup the right permissions for CNCF admins, etc. Why not a repo in this org?

trask:
@bogdandrutu it's not meant to be a repository, it's a bot user account (see @dyladan's comment linked above)

dyladan:
When i go to https://github.com/opentelemetry-java-bot I get a 404. Is the account private or did you delete it?

trask:
😞 github flagged and hid the account b/c I already had one "machine" account. trying to get it fixed, will post here once it's back online

trask:
@opentelemetry-java-bot account is unflagged and ready for EasyCLA allowlisting 🙏

dyladan:
@bogdandrutu what made you think its an org? Looks like a user to me

trask:
bump 🙏

dyladan:
I'm not sure what the process is to get this done. Last time we had something allowlisted I think @lizthegrey helped us so maybe she can help here?

trask:
I realized that since its our own bot account, I can just log in to it and sign the CLA (using my real name on the CLA signature)

PR #1285, opened 2022-11-03
Add opentelemetrybot to assets page

### LINKED ISSUE #809: Add `github-actions[bot]` to EasyCLA allowlist
Similar to https://github.com/open-telemetry/community/issues/306 which was resolved by @lizthegrey  making a request to the CNCF: https://jira.linuxfoundation.org/servicedesk/customer/portal/4/SUPPORT-1388

I would make a similar request myself (as a member of the GC) but I'm honestly not sure if I have the authority to do so without official approval of the GC or TC.

Example PR with failing CLA https://github.com/open-telemetry/opentelemetry-js/pull/2409

trask:
Based on discussions in #809 and #954

Resolves #809


svrnm (approved):

assets.md
```diff
@@ -118,3 +118,11 @@ Link: https://project.lfcla.com/#/project/a0941000002wBz4AAE/cla
 ### Docker Hub
 
 - We publish images from CI to Docker hub using https://hub.docker.com/u/otelbot account. The bot is registered using cncf-opentelemetry-tc@lists.cncf.io email address and Technical Committee members are owners of this account. The Admin for bot security credentials for CI is @tigrannajaryan
+
+### OpenTelemetry Bot
+
+This is a bot that has signed the CNCF CLA and can be used as the author of automated PRs.
```
arminru, line 124: Could you mention which repos are making use of this account?
This would make it easier to adapt permissions or rotate tokens if we know where things might break and need to be adapted.

assets.md
```diff
@@ -118,3 +118,11 @@ Link: https://project.lfcla.com/#/project/a0941000002wBz4AAE/cla
...
+
+Link: [@opentelemetrybot](https://github.com/opentelemetrybot)
+
+- Admins: @trask
```
arminru, line 128: We should have some redundancy here. Could you share the credentials with one or two other maintainers?
Ideally, the credentials would be added to the GC-owned 1Password account listed in this document.

arminru (approved):

assets.md
dyladan: @trask if you don't have access to the GC 1password account ping me on slack

assets.md
trask: So far:
* opentelemetry-java
* opentelemetry-java-instrumentation
* opentelemetry-java-contrib
* opentelemetry-python
* opentelemetry-sandbox-web-js
* opentelemetry.io (soon)

What do you think of going ahead and making it an org secret? (e.g. `BOT_TOKEN`, `OPENTELEMETRYBOT_TOKEN`, ...)

arminru: Thank you for the list!

It would be an option. On the other hand, if they have to be added manually, there is better control over where they are being used in case something needs to be changed (although grepping for the usage of `OPENTELEMETRYBOT_TOKEN` should be just as fine). Also, the risk of the credentials being misused is restricted to only the repos where it was manually added. If it's a repo secret, individual tokens can be assigned to each repo, each with their own permissions (and maybe separate audit logs, not sure about that). Since new repos won't be set up to use the bot that often, I think the maintenance effort is small enough to keep the manual process and have the other benefits instead.
We could discuss that further on the maintainer's call or a dedicated issue if you'd like.

trask: I added a note clarifying that @opentelemetrybot does not need to be (and should not be) given any permissions to any OpenTelemetry repositories. If the PAT leaked, the worst that could be done is to submit a PR as @opentelemetrybot, which would still require a maintainer to approve and merge it.

If you think this should have an individual token per repo, I'll send out new keys to folks, and create a list of which repo is using which PAT in the GC-owned 1password account (did you want the list of just the repos to be maintained in this doc also?).

If you don't think it needs an individual token per repo, I think we can still restrict an org secret to specific repos if we want to track which ones are using it.

arminru: Good point about the permissions! f we can be that restrictive about them, then I think an org secret should work and would also require lower maintenance. I was not aware of the option to restrict them to select repositories but if we do it like that, I think we should even be fine without having to manually (=error-prone) maintain such a list in 1Password or here.

trask: @arminru I made some updates to incorporate the org secret approach, ptal, thx

arminru: LGTM, thank you!

arminru (approved):

trask:
I've confirmed this approach (signing the CLA for our own bot) with the EasyCLA team:

> The Community usually ask EasyCLA support to add bot accounts authorization on their repos/projects. We, the support team, manually add the bot authorization as an Individual CLA. Since you signed the ICLA in behalf of https://github.com/opentelemetrybot, there is further actions required, as the process you followed is the same the support team would have done.

trask:
@dyladan I believe this PR will resolve #809?

dyladan:
Yes I think so

dyladan (approved):

dyladan:
How does this work with the default branch protection rules? As far as I understand it you can push a branch and create a PR but then the branch isn't updateable so if the PR gets out of date it can never be merged.

trask:
> How does this work with the default branch protection rules? As far as I understand it you can push a branch and create a PR but then the branch isn't updateable so if the PR gets out of date it can never be merged.

these are the branch protection rules we're using in the java-instrumentation repo for `opentelemetrybot/*` branches: https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/.github/repository-settings.md#dependabot-and-opentelemetrybot

trask:
I'll add `dependabot/**/**` branch protection to the base repository-settings.md in #1311, and mention that the same can be used for opentelemetrybot branches as well

dyladan:
> > How does this work with the default branch protection rules? As far as I understand it you can push a branch and create a PR but then the branch isn't updateable so if the PR gets out of date it can never be merged.
> 
> these are the branch protection rules we're using in the java-instrumentation repo for `opentelemetrybot/*` branches: https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/.github/repository-settings.md#dependabot-and-opentelemetrybot

Cool that should work. You should probably also mention that the order in which branch protection rules are created is important so you have to delete the `**/**` rule, create the exemption branch rules, then recreate `**/**` in order for the exemptions to apply.

trask:
> You should probably also mention that the order in which branch protection rules are created is important so you have to delete the `**/**` rule, create the exemption branch rules, then recreate `**/**` in order for the exemptions to apply.

done 👍

tedsuo (approved):

Issue #1137, opened 2022-08-19
Ensure security policy is available on all repos


austinlparker:
Planeposting, apologies for lack of links --

Someone raised a security vulnerability report for the collector (presumably) in slack. The security tab on the collector repo doesn't actually have the right security.md doc, it has one that specifies architecture/user concerns around security while operating the collector. 

Posting this issue to accomplish the following:

- ensure each repo has the correct security doc available through the security tab in GH
- ensure community repo has security/vuln reporting info linked from its readme

hickeyma:
I was the person chatting on slack with @austinlparker @seankhliao @mhausenblas about this. I was looking for the Open Telemetry security policy but couldn't find it first off.

@seankhliao pointed me to the Go security tab:  https://github.com/open-telemetry/opentelemetry-go/security/policy. 

Upon further investigations I found that the 

- Collector security tab collector displays to the collector security documentation/guideline instead: https://github.com/open-telemetry/opentelemetry-collector/security/policy
- Community security tab displays the Open Telemetry security policy: https://github.com/open-telemetry/community/security/policy

trask:
hey @austinlparker, it looks like this has been fixed in the collector repo, and I don't see any other repos overriding `SECURITY.md`:

https://github.com/search?q=org%3Aopen-telemetry+path%3A**%2Fsecurity.md&type=code

Issue #933, opened 2021-12-11
Security response in OpenTelemetry


flands:
Hey everyone -- as many have seen, there is a major security vulnerability in log4j 2.0 - 2.14.1. If you [search for log4j in the opentelemetry-java-instrumentation repository](https://github.com/open-telemetry/opentelemetry-java-instrumentation/search?q=log4j) it may appear that the instrumentation is vulnerable. Turns out, the instrumentation instruments log4j but does not pull in log4j as a dependency even though from the build.gradle it looks like it does (my understanding is that the dependency is handled via muzzle and only uses what is available on the application if anything -- it is not pulled in by the instrumentation itself).

This issue is interesting for a couple of reasons:
1. It raises the concern of security response in OTel -- clearly, a [SECURITY.md](https://github.com/open-telemetry/.github/blob/main/SECURITY.md) exists, but what is the response? For example, [Kubernetes](https://kubernetes.io/docs/reference/issues-security/security/#security-vulnerability-response))
2. It raises the concern of how issues like this should be handled. Should the minimum version be increased so scanners do not report a false-positive, should an issue be created to track per repo (thus something publicly referenceable by others), should a note be added to some README (again something public to reference), should something else be done?

tigrannajaryan:
> 1. It raises the concern of security response in OTel -- clearly, a [SECURITY.md](https://github.com/open-telemetry/.github/blob/main/SECURITY.md) exists, but what is the response? For example, [Kubernetes](https://kubernetes.io/docs/reference/issues-security/security/#security-vulnerability-response))

The vulnerabilities reported to cncf-opentelemetry-tc are received by TC. The TC typically creates an appropriate private issue in the affected repo and asks the maintainers to work on it. We can document this process if we want.

For log4j we have not received any report. I guess everyone was busy patching their own stuff.

>It raises the concern of how issues like this should be handled. Should the minimum version be increased so scanners do not report a false-positive, should an issue be created to track per repo (thus something publicly referenceable by others), should a note be added to some README (again something public to reference), should something else be done?

@open-telemetry/java-instrumentation-approvers please comment on this.

trask:
> Should the minimum version be increased so scanners do not report a false-positive?

I'd prefer not to do this, as the instrumentation intentionally supports old versions of libraries (but does not ship or transitively pull in those old versions).

> Should an issue be created to track per repo (thus something publicly referenceable by others)?

[...]

Issue #690, opened 2021-03-31
Need an org-scoped access token for docs automation


austinlparker:
We're implementing a solution for allowing authoritative website docs to live in each SIG repository, but the automation will need a PAT or other token that can create PR's in the website repository. You can see the issue in this PR: https://github.com/open-telemetry/opentelemetry-java/pull/3080

SergeyKanzhelev:
Here we discussed that PAT from one of SIG members may be used: https://github.com/open-telemetry/community/issues/551 Or SIG-specific bot account created. SIG members are in the best position to maintain this.

If you have a better idea in mind on how to manage this - please comment.

anuraaga:
Hi @SergeyKanzhelev - to give more context, this is about the shared docs repo at https://github.com/open-telemetry/opentelemetry.io and not quite a SIG-specific request. Each SIG will host their language's documentation in their repo, but we will have automation to push the code changes into that shared repo for deployment to opentelemetry.io. So we need a token with push access to the shared repo - I don't think we should expect SIGs to have to maintain this. As it stands, for example, Java maintainers do not have write access to that repo, so there's not really anything we can do to make the automation work.

I'd recommend as this is an open-telemetry shared process / account, there be an organization secret with a TC member PAT in it (I guess TC members would have write access to all the repos including opentelemetry.io) and expose that to language repos for use in this workflow. Does it make sense?

SergeyKanzhelev:
I understand the context, but wouldn't any website's SIG member token work for the purpose of merging to website repository? Or create a PR so it may be manually approved.

I'm advocating for less dependencies and more independence.

trask:
@austinlparker do you still have this need?

Issue #681, opened 2021-03-16
Propose User Research Workstream


mattmccleary:
See [Proposal](https://docs.google.com/document/d/12YIqGoqyYpTteTa9nzv-KXNkjPEGuQHN_ScawW5QXAM/edit?usp=sharing).

This proposal:
- Compliments @tedsuo's [OpenTelemetry Tracing Roadmap](https://docs.google.com/document/d/1jwtB0vFZ0uXOxHZJpuIaiFLnrFSiGDJRtX7VNI3DULc/edit?usp=sharing), including the [Convenience API](https://docs.google.com/document/d/1bjiw5L4E8FztRtfsy4Nu9A6qSP-BJlK6BBzl6Rea-Ds/edit?usp=sharing).
- Uses some similar concepts that @bobstrecansky used in previous [PHP user research](https://medium.com/opentelemetry/opentelemetry-garnering-user-feedback-for-a-better-product-650b4433e805).

trask:
@open-telemetry/end-user-wg does this fall under your workstream now? thx!

reese-lee:
Ah I think we've been unaware of this proposal. We'll take a look.

musingvirtual:
I support this proposal mostly as written, although a couple of these functions (named customer interviews & the survey) already exist and we want to integrate them.  

@mattmccleary I pinged you in a thread in Slack so we can discuss ideas and how to work together more easily, but happy to chat here too.  I believe the general thrust is that Reese and I are happy to work with you on this as a function of the end user working group.

reese-lee:
I would like to learn more about the usability test portion, and what it would entail to create one and run it.

PR #1326, opened 2022-12-16
Add 'needs author feedback' workflow

trask:
I like this stale workflow, which requires maintainers to add a "needs author feedback" label to an issue or PR before the stale bot takes effect, and the "needs author feedback" label is automatically removed on any further comment by the issue or PR author.

These are some example issues where I would like to apply this label: #933, #1137, #690, #681.

### DIFF
--- a/.github/workflows/issue-management-feedback-label.yml
+++ b/.github/workflows/issue-management-feedback-label.yml
@@ -1,0 +1,21 @@
+name: Issue management - remove needs feedback label
+
+on:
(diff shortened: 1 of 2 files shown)


yurishkuro (approved):

trask:
(I should mention, planning to wait until after the holidays to start applying the label)

arminru (approved):

dyladan (approved):

Issue #1173, opened 2022-09-09
2022 Election Announcements 


jpkrohling:
This issue will contain election announcements and the complete list of eligible voters. Subscribe to this issue if you wish to be notified of changes. Contact @open-telemetry/otel-elections if you have questions.

If you are not automatically detected as a member of standing and believe your contributions should make you eligible to vote, you may request an exemption by filling out a [registration form](https://forms.gle/mEDWyn6G7iCe4bvJ7).

People mentioned in this issue have been detected as members of standing, as [defined in the project's governance charter](https://github.com/open-telemetry/community/blob/main/governance-charter.md#members-of-standing). 

You can find more information about the voting process [here](https://github.com/open-telemetry/community/blob/main/elections/2022/governance-committee-election.md)

Due to the potential notification storm that this issue may cause, the conversation is locked. If you have concerns, please contact the group @open-telemetry/otel-elections, or any of the members (@jpkrohling or @bogdandrutu) individually. They are also available on the [CNCF Slack](http://slack.cncf.io/).

jpkrohling:
We invite the people mentioned in this comment to vote in the OpenTelemetry Governance Committee, as they have provided more than 20 contributions to the project over the last year via GitHub. Your contributions were comments, code reviews, pull requests, among others. Thank you!

* @﻿1046102779
* @aabmass
* @aaron-ai
* @aaronontheweb
* @aateeqi
* @abe545
* @abelperezok
* @abhee11
* @abhide
* @aboguszewski-sumo
* @acm19
* @adilhusain-s
* @adriankostrubiak-tomtom
* @ahayworth
* @aishyandapalli
* @alanwest
* @albertteoh
* @alexvanboxel
* @allex1
* @alolita
* @alvinhom
* @amber0612
* @amoscatelli
* @andrewhr
* @andrewzenkov
* @andyfleming
* @andykellr
* @aneurysm9
* @antonblock
* @anupamdalmia10
* @anuraaga
* @anuragagarwal561994
* @aptomaketil
* @arhell
* @arielvalentin
* @arminru
* @armstrmi
* @arun-shopify
* @aschugunov
* @ash2k
* @ashu658
* @astencel-sumo
* @astitcher
* @atingchen
* @atoulme
* @austinlparker
* @avadhut123pisal
* @avillela
* @axw
* @ayandapalli

jpkrohling:
[...]

PR #1371, opened 2023-02-23
Add Jack Berg as a TC member

reyang:
The TC (Technical Committee) members have discussed and decided to have @jack-berg joining the TC.

### DIFF
--- a/community-members.md
+++ b/community-members.md
@@ -25,6 +25,7 @@ in alphabetical order:
 - [Armin Ruech](https://github.com/arminru), Dynatrace
 - [Bogdan Drutu](https://github.com/BogdanDrutu), Splunk
 - [Carlos Alberto](https://github.com/carlosalberto), Lightstep
+- [Jack Berg](https://github.com/jack-berg), New Relic
 - [Josh MacDonald](https://github.com/jmacd), Lightstep
 - [Josh Suereth](https://github.com/jsuereth), Google
 - [Reiley Yang](https://github.com/reyang), Microsoft
@@ -53,7 +54,6 @@ Trace Approvers:
 Metrics Approvers:
 
 - [Cijo Thomas](https://github.com/cijothomas), Microsoft
-- [Jack Berg](https://github.com/jack-berg), New Relic
 - [Leighton Chen](https://github.com/lzchen), Microsoft
 - [Tyler Yahn](https://github.com/MrAlias), Splunk
 
@@ -62,7 +62,6 @@ Logs Approvers:
 - [Christian Beedgen](https://github.com/kumoroku), Sumo Logic
 - [Daniel Jaglowski](https://github.com/djaglowski), observIQ
 - [David Poncelow](https://github.com/zenmoto), Splunk
-- [Jack Berg](https://github.com/jack-berg), New Relic
 
 ## Java
 


arminru (approved):

yurishkuro (approved):

MrAlias:
Can someone on the TC link to the election? I wasn't even aware that the election process for a new TC member was open. Given [the charter](https://github.com/open-telemetry/community/blob/main/tech-committee-charter.md#election-of-tc-members) allows for any member with standing to nominate a candidate it would have been nice to have been informed this election was taking place.

reyang:
> Can someone on the TC link to the election? I wasn't even aware that the election process for a new TC member was open. Given [the charter](https://github.com/open-telemetry/community/blob/main/tech-committee-charter.md#election-of-tc-members) allows for any member with standing to nominate a candidate it would have been nice to have been informed this election was taking place.

The link was only shared among TC members.

@MrAlias to give you some context:
* I nominated and sponsored @jack-berg.
* @arminru created the vote and shared the link with existing TC members, following [_"A candidate can be elected to the TC by the super-majority vote (greater than two thirds) of the existing TC members. The voting by the TC members is anonymous"_](https://github.com/open-telemetry/community/blob/main/tech-committee-charter.md#election-of-tc-members).
* After a week we got votes from all the TC members and the decision was made.

MrAlias:
From [the charter](https://github.com/open-telemetry/community/blob/main/tech-committee-charter.md#election-of-tc-members):

> New TC members can be nominated by any Member in Standing as defined in the [OpenTelemetry Governance Charter](https://github.com/open-telemetry/community/blob/master/governance-charter.md#members-of-standing)

There was no "call for nominees" made to members in standing as far as I can tell. I was expecting something similar to https://github.com/open-telemetry/community/issues/1173

> @MrAlias to give you some context:
> 
> * I nominated and sponsored @jack-berg.
> * @arminru created the vote and shared the link with existing TC members, following [_"A candidate can be elected to the TC by the super-majority vote (greater than two thirds) of the existing TC members. The voting by the TC members is anonymous"_](https://github.com/open-telemetry/community/blob/main/tech-committee-charter.md#election-of-tc-members).
> * After a week we got votes from all the TC members and the decision was made.

Sure, I'm concerned that there was a closed election that was not announced to the OTel community where a single candidate was nominated and sponsored by a single TC member and the election results are not verifiable nor public. This lack of transparency is troubling given this is supposed to be an open project and it seems contrary to the letter and spirit of the TC charter.

Can the election results be made public?

When is the next TC election expected and when can members of standing nominate candidates?

MrAlias:
To be clear, I happy to hear @jack-berg is on the TC, but I'm more concerned about the transparency and accountability of the election that was held.

trask:
Whether the TC election process should be updated is a fair question and one that the TC and GC have on the agenda to discuss in two weeks.

I do believe the TC followed the currently documented process:

> TC memberships are not time-limited. There is no maximum size of the TC. The size is expected to vary in order to ensure adequate coverage of important areas of expertise, balanced with the ability to make decisions efficiently.

My understanding based on this is that there is no set schedule for TC elections, and that TC elections are not necessarily about filling a certain number of seats.

> New TC members can be nominated by any Member in Standing ... and must be sponsored by one of the existing TC members to bring it to a vote.

My understanding based on this is that any Member in Standing can nominate anyone to the TC at any time, but it requires sponsorship from an existing TC member to bring it to a vote.

MrAlias:
> My understanding based on this is that any Member in Standing can nominate anyone to the TC at any time, but it requires sponsorship from an existing TC member to bring it to a vote.

So this effectively means only the TC can nominate a candidate and they can do so whenever they see fit?

What about the election results? Why are these not published? Is the TC the only body with oversight in the election?

To be clear, I don't know if this is intentional, but this gives the perception that the OTel is not an open community run by the community, but one run by a small subset that all see eye-to-eye.

It seems like diversity of opinions, and other dimensions, might improve if this wasn't the case.

MrAlias:
Reflecting on where this conversation has gone, it is clear now that there are deeper issues than I initially thought. Having this discussion in a closed, and unrelated, PR does not seem appropriate. I'll plan to open an issue here in the community repository instead to address those issue.

If there is a record of the election results though, I would appreciate it if it could be posted here for future reference.

yurishkuro:
@MrAlias can you elaborate on what kind of "record of the election results" you are looking for?  Per the charter, the voting is anonymous, similar to the GC election voting, so the "result" is essentially this PR.

MrAlias:
https://vote.heliosvoting.org/helios/elections/76558134-3384-11ed-8688-02871af94755/view

jsuereth:
@MrAlias I think there's a few points I'd like to make here:

- Sometimes the vote becomes unanonymous - I.e. when there's a no we discuss it and look to resolve concerns over time.  This has been really helpful in terms of understanding concerns, addressing them and thinking holistically about technical representation.
- The vote for TC members, so far, has been unanimous "yes".  We haven't resorted to supermajority IIRC.
- TC is balanced by GC owning our charter.  They can rework our charter if this is a need.  I suspect you'll see some subtle and not-so-subtle changes coming.  We're currently trying to set up our own structure for the best success of the project going forward.

> What about the election results? Why are these not published? Is the TC the only body with oversight in the election?
>
> To be clear, I expect this wasn't the intention, but this gives the perception that the OTel is not an open community run by the community, but one run by a small subset that all see eye-to-eye.
>
> It seems like diversity of opinions, and other dimensions, might improve if this wasn't the case.

I think not having an open election for technical leadership leads to stronger OSS projects.   Generally there's a non-popularity-bar that needs to be met where folks understand the implications, goals and focus of a project.  Diversity of opinion is good.  Diversity of vision can be lethal.  

By this I mean if we all agree we're building a sailboat, the person wanting to build a submarine should go build a submarine in a different project (unless we all agree we'll build a sailboat and a submarine, in which case we may end up building neither).

This is where having the TC + GC balance each other is helpful.  GC calls out poor diversity of thought.  TC calls out lack of focused vision and scope-death of the project.

PR #1401, opened 2023-03-29
Update opentelemetrybot details

trask:
Adds note about 2FA being available now in 1Password.

Removes note about maintainers needing to request access to org secret, because I believe the secret has been made available to all repos (need someone from @open-telemetry/technical-committee to confirm).

Adds note about additional scope that has been granted to the personal access token.


arminru (approved):
Thank you for sorting out 2FA as well now!

arminru (approved):

PR #1408, opened 2023-04-04
Update community-members.md

bryce-b:
Added OpenTelemetry-swift to list

PR #1411, opened 2023-04-05
Add an entry for opentelemetry-swift

kevinearls:
Add an entry for the open telemetry-swift project


arminru (commented):
@bryce-b already started the same in #1408.
Which version would you like to move forward with?

PR #1433, opened 2023-04-13
Fix old abbreviation: TSC -> TC

xiang17:
Didn't find any reference of TSC in this repo. A search in the issues reveals it's an old abbreviation: #248 #161


yurishkuro (approved):
thanks

trask (approved):

PR #1450, opened 2023-04-23
Add StackOverflow -> Slack Channeler

austinlparker:
To support community efforts at answering questions on StackOverflow, this Lambda function will query the SO API and post new results in a specified channel via Slack webhook.

We'll need to figure out what AWS account to run this on -- I can probably start it on my team account, but we'll need a Slack app connected to the CNCF Slack instance which I can request. Mostly just wanted to make sure the code was archived in this repo.

### DIFF
--- a/tools/so-channeler/README.md
+++ b/tools/so-channeler/README.md
@@ -1,0 +1,58 @@
+# StackOverflow -> Slack Question Channeler
+
+This is a Lambda function designed to channel questions from StackOverflow
+tagged with `#open-telemetry` into a designated Slack channel.
+
+## Packaging and Deployment
+
+1. Run `pip3 install requests -t ./package` to install the `requests` library
+   into the `package` directory.
+2. Copy `lambda_function.py` into the `package` directory.
+3. Zip the contents of the `package` directory into a zip file named
+   `lambda_function.zip`. **Do not zip the `package` directory itself, zip its
+   contents**.
+4. Upload the zip file to AWS Lambda. Deploy the function.
+5. Adjust the timeout to ~1 minute (adjust as needed) and set environment
+   variables.
+6. Create an EventBridge trigger with the desired schedule (hourly would be
+   `cron(0 * * * ? *)`) and set the target to the Lambda function.
+
+## Requirements
+
+The Lambda function requires the following environment variables to be set:
+
+- `SLACK_WEBHOOK_URL`: The URL of the Slack webhook to post to.
+- `S3_BUCKET_NAME`: The name of the S3 bucket to store the last timestamp in.
+
+It also requires an IAM Policy and Role. The policy is below (replace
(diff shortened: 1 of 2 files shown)


yurishkuro:
Could we run this as GH workflow on schedule?

tools/so-channeler/README.md
```diff
@@ -0,0 +1,58 @@
...
+   variables.
+6. Create an EventBridge trigger with the desired schedule (hourly would be
+   `cron(0 * * * ? *)`) and set the target to the Lambda function.
+
+## Requirements
+
+The Lambda function requires the following environment variables to be set:
+
+- `SLACK_WEBHOOK_URL`: The URL of the Slack webhook to post to.
+- `S3_BUCKET_NAME`: The name of the S3 bucket to store the last timestamp in.
```
yurishkuro, line 25: another idea - use a dedicated GitHub repo as a "database". The volume of updates is relatively small.

austinlparker:
No reason we couldn't run it as an action, although we'd need to have an API key generated that could write/commit to the repo. Lambda was just easier for me to spin up.

austinlparker:
Here's a github actions version that will read/write to a state file.

yurishkuro:
Does CNCF slack already has an "app" for which webhooks can be activated? I can't just create one.

image

austinlparker:
> Does CNCF slack already has an "app" for which webhooks can be activated? I can't just create one.
> 
> 
> 
> image
> 
> 

Quite possibly, but I'll have to open a Service Desk ticket to get an admin to find out.

tools/so-channeler/gha/README.md
```diff
@@ -0,0 +1,29 @@
+# StackOverflow to Slack GitHub Action
+
+This GitHub Action fetches new StackOverflow questions tagged with `open-telemetry` and posts them to a Slack channel. It runs on an hourly schedule.
+
+## Setup
+
+1. Clone this repository or create a new repository with the provided `github_action.py` script and `.github/workflows/main.yml` workflow file.
```
arminru, line 7: Could we also just create the repo where this will be running in right away rather than having it in the community repo?
To separate the implementation from the state that's being kept it could just maintain this `state.txt` on a separate branch, similar to the gh-docs branches some repos use.

austinlparker:
Yes, if someone wants to go ahead and create a repo for me I can close this PR and set everything up. We do have a webhook now and I'm running the lambda version to send questions into Slack, once I get it migrated I'll turn off my lambda and use the repo.

dyladan:
I believe only the @open-telemetry/technical-committee can create a repo and to my understanding that is the current blocker for this PR

yurishkuro:
added https://github.com/open-telemetry/stackoverflow2slack

PR #1463, opened 2023-05-01
Remove outdated statuses from maturity-matrix.yaml

trask:
I didn't know this existed until @tigrannajaryan pointed it out. Probably worth deleting it since the website documents stability of different languages / signals, and the spec is probably where "definition of stability terms" should reside (and is being discussed in https://github.com/open-telemetry/opentelemetry-specification/issues/3459)


dyladan:
Is this yaml used by anything? Website for example? I'm fine removing it since its outdated anyway but it seems weird to me that a yaml file would have no consumer.

maturity-matrix.yaml
arminru, line 1: Seems like both https://github.com/open-telemetry/opentelemetry-proto/blob/main/README.md#maturity-level and https://github.com/open-telemetry/opamp-spec/blob/main/specification.md?plain=1#L3 refer to this.

tigrannajaryan: Yes, if we remove this we need to replace it by something that the OTLP and OpAMP protocols can refer to.

trask: I brought back the maturity levels, but left the (outdated) matrix deleted

arminru (approved):

tigrannajaryan (approved):

dyladan (approved):

trask:
TODO post in #otel-comms

PR #1547, opened 2023-06-14
Define process for requesting TC decision, clarify TC voting process

jack-berg:
Attempting to capture the discussion in the 6/14/23 TC meeting. 

cc @open-telemetry/technical-committee.


tech-committee-charter.md
```diff
@@ -100,7 +100,13 @@ The TC follows a [Consensus Seeking](https://en.wikipedia.org/wiki/Consensus-see
 
 If an agenda item cannot reach a consensus a TC member can call for either a closing vote or a vote to table the issue to the next meeting. The call for a vote must be seconded by a majority of the TC or else the discussion will continue.
 
-For all votes, a simple majority of all TC members for, or against, the issue wins. A TC member may choose to participate in any vote through abstention.
+For all votes, a simple majority of all TC members for, or against, the issue wins. A TC member may choose to participate in any vote through abstention. TC votes may be done synchronously or asynchronously. Asynchronous votes remain open for a period of time agreed upon by the TC when the issue is put to a vote. Aggregated TC vote results are shared publicly but the vote selections of individual members are not shared. The TC decides whether vote selections by members are anonymous or shared among members when the issue is put to a vote.
+
+### Requesting a TC decision
+
+When a project issue fails to reach consensus, participants may request that the TC make a decision.
```
reyang, line 107: Are we okay with any participants or we want to scope this to project leads for now? I'm considering the situation where the SIG reached consensus among majority of folks while a single participant wanted to escalate to the TC.

reyang (approved):
LGTM with a non-blocking comment.

tech-committee-charter.md
```diff
@@ -100,7 +100,13 @@ The TC follows a [Consensus Seeking](https://en.wikipedia.org/wiki/Consensus-see
 
 If an agenda item cannot reach a consensus a TC member can call for either a closing vote or a vote to table the issue to the next meeting. The call for a vote must be seconded by a majority of the TC or else the discussion will continue.
 
-For all votes, a simple majority of all TC members for, or against, the issue wins. A TC member may choose to participate in any vote through abstention.
+For all votes, a simple majority of all TC members for, or against, the issue wins. A TC member may choose to participate in any vote through abstention. TC votes may be done synchronously or asynchronously. Asynchronous votes remain open for a period of time agreed upon by the TC when the issue is put to a vote. Aggregated TC vote results are shared publicly but the vote selections of individual members are not shared. The TC decides whether vote selections by members are anonymous or shared among members when the issue is put to a vote.
```
yurishkuro, line 103: > a simple majority of all TC members

I think this is a bit ambiguous: majority of all TC, or majority of those who voted? If the latter, then there should be a quorum requirement.

tech-committee-charter.md
jack-berg: I think minimally we should scope it to otel community members. Further narrowing to project leads seems helpful if there's a simple definition for what a project lead is. Could say that it has to be 1 or 2 maintainers / approvers.

tigrannajaryan (approved):

tech-committee-charter.md
arminru: Agreed!

carlosalberto (approved):

tech-committee-charter.md
```diff
@@ -100,7 +100,13 @@ The TC follows a [Consensus Seeking](https://en.wikipedia.org/wiki/Consensus-see
 
 If an agenda item cannot reach a consensus a TC member can call for either a closing vote or a vote to table the issue to the next meeting. The call for a vote must be seconded by a majority of the TC or else the discussion will continue.
 
-For all votes, a simple majority of all TC members for, or against, the issue wins. A TC member may choose to participate in any vote through abstention.
+For all votes, a simple majority of a quorum of TC members for, or against, the issue wins. A TC member may choose to participate in any vote through abstention. A quorum is defined as > 1/2 of TC members voting for or against (members who abstain do not count towards the quorum). TC votes may be done synchronously or asynchronously. Asynchronous votes remain open for a period of time agreed upon by the TC when the issue is put to a vote. Aggregated TC vote results are shared publicly but the vote selections of individual members are not shared. The TC decides whether vote selections by members are anonymous or shared among members when the issue is put to a vote.
```
yurishkuro, line 103: > members who abstain do not count towards the quorum

What is the motivation for this? in the TC discussion the feeling was that maybe abstaining should still count as a form of a vote and be part of the quorum.

tech-committee-charter.md
```diff
@@ -100,7 +100,13 @@ The TC follows a [Consensus Seeking](https://en.wikipedia.org/wiki/Consensus-see
...
+
+### Requesting a TC decision
+
+When a project issue fails to reach consensus, OpenTelemetry [community members](./community-membership.md#member) may request that the TC make a decision. If the issue falls under a project with assigned [approvers](./community-membership.md#approver) and / or [maintainers](./community-membership.md#maintainer), at least two of those members should agree to and coordinate requesting a TC decision.
+
+The request for a TC decision must be made by a comment on a public issue. The comment must explicitly tag the TC team (@open-telemetry/technical-committee), and summarize the various options and their relative tradeoffs. A TC member should review the summary and may request additional details or other changes to more accurately frame the issue. The request will then be added to the TC agenda, and the TC will work towards a decision using the [TC voting process](#voting-on-project-issues).
```
yurishkuro, line 109: Do we want to specify a process for TC to decline to vote on an issue? For example, SCOTUS requires 4 out of 9 justices to vote yes to accept a case. I don't think TC should be obligated to take any conceivable disagreement that might arise in some SIG.

tech-committee-charter.md
jack-berg: That was just me misinterpreting the quorum [suggestion here](https://github.com/open-telemetry/community/pull/1547#discussion_r1230234126). I've pushed a commit which counts abstaining towards quorum.

tech-committee-charter.md
jack-berg: I'm open to this but don't want to be over specify the process. I think the "A TC member should review the summary and may request additional details or other changes to more accurately frame the issue" leaves room for the TC to reject the request, but maybe you could suggest some phrasing that makes that more explicit?

tech-committee-charter.md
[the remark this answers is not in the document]
jack-berg: I pushed a commit which counts abstaining towards the quorum requirement. This means we could scenarios like:

1. 4 members abstain, 1 member votes for option a: option a wins
2. 3 members abstain, 1 member votes for option a, 1 member votes for option b: the vote is not resolved until somebody breaks the tie by voting instead of abstaining, or an additional member casts their vote in favor of one of the options

So in scenario 2 there is still the possibility for an indeterminant vote, but I think that will be rare enough that we can work it out among ourselves.

yurishkuro (approved):

jack-berg:
@open-telemetry/governance-committee this is ready to merge.

Issue #1523, opened 2023-06-02
Register opentelemetry as a Pypi organization


ocelotl:
Pypi now supports [organizations](https://blog.pypi.org/posts/2023-04-23-introducing-pypi-organizations/). It would be a good idea to register an organization for our project.

It would probably be better to register this using [this user](https://pypi.org/user/opentelemetry/).

@open-telemetry/python-approvers
@open-telemetry/technical-committee

trask:
hi @ocelotl! do you happen to know who owns the [PyPI opentelemetry user](https://pypi.org/user/opentelemetry/)? I don't see it listed on our [assets page](https://github.com/open-telemetry/community/blob/main/assets.md).

lzchen:
@trask 
@codeboten Possibly you "own" it?

codeboten:
I do, I submitted a request for a new organization:
<img width="429" alt="Screenshot 2023-06-20 at 11 09 56 AM" src="https://github.com/open-telemetry/community/assets/223565/6e86aee3-841c-4661-9d81-0d924ecb3dc5">

trask:
@codeboten did the request go through? thx

codeboten:
It is still currently pending ![Screenshot 2023-07-06 at 11 28 04 AM](https://github.com/open-telemetry/community/assets/223565/86929b17-c4ea-48f1-93a7-42f07466438e)

bogdandrutu:
@codeboten any update?

codeboten:
i don't have access to the pypi account.... maybe @ocelotl or @lzchen can give an update

lzchen:
Still pending

<img width="480" alt="image" src="https://github.com/open-telemetry/community/assets/11580155/06271f66-f820-40e3-8dc0-07a8ab943e30">

trask:
@lzchen any luck?

lzchen:
@trask 

Still same status. We will try to reach out to Pypi representatives to see if we can streamline the process.

lzchen:
Update on this, Pypi has approved and created an organization for us, under `opentelemetry`! I will be adding all our packages under this org.

svrnm:
That's great news 🎉 

> https://pypi.org/user/opentelemetry/

are credentials for this user stored in 1password?

yurishkuro:
> are credentials for this user stored in 1password?

I don't see them.

lzchen:
@yurishkuro 

[...]

PR #1578, opened 2023-07-06
Add PyPI user to assets

trask:
Follow-up from https://github.com/open-telemetry/community/issues/1523#issuecomment-1599274558

cc @codeboten


assets.md
```diff
@@ -47,6 +47,12 @@ but these individual OpenTelemetry members have been given rights to publish und
 * [@jack-berg](https://github.com/jack-berg)
 * [@trask](https://github.com/trask)
 
+### PyPI
+
+Link: https://pypi.org/user/opentelemetry/
+
+- Owner: [@codeboten](https://github.com/codeboten)
```
arminru, line 54: Could we have a backup here, e.g., a user shared among the GC and stored in your 1Password account?

arminru (approved):

assets.md
codeboten: The username/password is shared in 1Password. What isn't shared is the MFA on the account. I've shared the recovery codes w/ @ocelotl

codeboten (approved):

Issue #1608, opened 2023-07-20
Can we remove iCalendar link that often times out in CI?


yurishkuro:
In the README there is a link to Google calendar in the form of `.ics` iCalendar file download. https://github.com/open-telemetry/community/blob/24d0472a852699f2dcd3766525568e2a66dcea54/README.md?plain=1#L67

This link is very slow, 10-15seconds, so it often times out in the CI and breaks the `markdown-link-check` linter
```
  [✖] https://calendar.google.com/calendar/ical/google.com_b79e3e90j7bbsa2n2p5an5lf60%40group.calendar.google.com/public/basic.ics → Status: 0
```
Since we already provide two other links to the community calendar, can we remove this one?  I don't see an option in `markdown-link-check` to configure timeout to be longer.

yurishkuro:
If we can make the linter ignore the link, it's better than removing it.

PR #1609, opened 2023-07-24
Ignore link

trask:
Related to #1608

I'm not opposed to removing the link altogether, but figured I'd send this in the meantime.


yurishkuro (approved):

yurishkuro:
I did not find a way to control the timeout of the linter.

PR #1729, opened 2023-10-12
[election] Fix number of open seats

djaglowski:
There are five governance committee seats expiring this year.

The [charter also specifies](https://github.com/open-telemetry/community/blob/main/governance-charter.md#establishment-of-a-governance-committee) that 2023 will fill five seats.


arminru:
@open-telemetry/otel-elections PTAL

PR #1764, opened 2023-11-01
Tweaks to the election's eligibility criteria and company representation

jpkrohling:
During this last election, we noticed a couple of places that could have better wording.

I believe this PR reflects what was discussed during the GC meeting that checked the eligibility of the candidates, and agreed to be made as suggestions only after the election has been completed.

### DIFF
--- a/governance-charter.md
+++ b/governance-charter.md
@@ -90,8 +90,9 @@ around a more robust idea of community membership.
 
 ### Eligibility for candidacy
 
-Anyone may nominate either themselves or someone else to be a candidate in the
-election.
+Anyone can be a candidate, nominated by either themselves or someone else,
+unless two other employees of the same company are already on the committee and
+not up for re-election (see [Maximal representation](#maximal-representation)).
 
 To be ratified, candidates need three endorsements in total from members of
 standing from three different companies. When self-nominating, a candidate that
@@ -111,9 +112,7 @@ be ambiguous on the intentions of the review or reactions.
 ### Eligibility for voting
 
 All Members of Standing are eligible to vote for the governance committee
-members, unless two other employees of the same company are already on the
-committee and not up for re-election
-(see [Maximal representation](#maximal-representation)).
+members.
 
 ### Election process
 
--- a/governance-charter.md
+++ b/governance-charter.md
@@ -104,6 +104,10 @@ Nominators are free to nominate as many people as they wish to. Members of
 Standing may endorse multiple nominees, but we expect endorsements to be in good
 faith. If this turns out to be a problem, this will be reconsidered.
 
+Endorsements should be comments containing the phrase "I support". Simply adding
+a code review or emoji reactions to the description is NOT sufficient and might
+be ambiguous on the intentions of the review or reactions.
+
 ### Eligibility for voting
 
 All Members of Standing are eligible to vote for the governance committee
@@ -119,16 +123,17 @@ to the respective positions.
 
 To encourage diversity there will be a maximum of two Governance Committee
 members from any one company at any time. If the outcomes of an election result
-in more than two candidates from the same company being elected, the lowest vote
-getters from any particular company will be removed.
+in more than two candidates from the same company being at the GC, the lowest vote
+getters from any particular company will be removed. 
 
 If representation changes because of job changes, acquisitions, or other events,
-sufficient members of the committee must resign until the maximum representation
-is respected. If it is impossible to find sufficient members to resign, the
-entire company’s representation will be removed and new special elections held.
-In the event of a question of company membership (for example evaluating
-independence of corporate subsidiaries) a majority of all non-involved
-Governance Committee members will decide.
+sufficient members of the committee must resign within 28 days until the maximum
+representation is respected. If it is impossible to find sufficient members to
+resign after that period, the entire company’s representation will be removed.
+In both cases a special election will be held for those position, as [outlined
+below](#special-elections). In the event of a question of company membership
+(for example evaluating independence of corporate subsidiaries) a majority of
+all non-involved Governance Committee members will decide.
 
 ### Special Elections
 
--- a/governance-charter.md
+++ b/governance-charter.md
@@ -90,9 +90,13 @@ around a more robust idea of community membership.
 
 ### Eligibility for candidacy
 
-Anyone can be a candidate, nominated by either themselves or someone else,
-unless two other employees of the same company are already on the committee and
-not up for re-election (see [Maximal representation](#maximal-representation)).
+Anyone can be a candidate nominated by either themselves or someone else unless
+two other employees of the same company are already on the committee and not up
+for re-election (see [Maximal representation](#maximal-representation)). In case
+enough employees of the same company get enough votes to be elected so that the
+maximal representation is not respected anymore, only the highest-voted
+candidate(s) among them will get elected, with the other candidates having their
+candidacies nullified.
 
 To be ratified, candidates need three endorsements in total from members of
 standing from three different companies. When self-nominating, a candidate that
--- a/governance-charter.md
+++ b/governance-charter.md
@@ -111,7 +111,8 @@ be ambiguous on the intentions of the review or reactions.
 ### Eligibility for voting
 
 All Members of Standing are eligible to vote for the governance committee
-members. 
+members, unless two other employees of the same company are already on the
+committee (see [Maximal representation](#maximal-representation)).
 
 ### Election process
 
@@ -122,9 +123,7 @@ to the respective positions.
 ### Maximal representation
 
 To encourage diversity there will be a maximum of two Governance Committee
-members from any one company at any time. If the outcomes of an election result
-in more than two candidates from the same company being at the GC, the lowest vote
-getters from any particular company will be removed. 
+members from any one company at any time.
 
 If representation changes because of job changes, acquisitions, or other events,
 sufficient members of the committee must resign within 28 days until the maximum
--- a/governance-charter.md
+++ b/governance-charter.md
@@ -91,9 +91,14 @@ around a more robust idea of community membership.
 ### Eligibility for candidacy
 
 Anyone may nominate either themselves or someone else to be a candidate in the
-election. To be ratified as a candidate, the nominee must accept the nomination
-and three Members of Standing (including the nominator, if she/he has standing)
-from three different employers, must endorse the nomination.
+election.
+
+To be ratified, candidates need three endorsements in total from members of
+standing from three different companies. When self-nominating, a candidate that
+is member of standing will need two more endorsements from members of standing
+from two other companies. When being nominated by a member of standing from a
+different company and accepting the nomination, the candidate that is a member
+of standing will have two endorsements already, needing only one more.
 
 Nominators are free to nominate as many people as they wish to. Members of
 Standing may endorse multiple nominees, but we expect endorsements to be in good
@@ -112,19 +117,17 @@ to the respective positions.
 
 ### Maximal representation
 
-To encourage diversity there will be a maximum of one-third representation on
-the Governance Committee from any one company at any time. If the outcomes of
-an election result in greater than 1/3 representation (or maximum of two,
-whichever is greater), the lowest vote getters from any particular company will
-be removed until representation on the committee is less than one-third.
+To encourage diversity there will be a maximum of two Governance Committee members from any one company at any time. If the outcomes of
+an election result in more than two candidates from the same company being elected, the lowest vote getters from any particular company will
+be removed.
 
-If percentages shift because of job changes, acquisitions, or other events,
-sufficient members of the committee must resign until max one-third
-representation is achieved. If it is impossible to find sufficient members to
-resign, the entire company’s representation will be removed and new special
-elections held. In the event of a question of company membership (for example
-evaluating independence of corporate subsidiaries) a majority of all
-non-involved Governance Committee members will decide. 
+If representation changes because of job changes, acquisitions, or other events,
+sufficient members of the committee must resign until the maximum representation
+is respected. If it is impossible to find sufficient members to resign, the
+entire company’s representation will be removed and new special elections held.
+In the event of a question of company membership (for example evaluating
+independence of corporate subsidiaries) a majority of all non-involved
+Governance Committee members will decide.
 
 ### Special Elections
 
--- a/governance-charter.md
+++ b/governance-charter.md
@@ -133,7 +133,7 @@ If representation changes because of job changes, acquisitions, or other events,
 sufficient members of the committee must resign within 28 days until the maximum
 representation is respected. If it is impossible to find sufficient members to
 resign after that period, the entire company’s representation will be removed.
-In both cases a special election will be held for those position, as [outlined
+In this case a special election will be held for those position, as [outlined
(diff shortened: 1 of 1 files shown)


trask (approved):

atoulme (approved):

arminru (approved):

mtwo:
@jpkrohling this changes the maximal representation from three members to two. Not that I'm against it (I don't think that we will have three from any single firm in the foreseeable future), but this does have an impact.

governance-charter.md
```diff
@@ -138,9 +138,14 @@ around a more robust idea of community membership.
...
-and three Members of Standing (including the nominator, if she/he has standing)
-from three different employers, must endorse the nomination.
+election.
+
+To be ratified, candidates need three endorsements in total from members of
+standing from three different companies. When self-nominating, a candidate that
+is member of standing will need two more endorsements from members of standing
+from two other companies. When being nominated by a member of standing from a
+different company and accepting the nomination, the candidate that is a member
+of standing will have two endorsements already, needing only one more.
```
svrnm, line 148: I think we should be clear how such an endorsement looks like, in the past elections PRs from self-nominations got "thumbs ups", comments with "I support" or review approvals and it's unclear to me what counts. Since this is what we do already for people being added to the OpenTelemetry GitHub Org, I suggest that a comment like "I support" or "I endorse" should do it.

governance-charter.md
```diff
@@ -159,19 +164,18 @@ to the respective positions.
...
-sufficient members of the committee must resign until max one-third
-representation is achieved. If it is impossible to find sufficient members to
-resign, the entire company’s representation will be removed and new special
-elections held. In the event of a question of company membership (for example
-evaluating independence of corporate subsidiaries) a majority of all
-non-involved Governance Committee members will decide. 
+To encourage diversity there will be a maximum of two Governance Committee
+members from any one company at any time. If the outcomes of an election result
+in more than two candidates from the same company being elected, the lowest vote
+getters from any particular company will be removed.
```
svrnm, line 170: What about the case that one (or two) employees from one company are already members of the GC, and one (or more) additional employees of that same company are elected? In the case of 2 existing members, are additional employees even _able_ to be elected? In the case of 1 existing member, and 2 additional being elected, is the same logic applied (the lowest vote getters will be removed)?

governance-charter.md
```diff
@@ -159,19 +164,18 @@ to the respective positions.
...
+
+If representation changes because of job changes, acquisitions, or other events,
+sufficient members of the committee must resign until the maximum representation
+is respected. If it is impossible to find sufficient members to resign, the
+entire company’s representation will be removed and new special elections held.
```
svrnm, line 175:
```suggestion
If representation changes because of job changes, acquisitions, or other events,
sufficient members of the committee must resign until the maximum representation
is respected. If it is impossible to find sufficient members to resign, the
entire company’s representation will be removed. In both cases a special election will be held for those position, 
as [outlined below](#special-elections)
```

The current wording creates the assumption that special elections are only held, if the members are removed (forcefully), but reading the section about Special Elections below, this is also the case when members resign voluntarily.

governance-charter.md
```diff
@@ -159,19 +164,18 @@ to the respective positions.
...
-evaluating independence of corporate subsidiaries) a majority of all
-non-involved Governance Committee members will decide. 
+To encourage diversity there will be a maximum of two Governance Committee
+members from any one company at any time. If the outcomes of an election result
+in more than two candidates from the same company being elected, the lowest vote
+getters from any particular company will be removed.
+
+If representation changes because of job changes, acquisitions, or other events,
+sufficient members of the committee must resign until the maximum representation
+is respected. If it is impossible to find sufficient members to resign, the
```
svrnm, line 174:
```suggestion
If representation changes because of job changes, acquisitions, or other events,
sufficient members of the committee must resign within 28 days until the maximum representation
is respected. If it is impossible to find sufficient members to resign after that period, the
```

I suggest to introduce a fixed period (vs a "as soon as possible"), this (a) allows the affected GC members to make up their mind and coordinate who stays & who resigns, (b) gives a clear end-date for when the removal will be applied which removes ambiguity when it will be applied (immediatly? After the remaining GC says "now"?) and (c) it may make things easier if upcoming elections are close.

I suggest days as a unit of measure since "weeks" or "months" have debatable length.

For the length of the period I think it shouldn't be a lot less than 21 days, since individuals may be away for 2+ weeks (long vacation, medical situation, etc.) and it allows to form a special election comittee in preparation already. I also think more than 42 days (~6 weeks) shouldn't be applied, to get this undesired state resolved quickly.

svrnm (commented):
A few non-blocking comments, that either can be resolved here or create some follow up PRs.

governance-charter.md
danielgblanco: +1 on making it clear. For the purpose of clarity and auditability, should that endorsement be a PR approval? Comments can be edited and not straight-forward to find their edit trail. PR approvals (even if the endorser is not an approver of the repo itself) cannot be modified. The comment in the approval can be changed, but not the approval itself after a PR is merged.

jpkrohling: We found PR approvals to be ambiguous, as the reviewer might be approving the markdown that is being added to the repository, not necessarily endorsing the candidate.

governance-charter.md
jpkrohling: I think the first sentence summarizes the spirit of the rule:

> To encourage diversity there will be a maximum of two Governance Committee
members from any one company at any time.

If the are two people already at the board, no further candidates can be elected from the same company without one of those resigning.

governance-charter.md
jpkrohling: I like the suggestion, incorporating.

governance-charter.md
jpkrohling: I added a sentence here about being explicit by adding "I approve" as a comment.

governance-charter.md
svrnm: > To encourage diversity there will be a maximum of two Governance Committee
members from any one company at any time. If the outcomes of an election result
in more than two candidates from the same company being at the GC, the lowest vote
getters from any particular company will be removed. 

Sorry, for being nit-picking, but while this wording is improved around the fact that there may be GC members already, it's not entirely clear what "the lowest vote getters from any particular company will be removed" means: is this only comparing votes of the current election, or also from previous elections? E.g. if a GC member got 100 votes in year 1 and then two new GC members are elected with 101 and 102 votes in year 2, is the member with 100 votes in year1 removed or the one with 101 votes from year 2? In the case of the existing GC member from year1 being removed (or resigning deliberately), will there be another special election or will one additional person be added to the GC in this election (and then only serve one year?)?

svrnm: A different proposal may be the following:

* Name GC members elected in odd years "Cohort 1" and those elected in even years "Cohort 2".
* In each cohort there must only be a maximum of one GC members from any one company at any time. 

In a case of an election there will be no more conflict with "existing" and "newly elected" members. In the case of a company change the existing rules may apply.

dyladan: I agree with @svrnm it should be clear that only the current election is affected. We should also think about what happens if there is a merge/acquisition that changes the GC composition outside of an election cycle. Do we let it ride until the elections naturally fix the issue, do we allow the employees of the company to decide amongst themselves who should step down and have a special election, or do we drop the lowest voted member of the most recent cohort?

trask:
> Before coming to this phrase, we all thought that the maximum was two, which is why I'm making this proposal to set it to two.

I was aware it was three with our current size, but I think ensuring more company diversity is worth reducing it to two.

dyladan (approved):

governance-charter.md
jpkrohling: Keeping track of old election results and assuming they are still accurate by the time a decision is made doesn't sound right to me. I'm also not very fond of having a max per cohort.

I believe the effective rule right now is that new members cannot be elected if two members from the same company are on the board already. I changed the wording on this doc to reflect that, but we can certainly discuss other alternatives if that's not the understanding of the majority.

svrnm: > it should be clear that only the current election is affected.

That's why I suggested the solution of having 1 person per company per election. It _might_ make things easier

> We should also think about what happens if there is a merge/acquisition that changes the GC composition outside of an election cycle. Do we let it ride until the elections naturally fix the issue, do we allow the employees of the company to decide amongst themselves who should step down and have a special election, or do we drop the lowest voted member of the most recent cohort?

There is a section about the below the election ("_If representation changes because of job changes, acquisitions, or other events, sufficient members of the committee must resign within 28 days until the maximum representation is respected._"). So it is left to the affected GC members to figure out a solution themselves.t.

jpkrohling:
@open-telemetry/governance-committee, in my opinion, there are enough changes here to warrant a super-majority approval of this. I can't seem to be able to request a new review from people who reviewed already, so, would you please take another look and give another approval if it looks OK to you?

governance-charter.md
svrnm: I wrote & send my last comment while yours arrived. I'll leave it for completeness but no need to respond.

> I believe the effective rule right now is that new members cannot be elected if two members from the same company are on the board already. I changed the wording on this doc to reflect that, but we can certainly discuss other alternatives if that's not the understanding of the majority.

While not _perfect_ this is a good solution (also compared to the other ones), although there are 2 clarifications that need to be made:

* It should be explicit, that those 2 existing members should not be up for re-election
* If there is one existing member of a company, and more than one (2+) get elected the rule with the "most votes" should still apply.

I can try to make some suggestions

governance-charter.md
```diff
@@ -112,19 +122,17 @@ to the respective positions.
...
-
-If percentages shift because of job changes, acquisitions, or other events,
-sufficient members of the committee must resign until max one-third
-representation is achieved. If it is impossible to find sufficient members to
-resign, the entire company’s representation will be removed and new special
-elections held. In the event of a question of company membership (for example
-evaluating independence of corporate subsidiaries) a majority of all
-non-involved Governance Committee members will decide. 
+To encourage diversity there will be a maximum of two Governance Committee
+members from any one company at any time.
```
svrnm, line 126:
```suggestion
To encourage diversity there will be a maximum of two Governance Committee
members from any one company at any time. Therefore, there are limits on the eligibility for employees from the same company (see [Eligibility for voting](#eligibility-for-voting)). 

If the outcomes of an election would result in more than two GC members from the same company the following rules will be applied to ensure maximal representation: In the case of one existing GC member of that company, who was not up for re-election, this existing member remains, and all except the highest vote getter from any particular company will be removed. In the case of no existing GC member of that company, all except the two highest vote getters from any particular company will be removed. 
```

governance-charter.md
```diff
@@ -112,19 +122,17 @@ to the respective positions.
```
svrnm, line 122:
```suggestion

In the case of a tie among candidates, which would lead to one candidate being elected and the others being not elected, a (special election)[#special-elections] will be held among those candidates affected by the tie.
```

Sorry for doing that, but while thinking about all things that could go wrong, I ran into the question of ties. Not sure if this should be part of this PR or this should be a follow-up, so feel free to drop this comment and I will do a follow-up, @jpkrohling

svrnm:
> @open-telemetry/governance-committee, in my opinion, there are enough changes here to warrant a super-majority approval of this. I can't seem to be able to request a new review from people who reviewed already, so, would you please take another look and give another approval if it looks OK to you?

I think, you can dismiss existing reviews by clicking the "..." -> Dismiss review and then re-ask for reviews. 

I wonder if we should introduce the following branch protection rule for the community repo:

>  Dismiss stale pull request approvals when new commits are pushed

governance-charter.md
jpkrohling: I think this should be part of a new PR, as it's a new topic.

governance-charter.md
jpkrohling: If people are not eligible when there are already two colleagues on the board, when would this outcome happen?

svrnm: If there are already 2 GC members of a company, yes, there is no problem.

But, if there is only 1 GC member of a company, and additional `n` employees of the same company run, there could be an outcome where there is 1 GC member and `n` new. Only one of `n` can get a seat then.

And even if there is 0 GC member of a company, there is still the possibility that `n` get elected and the top 2 of them get a seat then.

danielgblanco: It's probably common sense that the most voted will be the one getting a seat from the number of available seats per company (which could be 1 or 2), but I think it should be stipulated here in some way to avoid any doubt.

svrnm: Fair point, you could infer that from the rest of the election rules laid out, but being explicit might help to avoid future issues

governance-charter.md
```diff
@@ -91,18 +91,29 @@ around a more robust idea of community membership.
...
 
+Endorsements should be comments containing the phrase "I support". Simply adding
+a code review or emoji reactions to the description is NOT sufficient and might
+be ambiguous on the intentions of the review or reactions.
+
 ### Eligibility for voting
 
 All Members of Standing are eligible to vote for the governance committee
-members. 
+members, unless two other employees of the same company are already on the
```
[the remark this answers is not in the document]
svrnm: See this discussion: https://github.com/open-telemetry/community/pull/1764#discussion_r1384692333

tl;dr: no matter how we do it, the moment we introduce a limit with "2 people per company" we need to introduce some ruling to enforce this. And all of them add different layers of complexity. There were multiple suggestions:

* Fix it only via voting, which gets complicated, the moment you have 2 existing GC members but at least 1 new elected from the same company
* Fix it via a limit per year (1 member per company per year can be elected), this also leads to 2 seats max
* Fix it via the eligibility (what you see here)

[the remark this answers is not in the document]
jpkrohling: Fixed by moving to the right section.

governance-charter.md
svrnm: solved by the https://github.com/open-telemetry/community/pull/1764/commits/73b3e9d0ead0edb39567f8e46b62fe1d9bd99428

svrnm (approved):

svrnm:
LGTM

danielgblanco (approved):

dyladan (approved):

governance-charter.md
```diff
@@ -112,19 +126,17 @@ to the respective positions.
...
-evaluating independence of corporate subsidiaries) a majority of all
-non-involved Governance Committee members will decide. 
+To encourage diversity there will be a maximum of two Governance Committee
+members from any one company at any time.
+
+If representation changes because of job changes, acquisitions, or other events,
+sufficient members of the committee must resign within 28 days until the maximum
+representation is respected. If it is impossible to find sufficient members to
+resign after that period, the entire company’s representation will be removed.
+In both cases a special election will be held for those position, as [outlined
```
trask, line 136:
```suggestion
In this case a special election will be held for those position, as [outlined
```

trask (approved):

austinlparker (approved):

svrnm:
With 5 approvals (+ @jpkrohling as the author of this PR) we have a super-majority (9*(2/3)=6) and this can be merged (after applying https://github.com/open-telemetry/community/pull/1764#discussion_r1425657651)

svrnm:
🎉 Thank you @jpkrohling!

PR #1776, opened 2023-11-03
Fix headings and their anchors in mission-vision-values.md

chalin:
/cc @open-telemetry/docs-approvers @trask

### DIFF
--- a/mission-vision-values.md
+++ b/mission-vision-values.md
@@ -12,12 +12,12 @@ weight: -10
 
 # OpenTelemetry mission, vision, and values
 
-## Mission &mdash; our overall north star as a community
+## Mission &mdash; our overall north star as a community {#mission}
 
 OpenTelemetry's Mission: **to enable effective observability by making
 high-quality, portable telemetry ubiquitous.**
 
-## Vision &mdash; the world we imagine for OTel end-users
+## Vision &mdash; the world we imagine for OTel end-users {#vision}
 
 Effective observability is powerful because it enables developers to innovate
 faster while maintaining high reliability. But effective observability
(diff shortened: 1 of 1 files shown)


arminru (approved):

mission-vision-values.md
```diff
@@ -12,12 +12,12 @@ weight: -10
 
 # OpenTelemetry mission, vision, and values
 
-## Mission &mdash; our overall north star as a community
+## Mission &mdash; our overall north star as a community {#mission}
```
trask, line 15: is there a way to do this without it showing up when viewed in the community repo? (https://github.com/chalin/otel-community/blob/chalin-im-mission-fix-heading-anchors-2023-11-03/mission-vision-values.md#mission--our-overall-north-star-as-a-community-mission)

chalin: The canonical URL for pages that are published on the OTel website are website URLs: so for this file, it's https://opentelemetry.io/community/mission/. Folks shouldn't be using the repo to view pages published via the website. So in that sense, it shouldn't really be a problem to see `{#mission}` when viewing the GH-rendered file, right?

But, to answer your question, the simplest way to achieve the same effect is to change the heading to "Mission", and move the rest of the heading text to somewhere else -- e.g., a blockquote at the start of the section. I wanted to keep this update simple. Let me know what you think.

/cc @cartermp @svrnm

arminru: Good point, @trask.
I think some might still view the MD files directly in the repo. For that, simplifying the headers and putting the subtitle just below and maybe with some emphasis should be fine.

svrnm: +1 for simplifying the headers and putting the subtitles below

chalin: I agree that that's the best solution. Done!

chalin:
Hi all: here's what it looks like with the "subheadings":

> image

svrnm (approved):

danielgblanco (approved):

svrnm:
@chalin, do we need to update the git submodule for the community forum at opentelemetry.io then?

chalin:
> @chalin, do we need to update the git submodule for the community

Yes:

- https://github.com/open-telemetry/opentelemetry.io/pull/3876

Issue #1596, opened 2023-07-14
[Proposal] Use an infrastructure-as-code solution to manage the open-telemetry github


jaronoff97:
Hello all, after some of the past week's challenges around repository maintenance, some operator contributors had the idea that we could use Github's official terraform provider to provision and manage the various SIG github groups, repositories, branch protection rules, etc. that the community repo is currently used to manage entirely manually. This would then allow the TC to just approve and merge PRs that change the governance of the open-telemetry github rather than needing to make a slew of manual changes. 

Benefits
-----------------
* Organization-as-code for OpenTelemetry would enable self-describing SIGs and repos that are entirely clear to maintainers
* TC can automatically ping code owners when repository maintenance is needed
* Maintainers that need repository maintenance can make the changes themselves
* TC members who are not familiar with the github UI can focus solely on the requested change
* (i'm sure there are more, these are the ones I just thought of initially)

Risks
-----------------
* Will take some time to migrate us to this new strategy (this can be done piecemeal through terraform by importing resources as needed)
* Operational overhead
  * We will need someone to set up this terraform code including CI/CD
  * I think we should have the expertise within the organization to make this happen
* If any of the repository information isn't meant to be public, this may be a challenge
  * I'm not sure what wouldn't be okay being public
* Secret management will still need to be manual
  * This is no change from today, and we already don't have to update / manage these often right now

Design
-----------------
* Phase 1 (team membership)
  * Community would contain a new folder that holds its terraform configuration
  * All members of the org are imported via the [Membership resource](https://registry.terraform.io/providers/integrations/github/latest/docs/resources/membership)
  * CI/CD is automatically applying the state of the repository to ensure that the repo's state is always in-sync with the code 
  * At this point, all membership requests can now be made solely via terraform
* Phase 2 (SIGs)
  * A terraform module for OpenTelemetrySIG is created that can be used to create and manage the membership and repositories for a SIG
    * A module is used here to abstract the logical grouping of a SIG within the OpenTelemetry community
  * Each SIG's groups can be imported as teams and team members [[doc](https://registry.terraform.io/providers/integrations/github/latest/docs/resources/team_members)]
* Phase 3 (Repositories)
  * SIGs can opt-in initially to import their [repository](https://registry.terraform.io/providers/integrations/github/latest/docs/resources/repository) configuration to the SIG repository, some SIGs may choose to initially opt-out if they have special use cases
[...]

PR #1793, opened 2023-11-14
Manage membership thru IaC

austinlparker:
Per #1596, I wanted to come up with a quick spike of what this would look like. This does have a slight flaw that we need to correct -- it doesn't have child teams right now. I'm also not entirely sure what'll happen when we run it...? Will it delete/re-create all the teams?

Anyway, wanted to get some code here for people to look at and bikeshed about. :)


svrnm:
> I'm also not entirely sure what'll happen when we run it...? Will it delete/re-create all the teams?

I would guess that it does look at the current state and sync it to what ever desired state you want, isn't that how terraform is suposed to work? Maybe we can create some fake organisation to do some testing/trials and if we are satisfied with the results we can apply them on the real organisation?

Looking through the groups, this _may_ also be an opportunity to bring some structure into it, right now there are some SIGs that have teams with sub-teams, some do not have that structure, etc.
 
Is it possible to split `iac/membership/terraform.tfvars.json` into multiple files? At least put members and teams into their own files, but maybe even have a per-SIG structure, which we could combine with a CODEOWNERS entry (e.g. if there is a dotnet-teams.tfvars.json code-owned by dotnet-maintainers, the right maintainers (+TC+GC) get notified and asked for approval)

svrnm:
> > Is it possible to split `iac/membership/terraform.tfvars.json` into multiple files? At least put members and teams into their own files, but maybe even have a per-SIG structure, which we could combine with a CODEOWNERS entry (e.g. if there is a dotnet-teams.tfvars.json code-owned by dotnet-maintainers, the right maintainers (+TC+GC) get notified and asked for approval)
> 
> yes, although that does add an additional bit of baggage when you want to add a new one.
> 
> my 0.02 is that the goal would be two files; one that is just members, and one that is SIGs/repos/whatever.

If it adds to much complexity, I think the added value (having SIGs in the CODEOWNER for those files) is not big enough, so 2 files sounds good to me :thumbsup:

yurishkuro:
I am hoping in the final version the input files can be YAML, not JSON.

iac/membership/main.tf
```diff
@@ -0,0 +1,23 @@
+resource "github_membership" "member" {
```
yurishkuro, line 1: nit: I would put this under `tools/...`, e.g. `tools/org_settings`, I think `iac` is unnecessarily cryptic

yurishkuro:
Design question - on the original ticket someone mentioned that TF needs a state storage. Is that true? What's the approach we want to take w.r.t. such storage? I think it would be ideal if the whole thing could run on standard GH runners, not via some additional CNCF compute (which is harder to manage). A storage on local disk committed back to the repo could work (would need a different repo then).

danielgblanco:
Regarding what would happen if we run this: my guess is that it'd fail to create teams and members because those already exist with the same names/GH IDs. Starting from scratch, Terraform state will contain nothing so there's nothing that Terraform is tracking internally. When it does the refresh at the beginning of a `plan` or `apply` it will only refresh state from GH to whatever is in the state, which will be empty at that point.

To import resources into the tf state (which as @austinlparker mentioned we'd preferably manage with a SaaS), we'd have to do an [terraform import](https://developer.hashicorp.com/terraform/cli/import) from the current members and teams into the new empty state. After that point, Terraform will ensure that things are kept in sync.

See the import for teams for example which can work with team name or ID

https://registry.terraform.io/providers/integrations/github/latest/docs/resources/team#import

jaronoff97:
In theory the auto files go away after this initial PR though right? In the future, any new repos, teams, memberships should only be generated through terraform. I would also recommend maybe making some modules to manage these resources, that way you could define a user with their memberships in one place, not two. I wouldn't recommend import blocks, a script to go through it all is going to be your best bet.

iac/membership/main.tf
```diff
@@ -0,0 +1,23 @@
...
+  for_each = toset(var.members)
+
+  username = each.value
+  role     = "member"  // or "admin"
```
jaronoff97, line 5: creating a module where the variables would be:
```
username = var.username
role = var.role // default role = member
team_memberships = [
    {
         name = "<team>",
         role = "<member> // default is member
    }, ...
]
```

jaronoff97:
@austinlparker Sounds good to me RE: testing on a dummy org, lmk if there are extra changes we should make. Also, it looks like we need to copy the output JSON to auto.tfvars, right now there are no members or owners set.

svrnm:
what's needed to push this forward? I still think implementing this will have a lot of benefits!

tools/membership/main.tf
```diff
@@ -0,0 +1,502 @@
...
+    source = "./modules/sig"
+    name = "collector-contrib"
+    triagers = ["frzifus", "crobert-1", "gbbr", "JaredTan95", "mwear", "Frapschen", "gouthamve", "bryan-aguilar"]
+    approvers = ["atoulme", "MovieStoreGuy", "dmitryax", "bogdandrutu", "TylerHelmuth", "astencel-sumo", "codeboten", "mx-psi", "Aneurysm9", "evan-bradley", "songy23", "fatsheep9146", "dashpole", "jpkrohling", "djaglowski"]
+    maintainers = []
+}
+
+module "collector-contrib-maintainer_wg" {
+    source = "./modules/wg"
+    name = "collector-contrib-maintainer"
```
jpkrohling, line 168: What's the difference between this and the previous one?

tools/org_helpers/getmembers.py
```diff
@@ -0,0 +1,156 @@
+import requests
+import json
+import subprocess
+import pprint
+
+# Replace these variables with your GitHub organization and personal access token
+ORG_NAME = 'open-telemetry'
+ACCESS_TOKEN = subprocess.run(["gh", "auth", "token"], capture_output=True).stdout.decode('utf-8').strip()
```
jpkrohling, line 8: is it possible to use an env var, and if missing, call gh auth token?

tools/org_helpers/getmembers.py
```diff
@@ -0,0 +1,156 @@
...
+
+def print_group_imports(group, member_group, suffix):
+    group_name = group["name"]
+    print(f"tofu import 'module.{group_name}_{suffix}.github_team.{member_group}' {group_name}-{member_group}")
+    # singular = member_group[:-1]
+    # for m in group[member_group]:
+    #     print(f"tofu import 'module.{group_name}_{suffix}.github_team_membership.sig_{singular}[\"{m}\"]' {group_name}-{member_group}:{m}")
+
+# GENERATE TF FILE
+with open('output.tf', 'w') as f:
```
jpkrohling, line 137: Nice, I was wondering what was the origin of the members file :-) Perhaps it needs an update? Would this script be around once we make the switch? 

Small nit: can we order the modules alphabetically?

tools/org_helpers/output.tf
```diff
@@ -0,0 +1,502 @@
+
+module "memberships" {
```
jpkrohling, line 2: Is this a duplicate of tools/membership/main.tf?

tools/membership/main.tf
```diff
@@ -0,0 +1,502 @@
...
+    triagers = ["rbailey7210", "andrewhsu"]
+    approvers = ["tigrannajaryan", "reyang", "jack-berg", "arminru", "bogdandrutu", "jsuereth", "jmacd", "yurishkuro", "carlosalberto"]
+    maintainers = []
+}
+
+module "ruby_sig" {
+    source = "./modules/sig"
+    name = "ruby"
+    triagers = []
+    approvers = ["robbkidd", "ericmustin", "arielvalentin", "ahayworth", "plantfansam"]
```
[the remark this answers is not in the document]
kaylareopelle: Thanks for calling this out, @jpkrohling! 

I'm not really familiar with who counts as a SIG member. Should the list include everyone with a specific role on the opentelemetry-ruby repo? Should it also include folks with roles in the opentelemetry-ruby-contrib repo?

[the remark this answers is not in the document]
austinlparker: This PR is very old and doesn't include changes made to members/groups over the past several months; Don't worry, people won't get missed.

PR #1813, opened 2023-11-21
Remove special note from GC charter elections policy

svrnm:
There are no continuity members in the GC anymore, so this special note served it's purpose and can be safely removed.


arminru (approved):

jpkrohling (approved):

jpkrohling:
I think this whole doc needs a bit of love. There are some constructions in the future, but the future is now past. That said, this change LGTM and is ready to be merged.

PR #1687, opened 2023-09-21
Proposal for SIG OpenTelemetry on Mainframe

rrschulze:
This is the proposal for a work group and project focusing on enhancing OpenTelemetry to support the Mainframe. For details, please see the proposal in projects/mainframe.md.

Resolves #1686.

PR #1854, opened 2024-01-10
Added unknown words from projects/mainframe.md to .cspell.yaml

rrschulze:
Unknown words:
REXX
rrschulze
youngaaronm
Nale
davdai
jimporell
msomasu
jackjia


trask:
can you make this change instead directly in #1687?

rrschulze:
Making change under #1687 .

PR #1917, opened 2024-01-29
Updated maintainers meeting doc link

mtwo:
Linked to new notes doc (which itself links to archived notes). We switched to a new document as the old was was so long that it was causing performance issues.


trask (commented):
can you transfer ownership to the GC and give public access?

mtwo:
Both docs are owned by the GC account, I just forgot to make the link public! Fixed!

trask (approved):

arminru (approved):

PR #1932, opened 2024-02-08
Amend GC charter to include project management

tedsuo:
Clarifies the responsibilities that the GC has to the community regarding project management and maintaining SIG health.

### DIFF
--- a/governance-charter.md
+++ b/governance-charter.md
@@ -37,8 +37,10 @@ following:
   the project](https://github.com/cncf/toc/blob/main/proposals/sandbox/opentelemetry.adoc)
 - Define, evolve, and defend the **vision, values, mission, and scope** of the
   project - to establish and maintain the soul of OpenTelemetry.
-- Decide how and when **official releases** of OpenTelemetry artifacts are
-  made and what they include.
+- Define and execute the roadmap for OpenTelemetry, as defined in the project
+  management section below.
+- Ensure that SIGs maintain a healthy pool of maintainers and contributors and
+  remain focused on their feature roadmap and deliverables.
 - Establish processes regarding, and provide **a final escalation path** for any
   contested OpenTelemetry related decision.
 - Establish processes regarding other **project resources/assets**, including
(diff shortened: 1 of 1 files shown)


governance-charter.md
```diff
@@ -51,6 +53,49 @@ following:
...
+document describes the process for creating and executing projects of this size.
+
+The Governance Committee is responsible for reviewing project proposals
+and ensuring that they have clear goals and are staffed sufficiently to
+deliver on an agreed upon deadline. The Governance Committee is also
+responsible for deciding if a project is out of scope for the OpenTelemetry
+project.
+
+For proposals that involve [donating](https://github.com/open-telemetry/community/blob/main/CONTRIBUTING.md#donations)
+codebases to OpenTelemetry, the Governance Committee is responsible for
```
svrnm, line 75:
```suggestion
code bases to OpenTelemetry, the Governance Committee is responsible for
```

Of we add it to the known words in the cspell config, what ever works best.

svrnm (approved):
Only some non-blocking copy-edits. LGTM!

governance-charter.md
```diff
@@ -51,6 +53,49 @@ following:
...
 project groups like the [Technical Committee](./tech-committee-charter.md).
 
+### Project Management
+
+**Develop project proposals**  
+Efforts and initiatives that are too large, detailed, or important to be
+resolved in an issue or pull request are referred to as projects. These
+larger efforts often require collaboration between subject matter experts,
+the Technical Committee, and OpenTelemetry community members, and involve
+the creation of either a temporary or The Governance Committee is responsible
```
trask, line 63: I think there may be a word or two missing on this line

governance-charter.md
```diff
@@ -51,6 +53,49 @@ following:
...
+project board and the public roadmap stay up to date and accurately
+reflect reality.
+
+**Regular check-ins with SIG leads**  
+The Governance Committee must establish and publish a process whereby
+every maintainer and project lead has a check in with a Governance Committee
+member on a regular basis. During this check in, any project files or roadmap
+entries should be reviewed to verify that they are still accurate. If a SIG
+lead has any private concerns they wish to raise with the Governance Committee,
+this is an opportunity to do so.
```
trask, line 97: If a SIG has 3 maintainers, is the desire to have 3 separate 1-1 meetings, or a single meeting with all 3? Also, is it ok for this check-in to be done during a weekly SIG meeting assuming no private concerns? Or is the desire to ensure there's a private check-in to be more proactive about seeking potentially private concerns?

tedsuo: I was thinking that it would be separate 1:1 meetings, as there may be conflict between maintainers.

tedsuo: (I would suggest the check in just be a private DM)

tedsuo: (But I also don't want to define the implementation details directly in the GC charter)

governance-charter.md
```diff
@@ -51,6 +53,50 @@ following:
...
 project groups like the [Technical Committee](./tech-committee-charter.md).
 
+### Project Management
+
+#### Develop project proposals
+Efforts and initiatives that are too large, detailed, or important to be
+resolved in an issue or pull request are referred to as *projects*. These
+larger efforts often require collaboration between subject matter experts,
+the Technical Committee, and OpenTelemetry community members, and involve
+the creation of either a temporary or permanent SIG to accomplish their goals.
```
dyladan, line 63: I don't think there is any such thing as a "temporary SIG." I would change this to "working group"

governance-charter.md
trask: makes sense 👍

governance-charter.md
```diff
@@ -51,6 +53,50 @@ following:
...
+The Governance Committee is responsible for reviewing project proposals and
+ensuring  to accomplish their goals. The [project management](https://github.com/open-telemetry/community/blob/main/project-management.md)
```
trask, line 65:
```suggestion
ensuring they accomplish their goals. The [project management](https://github.com/open-telemetry/community/blob/main/project-management.md)
```

governance-charter.md
```diff
@@ -51,6 +53,50 @@ following:
...
+document describes the process for creating and executing projects of this size.
+
+The GC is responsible for reviewing project proposals, aligning their execution
+with the wider strategy and end-user priorities, and ensuring that they have
+clear goals and are staffed sufficiently to deliver on an agreed upon deadline.
+The Governance Committee is also responsible for deciding if a project is out of
+scope for the OpenTelemetry project.
+
+For proposals that involve [donating](https://github.com/open-telemetry/community/blob/main/CONTRIBUTING.md#donations)
```
dyladan, line 74: Should we have a specific deliverable for this? A template maybe for a report we should fill out? Is there any set deadline for reviewing proposals?

trask (approved):
❤️

governance-charter.md
```diff
@@ -37,8 +37,10 @@ following:
   the project](https://github.com/cncf/toc/blob/main/proposals/sandbox/opentelemetry.adoc)
 - Define, evolve, and defend the **vision, values, mission, and scope** of the
   project - to establish and maintain the soul of OpenTelemetry.
-- Decide how and when **official releases** of OpenTelemetry artifacts are
-  made and what they include.
+- Define and execute the roadmap for OpenTelemetry, as defined in the project
```
mtwo, line 40:
```suggestion
- Define and prioritize the roadmap for OpenTelemetry, as defined in the project
```

Prioritize might be better than execute, given that the GC generally isn't implementing most features. We're also not people's bosses; the best that we can do is ask that they focus their efforts on certain things.

governance-charter.md
mtwo: We should probably have a template for these reports (if only so that future GCs don't spend time re-inventing the process), though I don't think that it'll need to be linked here

danielgblanco (approved):

governance-charter.md
tedsuo: This whole line is actually a duplicate of the next paragraph, which has better wording. I've removed it.

PR #1955, opened 2024-02-20
SIG checkin process

svrnm:
This PR proposes a process for the "SIG checkin" as outlined in the GC charter.


project-management.md
```diff
@@ -36,3 +37,23 @@ The project lifecycle is as follows:
...
+## Project Check-Ins
+
+As defined in the [charter of the GC](./governance-charter.md#regular-check-ins-with-sig-leads) the GC will check in with maintainers and project leads on a regular basis. During this check in,
+
+* any project files or roadmap entries should be reviewed to verify that they are still accurate.
+* If a SIG lead has any private concerns they wish to raise with the Governance Committee, this is an opportunity to do so.
+
+The check in process is as follows:
+
+* For each project a GC member is appointed as a liaison person. The names of that person will be listed publicly.
```
trask, line 50:
```suggestion
* For each project a GC member is appointed as a liaison person. The name of that person will be listed publicly.
```

project-management.md
```diff
@@ -36,3 +37,23 @@ The project lifecycle is as follows:
...
+* If a GC member is the maintainer/project lead themselves, they are by default the liaison person for this project.
+* The liaison person is appointed for the duration of the project lifecycle, or until they do not get reelected into the GC. Outside of that regular duration, a liaison person can resign by providing a named successor.
+* The liaison person is responsible to check in with the maintainers/project leads at least once per month. This can either happen via a (private, non recorded) meeting or through a text-based conversation initiated by the liaison person.
+* The liaison person is responsible to bring the concerns of the SIG leads forward to the GC, and they are responsible for ensuring that all action items agreed upon during that check in are logged as issues on the project repository or community repository.
```
trask, line 54: I tend to prefer that the SIG leads open these issues, unless it's something sensitive and they wish to remain anonymous

danielgblanco: I would agree in principle to that, but do we have a documented way for this to happen? Nothing in the maintainers responsibilities [here](https://github.com/svrnm/community/blob/sig-checkin-process/community-membership.md#responsibilities-and-privileges-3) mentions scaling SIG health issues to the GC (only technical issues to the GC). Should we add something there instead? Do we propose a channel for it?

danielgblanco: Just to clarify: I'm not saying that we only rely on SIG leads raising issues with the GC to assess health. I fully support the quarterly health update below, driven by the liaison as well as maintainers.

trask: re-reading, I'm good with the current language:

> responsible for ensuring that all action items agreed upon during that check in are logged as issues

"responsible for" doesn't necessarily mean logging the issue if it's something that the SIG leads can do

trask (approved):

austinlparker (approved):

svrnm:
[76768d7](https://github.com/open-telemetry/community/pull/1955/commits/76768d752ca666008a462898f0c35423d4f80131) addresses the feedback from the last GC meeting:

* use short and long term SIGs / use one word vs WG/SIG
* Separate document (“GC checkins”)
* use the term “GC liaison” instead of “liaison person”

I did not address the following:

* "_Increasing the number of maintainers, actively engage around who is doing what (approvers, maintainers, triagers, etc.)_": I added a broader item to the list of what the check in contains, "the health of the SIG will be reviewed, and what measures can be applied to help the SIG". I think we can make this wording clearer when we have a better understanding of what needs to be done
* Define SIG Charter (every 2 years?): (cc @dyladan) although this is an important point I didn't want to just put it into the document with more context. We should take some extra time to address this separately

@open-telemetry/governance-committee PTAL

trask (approved):

gc-check-ins.md
```diff
@@ -0,0 +1,20 @@
...
+As defined in the [charter of the GC](./governance-charter.md#regular-check-ins-with-sig-leads) the GC will check in with maintainers of all long-term and short-term SIGs on a regular basis. During this check in,
+
+* any project files or roadmap entries should be reviewed to verify that they are still accurate.
+* if a SIG lead has any private concerns they wish to raise with the Governance Committee, this is an opportunity to do so.
+* the health of the SIG will be reviewed, and what measures can be applied to help the SIG.
+
+The check in process is as follows:
+
+* For each SIG a GC member is appointed as a GC liaison. The name of that person will be listed publicly.
+* If a GC member is the maintainer themselves, they are by default the GC liaison for this SIG.
```
jpkrohling, line 12: Newer SIGs follow the requirement to have at least two TC/GC sponsors, such as the Security SIG. Should the GC sponsor be the liaison in that case?

gc-check-ins.md
```diff
@@ -0,0 +1,20 @@
...
+* The GC liaison is appointed for the duration of the project lifecycle, or until they do not get reelected into the GC. Outside of that regular duration, a GC liaison can resign by providing a named successor.
+* The GC liaison is responsible to check in with the maintainers at least once per month. This can either happen via a (private, non recorded) meeting or through a text-based conversation initiated by the GC liaison.
```
jpkrohling, line 14: We kind of touched the idea that WGs and SIGs are the same for the purposes of this proposal. In that case, what to do with WGs without named leaders, like the End-User WG?

gc-check-ins.md
```diff
@@ -0,0 +1,20 @@
...
+* The GC liaison is responsible to bring the concerns of the SIG leads forward to the GC, and they are responsible for ensuring that all action items agreed upon during that check in are logged as issues on the project repository or community repository.
+* The GC liaison will attend the regular project meetings at least once per quarter. If this is difficult to accomplish due to timezone constraints, the GC liaison can be freed from this obligation by the GC. This exception should only rarely be approved.
+* The GC liaison and the maintainers are responsible to provide a quarterly health update, where they assess project health along three dimensions:
+  * Whether the SIG is "On Schedule" relative to its own published roadmap
```
jpkrohling, line 18: Thinking about the SIG Collector here, I have to ask: which roadmap? There's a specific project in the scope of Collector named "Collector v1", but I don't think that's what is meant here, right?

gc-check-ins.md
```diff
@@ -0,0 +1,20 @@
...
+  * The "contributor experience": PR turnaround time, newbie-friendly projects, advancement opportunities, automation, etc
```
jpkrohling, line 19: Ideas: # of new contributors, # of new triagers/approvers/maintainers, as well as the churn for those metrics.

jpkrohling (commented):
This uses different terms for the same set of people:

- maintainers
- SIG leads

Similarly, SIG and project seem to be used interchangeably. I think it's fine, but it might be worth a note somewhere, especially because we do have a definition of a project (ie, things under the projects directory in the community repo).

svrnm:
> This uses different terms for the same set of people:
> 
>     * maintainers
> 
>     * SIG leads
> 
> 
> Similarly, SIG and project seem to be used interchangeably. I think it's fine, but it might be worth a note somewhere, especially because we do have a definition of a project (ie, things under the projects directory in the community repo).

I tried to minimize that language based on the conversations we had on that recently. Let me review the material in that context once again

gc-check-ins.md
svrnm: My suggestion would be that the SIG names those leaders then. What is not defined here I guess is how a SIG manages that relationship: are all maintainers in contact with the GC liaison or do they appoint someone as well, or are they doing something "rolling". I think this (a) does not need to be defined yet, and (b) might be different from SIG to SIG

gc-check-ins.md
svrnm: Oh, good point, I think we talked about that during the GC meeting as well. @dyladan brought up the point of the "SIG Charter" in that context. We could make this wording a little weaker for now like ("the SIG is on schedula relative to its own goals", or "the SIG is making reasonable progress" until we figured out something better.

danielgblanco (approved):

gc-check-ins.md
svrnm: I added some wording on this topic

gc-check-ins.md
svrnm: I will resolve this conversation, since it has no impact on this document right now.

gc-check-ins.md
svrnm: fixed

gc-check-ins.md
svrnm: added

jpkrohling (approved):
:shipit:

PR #1983, opened 2024-03-01
Add actuated doc

dmathieu:
This adds a new documentation on using the actuated GitHub app.

### DIFF
--- a/docs/using-actuated.md
+++ b/docs/using-actuated.md
@@ -1,0 +1,39 @@
+# Using actuated
+
+[Actuated](https://actuated.dev/) is available as a GitHub App which can be
+enbabled on repositories, and allows GH Actions to execute on runners hosted by
+the CNCF. [See the announcement](https://actuated.dev/blog/arm-ci-cncf-ampere)
+
+## Installation
+
+Before it can be used, the App has to be enabled on the repository, through a
+maintenance request.  [See an
+example](https://github.com/open-telemetry/community/issues/1954).
+
+## Usage
+
+Using actuated on a GitHub Actions job consists in changing the `runs-on`
+section to be `actuated-*` (to run jobs on x86), or `actuated-arm64-*`, with
+the required CPUs and RAM in the name.
+
+For example:
+
+```yaml
+runs-on: actuated-4cpu-16gb
+```
+
+This configuration will ensure your job has 4 CPUs and 16GB of RAM available.
+
+While not specifying any CPU/RAM values (`runs-on: actuated`) works, using that
+format is not recommended by the actuated folks.
+
+## VMMeter
+
+As you set up your job, you may need an estimation of the resources it needs,
+to better set your requirements.
+
+Actuated provides VMMeter, a small action which will output system information
+at the end of a run.
+Follow [the
+instructions](https://gist.github.com/alexellis/1f33e581c75e11e161fe613c46180771)
+to set it up in your job.
--- a/docs/using-actuated.md
+++ b/docs/using-actuated.md
@@ -31,8 +31,8 @@ runs-on: actuated-arm64-4cpu-16gb
 
 This configuration will ensure your job has 4 CPUs and 16GB of RAM available.
 
-While not specifying any CPU/RAM values (`runs-on: actuated-arm64`) works,
-using that format is not recommended by the actuated folks.
+While not specifying any CPU/RAM values (`runs-on: actuated-arm64`) currently
+works, using that format is deprecated and will be removed.
 
 ## Right-sizing VMs
 
--- a/docs/using-actuated.md
+++ b/docs/using-actuated.md
@@ -34,7 +34,7 @@ This configuration will ensure your job has 4 CPUs and 16GB of RAM available.
 While not specifying any CPU/RAM values (`runs-on: actuated-arm64`) works,
 using that format is not recommended by the actuated folks.
 
-## VMMeter
+## Right-sizing VMs
 
 As you set up your job, you may need an estimation of the resources it needs,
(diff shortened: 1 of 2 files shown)


docs/using-actuated.md
```diff
@@ -0,0 +1,39 @@
...
+
+Before it can be used, the App has to be enabled on the repository, through a
+maintenance request.  [See an
+example](https://github.com/open-telemetry/community/issues/1954).
+
+## Usage
+
+Using actuated on a GitHub Actions job consists in changing the `runs-on`
+section to be `actuated-*` (to run jobs on x86), or `actuated-arm64-*`, with
+the required CPUs and RAM in the name.
```
svrnm, line 17: Is there a place where I can look up the available list of `actuated-*` ?

docs/using-actuated.md
```diff
@@ -0,0 +1,39 @@
+# Using actuated
+
+[Actuated](https://actuated.dev/) is available as a GitHub App which can be
+enabled on repositories, and allows GH Actions to execute on runners hosted by
+the CNCF. [See the announcement](https://actuated.dev/blog/arm-ci-cncf-ampere)
+
```
svrnm, line 6: Can you add a few words when to use actuated and when not to use it and to stick with GH provided runners? I assume that it is not the idea to do ALL runs on actuated?

svrnm (changes requested):
thank you for writing this down, some comments!

docs/using-actuated.md
dmathieu: They mention on their docs that `actuated-*` and `actuated-arm64-*` are available, but I'm not seeing any matrix that would give the list of all available values.
https://docs.actuated.dev/

docs/using-actuated.md
dmathieu: I've done that in d05b91c0c1abfc945e9a18211a6e306c0f9ee183.

docs/using-actuated.md
svrnm: Ok, then we might need to follow up with actuated team here? Let's put this in a follow up issue, such that it is non-blocking

docs/using-actuated.md
svrnm: thx!

svrnm (approved):

docs/using-actuated.md
dmathieu: I have pinged them on their slack.

dmathieu: So, their answer is actually that we could have anything we want. But CNCF orgs only have arm64 enabled (it's all on bare metal on equinix as far as I understand). 
I'll remove the mention of non-arm64.
https://actuated.dev/blog/arm-ci-cncf-ampere

svrnm: thanks for the follow up! what can we put there instead of the `*`? In the blog post I see

runs-on: actuated-arm64

and

runs-on: actuated-arm64-8cpu-16gb

and

actuated-arm64-8cpu-32gb.

So it's something like CPU number + RAM size?

dmathieu: Yes. As far as I can see, that's all that can be specified. I suppose they use cgroup (maybe through k8) limits, so setting a value higher than what is available may never start the runner.

@atoulme may know more about what machine we have available?

alexellis:
Hi, we already have docs that cover VM sizes, I don't think there's any need to repeat content here and have to maintain it in multiple places.

Please see:

https://docs.actuated.dev/examples/custom-vm-size/

And the CNCF announcement blog post.

If you should need anything more, please hit us up on the actuated Slack as we cannot keep an eye on all GitHub mentions.

docs/using-actuated.md
svrnm: as per @alexellis comment below https://docs.actuated.dev/examples/custom-vm-size/ is what we should be referencing then

docs/using-actuated.md
```diff
@@ -0,0 +1,46 @@
...
+For example:
+
+```yaml
+runs-on: actuated-arm64-4cpu-16gb
+```
+
+This configuration will ensure your job has 4 CPUs and 16GB of RAM available.
+
+While not specifying any CPU/RAM values (`runs-on: actuated-arm64`) works,
```
alexellis, line 34: Please say that `actuated-arm64` is deprecated. We will remove this.

docs/using-actuated.md
```diff
@@ -0,0 +1,46 @@
...
+using that format is not recommended by the actuated folks.
+
+## VMMeter
+
```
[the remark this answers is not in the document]
dmathieu: Done. I've named it `Righht sizing VMs` though, the `GitHub Actions` part would be redundant.

docs/using-actuated.md
```diff
@@ -0,0 +1,49 @@
...
+works, using that format is deprecated and will be removed.
+
+## Right-sizing VMs
+
+As you set up your job, you may need an estimation of the resources it needs,
+to better set your requirements.
+
+Actuated provides VMMeter, a small action which will output system information
+at the end of a run.
+Follow [the
```
alexellis, line 44: Should

```
[the
instructions]
```

Be split over two lines?

dmathieu: That has no impact over the rendering of markdown, and that's how VIM split it to keep the line within 80 characters.

trask (approved):
can you add a link to this from the [assets.md](https://github.com/open-telemetry/community/blob/main/assets.md) page?

pellared (approved):

svrnm:
Thanks!

PR #2025, opened 2024-03-28
Modify GC scopes to include GitHub permissions.

austinlparker:
Currently, the TC are the only org users with org admin access. This PR clarifies that the GC also is responsible for GitHub org level actions.


reyang (approved):
Makes sense to me.

danielgblanco (approved):

jpkrohling (approved):
I believe we need 2/3 approvals for this change. I would also be nice to give the TC a heads-up.

reyang:
> I believe we need 2/3 approvals for this change. I would also be nice to give the TC a heads-up.

I'll bring this up during the TC meeting this week.

tedsuo (approved):

dyladan (approved):

governance-charter.md
```diff
@@ -32,7 +32,8 @@ following:
   how contributors become committers/maintainers, approvers, reviewers, members,
   etc.
 - Decide, for the purpose of elections, who can vote.
-- Logo/landing page/marketing. 
+- Logo/landing page/marketing.
+- GitHub repository management, membership, and hosting.
```
jack-berg, line 36: It would be very appreciated to have help with many of the github repository management things, but hoping we can somehow retain a necessary level of critical review for things that could potentially have security implications. I'm thinking about responding to requests to add new plugins, org level secrets, etc.

Would be make sure that anything potentially sensitive has been considered by at least a few trusted people.

dyladan: I assume what you're saying is that the TC should have input on those things. The TC still has "GitHub repository management, membership and hosting" in its charter and this PR isn't removing it. I think this PR is just making it clear that it is a shared responsibility between TC and GC

jack-berg: Its good with me as long as everyone has a base level intuition about what types of things are potentially sensitive 👍

jmacd: Considering all the responsibilities of the GC listed above and below, I think this is an appropriate change. If the GC has responsibility as stated around lines 45/46, 47/48, and 52, then I think we have to assume they are acting responsibly. The technical committee is responsible for security concerns, which often crosses with Github repository management -- however it does so as delegated by the policy established by the GC, so I believe all we can really say is "be reponsible", and I think that is assumed. (The TC met to discuss and did not have quorum, today.)

svrnm (approved):

governance-charter.md
austinlparker: Wanted to close the loop here -- at the GC/TC meeting yesterday this PR was discussed, and we walked away with an understanding around the mutual goals:
- Our desired future state is to ensure least privilege principles for org admins/users, and we plan to accomplish this through automation/IaC
- The next step after this will be to identify a smaller group of combined GC/TC members who can work towards this goal. Once there's alignment on the composition of this group and how they will work, org admin access will be moved to that group.
- This group will then work to implement least privilege and automation per plan.

Does that sound good to folks here?

tigrannajaryan (approved):

PR #2042, opened 2024-04-05
Project Proposal: Feature Flag Semantic Conventions

dyladan:
This proposes a new project to stabilize feature flag semantic conventions. Created as a draft for now while we fill out the rest of the project template.

TODOs:
- [x] list engineers committed to working on the project
- [x] list prototype engineers and maintainers committed to review


projects/feature-flag.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Deliverables
+
+The project deliverable will be a stable feature flag semantic convention for flag evaluations, also called impressions, and feature flag changes.
+Both client and server uses should be considered, and prototypes should be created for at least one server and at least one client use case.
+
+While prototypes will be delivered as a part of the development effort of this project, stable instrumentation is out of scope of this project and is expected to come from the OpenFeature project in the form of OpenFeature hooks or native instrumentation in OpenFeature SDKs.
+Any use of baggage, including propagating fields to be used in feature flag evaluation context, is also out of scope.
+
+## Staffing
```
kinyoklion, line 56: I can at least help to review prototypes.

Ryan Lamb (OpenFeature TC, LaunchDarkly)

dyladan: Added you to the list

projects/feature-flag.md
```diff
@@ -0,0 +1,111 @@
+# Feature Flag Semantic Conventions
+
+OpenFeature is a CNCF Incubating project aiming to standardize feature flags.
```
arminru, line 3:
```suggestion
[OpenFeature](https://openfeature.dev/) is a CNCF Incubating project aiming to standardize feature flags.
```

projects/feature-flag.md
```diff
@@ -0,0 +1,111 @@
...
+They currently provide 11 open source SDKs in various languages and support more than 10 feature flagging backends.
+OpenFeature has had contributions from over 50 companies and is actively developed by a combination of end users, feature flagging vendors, and platform providers.
```
arminru, line 5: Could link to the lists at https://openfeature.dev/community/interested-parties.

projects/feature-flag.md
```diff
@@ -0,0 +1,111 @@
...
+
+OpenFeature provides observability data via OpenTelemetry hooks.
+A hook is an OpenFeature plugin which may execute logic at various stages of the feature flag evaluation process.
+Currently, those hooks use the experimental feature flagging semantic convention.
```
arminru, line 9:
```suggestion
Currently, those hooks use the experimental [feature flagging semantic convention](https://opentelemetry.io/docs/specs/semconv/feature-flags/).
```

projects/feature-flag.md
```diff
@@ -0,0 +1,111 @@
...
+
+This project also includes Dan Dyla acting as one of the project leads.
+Dan is a member of the GC, maintainer of OTel JS, and has experience developing semantic conventions.
+Dan will provide the necessary expertise to ensure the proposed semantic conventions meet the standards and expectations of the OpenTelemetry community and semantic conventions SIG.
+
+The following is a list of engineers committed to working on the project:
+
+- Daniel Dyla (OTel GC, OTel JS, Dynatrace)
+- Ryan Lamb (OpenFeature TC, LaunchDarkly)
+- TODO: more engineers
```
arminru, line 69: Should this be moved/merged into `### Required staffing`?

projects/feature-flag.md
```diff
@@ -0,0 +1,141 @@
...
+Once the project is approved a project board will be created and linked here.
+The project lead and all other relevant project members should have edit access to the board.
+The board will be pre-populated with at least the following issues to track deliverables and the timeline.
+
+- **signal type for feature flag impressions** - agree on which signal type or types should be used to represent feature flag impressions.
+- **signal type for feature flag changes** - agree on which signal type or types should be used to represent feature flag changes.
+- **semantic conventions for feature flag impressions** - make any necessary additions to the semantic conventions to support feature flag impressions
+- **semantic conventions for feature flag changes** - make any necessary additions to the semantic conventions to support feature flag change events
+- **prototype feature flag impressions**
+- **prototype feature flag change events** - feature flag change events may not be generated by SDKs or instrumentations, but by the feature flag management tools.
```
julealgon, line 141: Just a minor comment on this particular point here.

If the management tooling raises the event, this could create a temporal inconsistency where some or all of the applications have not yet fetched the updated values.

In the .NET ecosystem (I'm speaking only on the platform I have experience with here, but this might apply to others), it is incredibly common for one to set a new value to a key (for example, on Azure AppConfiguration), but then have many applications waiting on different refresh timings or cache expirations for that key, meaning they would potentially still serve N requests using the previous value. Even when using something like a per-request middleware, the refresh operation tends to happen in the background and will usually not affect the _current_ request or other parallel requests happening around the same time.

If the only source for a flag change events is the central tool, one could end up checking behavior still running on the old value while the metric/etc indicates a new value should be present.

I think this needs to be reconsidered from the perspective of each application/process as well. Maybe we need a change event from the configuration source, and a second change event on each application when the new value is fetched? That to me would be the most comprehensive approach, although I'm not sure exactly how those event streams would be represented.

I almost want to suggest that OpenTelemetry should define a 4th type of telemetry called "events" (or "state transitions" or something), generalizing the concept and then making span events a particular use case of that. Then, when looking at a particular trace, one could include "global events" like these to understand the complete behavior on a given flow.

Anyways... just throwing some ideas around here.

beeme1mr: We planned to define semantics for change events from the management tool (source of truth) and the services (SDKs). That way, you could track how long it took for a change to propagate and quickly identify which services haven't been updated.

OpenTelemetry already has an [events concept](https://opentelemetry.io/docs/specs/otel/logs/event-api/) that's basically semantics on top of logs.

julealgon: > We planned to define semantics for change events from the management tool (source of truth) and the services (SDKs). That way, you could track how long it took for a change to propagate and quickly identify which services haven't been updated.

Perfect. Love to hear it.

> OpenTelemetry already has an [events concept](https://opentelemetry.io/docs/specs/otel/logs/event-api/) that's basically semantics on top of logs.

I remember seeing this but always assumed it was just a "synonym" to logs. Looks like I was wrong as it even has its own classes and abstractions.

Am I correct in that observability tools don't quite have this separation between logs and events yet though? Should they? We use Datadog extensively and I never saw this distinction anywhere over there (although I'm not necessarily an expert in DD either... so it could be me just missing it). 

I also don't know what exactly maps to these events from the .NET APIs, since we only have `ILogger` (for logs), `Meter` and `Instrument`s (for metrics) and `Activity` (for spans) there. The only place where I ever saw the concept of "events" was for "activity events", and those only make sense if you _are_ under an activity in the first place.

projects/feature-flag.md
```diff
@@ -0,0 +1,141 @@
...
+
+**Engineers willing to write and review prototypes:**
+
+TODO: list prototype languages and engineers
+
+The following is a list of engineers committed to working on the project:
+
+- JS Prototype - Daniel Dyla (OTel GC, OTel JS, Dynatrace)
+- Review Prototypes - Ryan Lamb (OpenFeature TC, LaunchDarkly)
+- TODO: more engineers
```
federicobond, line 110: I can help with prototypes / review / testing.

- Federico Bond (OpenFeature Python, Independent)

AlexanderWert, line 110:
```suggestion
- Federico Bond (OpenFeature Python, Independent)
- TODO: more engineers
```

projects/feature-flag.md
dyladan: >I remember seeing this but always assumed it was just a "synonym" to logs. Looks like I was wrong as it even has its own classes and abstractions.

Events and logs are the same data type. Events are LogRecords which have a name which uniquely defines a particular class or type of event. All events with the same name have Payloads that conform to the same schema, which assists in analysis in observability platforms. Events are described in more detail in the [semantic conventions](https://opentelemetry.io/docs/specs/semconv/general/events/).

>Am I correct in that observability tools don't quite have this separation between logs and events yet though? Should they?

For the most part they don't. There could be arguments made either way as to if they should.

beeme1mr:
@askpt has volunteered to support us from a .NET perspective.

https://cloud-native.slack.com/archives/C06RT64NP37/p1715113701675879?thread_ts=1714057336.076709&cid=C06RT64NP37

projects/feature-flag.md
rossgrambo: Hello! I'm part of the `Microsoft.FeatureManagement` team.

> If the only source for a flag change events is the central tool, one could end up checking behavior still running on the old value while the metric/etc indicates a new value should be present.

We solve this with versioning emitted on the `feature_flag` event. For us- the events usually come from the SDK, but the centralized server dictates what version we're working with. This allows a given user/account/activity to be connected to the appropriate version of the flag, meaning you know which version applies to the rest of the metrics.

projects/feature-flag.md
```diff
@@ -0,0 +1,142 @@
...
+**Engineers willing to write and review prototypes:**
+
+TODO: list prototype languages and engineers
+
+The following is a list of engineers committed to working on the project:
+
+- JS Prototype - Daniel Dyla (OTel GC, OTel JS, Dynatrace)
+- Review Prototypes - Ryan Lamb (OpenFeature TC, LaunchDarkly)
+- Federico Bond (OpenFeature Python, Independent)
+- TODO: more engineers
```
askpt, line 111:
```suggestion
- André Silva (OpenFeature dotnet)
- TODO: more engineers
```

From @beeme1mr comment: https://github.com/open-telemetry/community/pull/2042#issuecomment-2099260443

lukas-reining:
As said @dyladan, I would love to be in for that. (OpenFeature TC)

dyladan:
> As said @dyladan, I would love to be in for that. (OpenFeature TC)

Awesome thanks. Added you

projects/feature-flag.md
rossgrambo: Sadly, I'm unable to join at this time. Very interested to see what y'all land on.

beeme1mr (approved):

AlexanderWert (approved):

dyladan:
Project leads/sponsors have approved. I can't approve because I opened the PR but to be clear, I do approve.

trask (approved):

projects/feature-flag.md
```diff
@@ -95,21 +95,23 @@ The project is sponsored by Dan Dyla and Alexander Wert, an OpenTelemetry semant
...
-
-- JS Prototype - Daniel Dyla (OTel GC, OTel JS, Dynatrace)
-- Review Prototypes - Ryan Lamb (OpenFeature TC, LaunchDarkly)
-- Federico Bond (OpenFeature Python, Independent)
+- Alexander Wert - (OTel SemConv, Elastic)
 - André Silva (OpenFeature dotnet)
-- TODO: more engineers
+- Daniel Dyla - (OTel GC, OTel JS, Dynatrace)
+- Federico Bond (OpenFeature Python, Independent)
+- Lucas Reining (OpenFeature TC)
```
lukas-reining, line 112: My name is Lukas with `k` @dyladan :) and maybe you can add `codecentric` as company.

projects/feature-flag.md
```diff
@@ -0,0 +1,145 @@
...
+This project aims to develop and stabilize feature flag semantic conventions, using the existing experimental semantic conventions as a starting point.
+
+## Problem 1: Determining Flag Impact
+
+Currently, it is impossible to determine the impact a feature flag has on key metrics for your services.
+For example, if a feature flag is enabled for 25% of users, additional data is required to know if that subset of users is experiencing faster, slower, or unchanged response times.
+This problem can become exponentially more challenging if multiple feature flags are evaluated over the course of a single transaction.
+
+Further, it may also be difficult or impossible to know exactly how many users are being affected by a particular feature flag.
+For example, if a feature flag targets users with a particular email domain, it may require additional monitoring data to determine what portion of your users or traffic is receiving a particular experience.
```
svrnm, line 22:
```suggestion
For example, if a feature flag targets users with a particular email domain, it may require additional observability data to determine what portion of your users or traffic is receiving a particular experience.
```

projects/feature-flag.md
```diff
@@ -0,0 +1,145 @@
...
+
+## Problem 2: Flag management separate from monitoring data
+
+Feature flags are typically managed outside your monitoring tooling.
+There is no standardized way for a flag management system to notify a monitoring system that a feature flag rule set is changed, or if a rule is enabled or disabled.
+This lack of standardization means that even if a flag management system emits a flag change event, the monitoring tool may not be able to correlate that change with the services that it impacts.
+It also means that a person responsible for monitoring an application often has to check two separate tools in order to determine if a flag change was the root cause of a change in service metrics or behavior.
+Because rule sets and monitoring data are managed in separate tools, it may be difficult to determine exactly which rule set was used and if it has been changed between when it was evaluated and when a person investigating an issue is looking at the monitoring data.
```
svrnm, line 30:
```suggestion
## Problem 2: Flag management separate from observability data

Feature flags are typically managed outside your observability tooling.
There is no standardized way for a flag management system to notify a observability system that a feature flag rule set is changed, or if a rule is enabled or disabled.
This lack of standardization means that even if a flag management system emits a flag change event, the observability tool may not be able to correlate that change with the services that it impacts.
It also means that a person responsible for the observability of an application often has to check two separate tools in order to determine if a flag change was the root cause of a change in service metrics or behavior.
Because rule sets and observability data are managed in separate tools, it may be difficult to determine exactly which rule set was used and if it has been changed between when it was evaluated and when a person investigating an issue is looking at the observability data.
```

projects/feature-flag.md
```diff
@@ -0,0 +1,145 @@
...
+
+## Target Use Cases
+
+- Analyze the impact a feature flag has on key service metrics such as response time and failure rate.
+  If a feature variant is enabled for some subset of users, it is necessary to be able to distinguish those transactions from other transactions in your monitoring backend.
```
svrnm, line 35:
```suggestion
  If a feature variant is enabled for some subset of users, it is necessary to be able to distinguish those traces from other traces in your observability backend.
```

projects/feature-flag.md
```diff
@@ -0,0 +1,145 @@
...
+
+OpenFeature is looking to expand their observability solution either by improving the currently available hooks or by providing native instrumentation within the OpenFeature SDK itself.
+In order to support that effort, OpenFeature requires a stable semantic convention to build instrumentation around.
+This project aims to develop and stabilize feature flag semantic conventions, using the existing experimental semantic conventions as a starting point.
+
+## Problem 1: Determining Flag Impact
+
+Currently, it is impossible to determine the impact a feature flag has on key metrics for your services.
+For example, if a feature flag is enabled for 25% of users, additional data is required to know if that subset of users is experiencing faster, slower, or unchanged response times.
+This problem can become exponentially more challenging if multiple feature flags are evaluated over the course of a single transaction.
```
svrnm, line 19:
```suggestion
This problem can become exponentially more challenging if multiple feature flags are evaluated over the course of a single trace.
```

svrnm (approved):
LGTM overall, some wording suggestions (transaction to trace, monitoring to observability)

julealgon:
@svrnm 
> ...monitoring to observability

Is this a known preferred naming convention? What about calling it "telemetry" instead, wouldn't that be even more consistent with "OpenTelemetry"?

lukas-reining (approved):

projects/feature-flag.md
```diff
@@ -0,0 +1,145 @@
...
+- Alexander Wert - Project Sponsor (OTel SemConv, Elastic)
+- Daniel Dyla - Project Lead and Sponsor (OTel GC, OTel JS, Dynatrace)
+- Michael Beemer - Project Lead (OpenFeature co-founder and GC, Dynatrace)
+
+**Engineers:**
+
+The following is a list of engineers committed to working on the project and write and review prototypes:
+
+- Alexander Wert - (OTel SemConv, Elastic)
+- André Silva (OpenFeature dotnet)
```
askpt, line 109:
```suggestion
- André Silva (OpenFeature dotnet, Lexis Nexis Risk Solutions)
```

dyladan:
not sure what's causing easycla to fail on some of the commits

svrnm:
> not sure what's causing easycla to fail on some of the commits

I _think_ the issue are the emails in Co-Authored-by, since this is a new feature in easyCLA to be verified as well, see

https://github.com/communitybridge/easycla/releases

Everyone might need to check if their @users.noreply.github.com address is also covered by the easyCLA

askpt:
> > not sure what's causing easycla to fail on some of the commits
> 
> I _think_ the issue are the emails in Co-Authored-by, since this is a new feature in easyCLA to be verified as well, see
> 
> https://github.com/communitybridge/easycla/releases
> 
> Everyone might need to check if their @users.noreply.github.com address is also covered by the easyCLA

I just signed in using my @users.noreply.github.com address and primary address. It seems it didn't update automatically.

svrnm:
see https://github.com/communitybridge/easycla/issues/4329

mtwo (approved):

danielgblanco (approved):

jpkrohling (approved):

PR #2061, opened 2024-04-18
Update TC sponsors to Sponsors

trask:
I meant to do this in #2058


trask:
Closing, going to combine this into another PR to avoid conflicts with myself

PR #2080, opened 2024-04-29
Add community activities to membership requirements

svrnm:
This updates the membership requirements to reflect that we have SIGs who express their contributions to the community outside of GitHub, and especially by organizing activities.

I think there are other parts of that document that require refinement but this is a starting point

### DIFF
--- a/.github/ISSUE_TEMPLATE/membership.md
+++ b/.github/ISSUE_TEMPLATE/membership.md
@@ -37,4 +37,5 @@ Please remember, it is an applicant's responsibility to get their sponsors' conf
 
 - PRs reviewed / authored
 - Issues responded to
+- Community activities I organized/ran
 - SIG projects I am involved with
--- a/community-membership.md
+++ b/community-membership.md
@@ -49,6 +49,7 @@ Defined by: Member of the OpenTelemetry GitHub organization
   may include, but is not limited to:
   - Authoring or reviewing PRs on GitHub
   - Filing or commenting on issues on GitHub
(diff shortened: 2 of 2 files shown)


danielgblanco (approved):
I think this is a great addition. Thanks @svrnm

jpkrohling (commented):
@jack-berg, based on the comments at a PR of yours about spec sponsors, do you think we need an extra point here about SIGs that are being formed?

arminru (approved):

trask (approved):

jack-berg:
> @jack-berg, based on the comments at a PR of yours about spec sponsors, do you think we need an extra point here about SIGs that are being formed?

I don't think so actually. There's an entry in the set of possible contributions which describes SIG work as contributions:

> Contributing to SIGs, subprojects, or community discussions (e.g. meetings, chat, email, and discussion forums)

I think this means that SIG work counts towards contributions from the perspective of being a spec sponsor.

jack-berg (approved):

svrnm:
> > Contributing to SIGs, subprojects, or community discussions (e.g. meetings, chat, email, and discussion forums)
> 
> I think this means that SIG work counts towards contributions from the perspective of being a spec sponsor.

Just as a note on that: while we have this written down, we still make use of CNCF DevStats to figure out if someone is a "member of standing" for example, that's why I am a big fan of the end user, mainframe and security SIG to have their own repositories to put down some of their work, as it is tracked through DevStats.

Issue #2088, opened 2024-05-03
Create missing documents for projects


svrnm:
We have a few already running projects[^1] that do not have a document in the [projects](https://github.com/open-telemetry/community/tree/main/projects) folder, main reason is because they pre-date the process. Can the GC liaisons work with the SIGs to (a) verify if a document is required (i.e. the SIG is "not permanent") and if so create those documents?

- [x] Client Instrumentation SIG (can be copied from [this issue](https://github.com/open-telemetry/opentelemetry-specification/issues/2734) probably), @danielgblanco
- [x] Profiling SIG (can be copied from [this issue](https://github.com/open-telemetry/opentelemetry-specification/issues/2734) probably), @mtwo, @open-telemetry/profiling-maintainers 
- [x] System SemConv (can be copied from [this issue](https://github.com/open-telemetry/opentelemetry-specification/issues/3556) probably), @svrnm, @open-telemetry/semconv-system-approvers 
- [x] Prometheus Interoperability SIG, @alolita, @open-telemetry/wg-prometheus 
- [x] Agent Management SIG, @tedsuo, @open-telemetry/opamp-spec-approvers (probably not required to have one?)

Please check the box if either your SIG does not match the need to require a document or when you have added the document to the folder. 

[^1]: Picked them from https://github.com/orgs/open-telemetry/projects/29/views/1 during project triage session

tigrannajaryan:
>Agent Management SIG

The SIG's work is actively in progress and there is no clear target when we can consider it "done". We can still add a project document if you prefer that. I don't have a strong opinion.

dashpole:
I'll write the document for the prometheus compatibility WG.  I hope to wrap up the WG in the next few months.

svrnm:
Thank you all!

> The SIG's work is actively in progress and there is no clear target when we can consider it "done". We can still add a project document if you prefer that. I don't have a strong opinion.

If the SIG is permanent (or very long running/with no clear end goal yet), I think we do not need a document.

svrnm:
Since creating the document for the profiling SIG is probably as simple as copying [this issue](https://github.com/open-telemetry/opentelemetry-specification/issues/2734) into a markdown file, I argue it's still nice to have (same applies to the client instrumentation SIG, which has such an issue as well).

danielgblanco:
[...]

PR #2089, opened 2024-05-03
Create system-semconv.md

svrnm:
Contributes to https://github.com/open-telemetry/community/issues/2088


danielgblanco (approved):

austinlparker (approved):

svrnm:
@open-telemetry/semconv-system-approvers please take a look as well

mx-psi (approved):
LGTM, the doc is more a historical plan than what is actually happening but it does reflect our original plans

projects/system-semconv.md
```diff
@@ -0,0 +1,86 @@
...
+  - [@kaiyan-sheng](https://github.com/kaiyan-sheng)
+  - [@braydonk](https://github.com/braydonk)
+  - [@bertysentry](https://github.com/bertysentry)
+  - [@dineshg13](https://github.com/dineshg13)
+
+## Meeting Times
+
+July 10 - August 21 (6 weeks):
+
+- Wednesday at 8:30-09:00 Pacific Time
```
ChrsMark, line 67: I know this was the original time. However could this be misleading since we have changed the time since then? Maybe an "update" section bellow could help here?

ChrsMark (approved):

projects/system-semconv.md
svrnm: I think that's a valid concern, but this applies to other documents as well, maybe something we fix for all of them at once.

trask: I think it would also be ok to remove the meeting times from this document

trask (approved):

projects/system-semconv.md
svrnm: good idea, I removed the meeting times.

PR #2113, opened 2024-05-15
Add semantic conventions tooling meeting.

jsuereth:
I wasn't sure whether to keep same sponsors as existing Semconv WG, so I removed anyone I hadn't asked explicitly (so just myself for now).

For now, we need to at least publicize the meeting.


svrnm:
Can you apply the update to sigs.yml and then run `make table-generation` to have the README updated from there automatically, otherwise the next PR will overwrite your change.

jsuereth:
`make table-generation` doesn't work for me.  Could you move to a docker-based tooling solution so we don't have to fight python environments?

svrnm:
There is a docker based fallback:

```
		docker run --rm -v ${PWD}:/repo -w /repo python:3-alpine python ./scripts/update-sig-tables.py --install; \
```

but I can make this the default/the only solution

arminru (approved):

jsuereth:
@svrnm Yeah, the docker command worked.  I see your if/else.  The issue is I *have* python, but not one compatible with your script, so it failed completely.  I'm not sure if you can improve the python env detection or just give up and default to docker the way we do in the Spec / semconv builds...

svrnm:
> @svrnm Yeah, the docker command worked. I see your if/else. The issue is I _have_ python, but not one compatible with your script, so it failed completely. I'm not sure if you can improve the python env detection or just give up and default to docker the way we do in the Spec / semconv builds...

I think I just go with docker like the spec / semconv builds

svrnm (approved):

trask (approved):

PR #2119, opened 2024-05-16
cspell: ignore github handles

svrnm:
This introduces an ignore pattern for cspell such that github handles are no longer spell checked when they are written with an `@` as prefix, e.g. `@user`


arminru (approved):
Nice, thank you!

danielgblanco (approved):

trask (approved):
💯

PR #2144, opened 2024-06-11
Propose Developer Experience project

tsloughter:


projects/convenience.md
```diff
@@ -0,0 +1,74 @@
+# Convenience Project
+
+## Background and description
+
+Convenience as a SIG was initially proposed by Ted Young back in 2021. It
```
svrnm, line 5:
```suggestion
Convenience as a SIG was initially proposed by [Ted Young](https://github.com/tedsuo) back in 2021. It
```

projects/convenience.md
```diff
@@ -0,0 +1,74 @@
...
+was postponed to prioritize completion of signal specs and stabilization. A
+document was written at this time describing user research to do and a
+brainstorming session,
+https://docs.google.com/document/d/1bjiw5L4E8FztRtfsy4Nu9A6qSP-BJlK6BBzl6Rea-Ds/edit#heading=h.b7qckyjgu28i
```
svrnm, line 9:
```suggestion
https://docs.google.com/document/d/1bjiw5L4E8FztRtfsy4Nu9A6qSP-BJlK6BBzl6Rea-Ds/edit#heading=h.b7qckyjgu28i
```

can you provide a tl;dr for this document?

projects/convenience.md
```diff
@@ -0,0 +1,74 @@
...
+
+### Current challenges
+
+Currently the various language implementations resolve user convenience
+individually and this can lead to drift between different implementations and
+with the spec. An example would be how to handle marking a code block as
+`untraced` which has implementations in at least Java, Ruby and Javascript, each
+done differently, named differently and with no specification.
```
svrnm, line 17: I am not sure if I understand this problem in the context of convenience, can you elaborate on why this is bad in that context?

projects/convenience.md
```diff
@@ -0,0 +1,74 @@
...
+
+### Goals, objectives, and requirements
+
+With the 3 main signals going stable the time is right to circle back and
+address issues users may have faced over the years they've now been using
+OpenTelemetry.
+
+This project aims to identify developer experience issues with using
+OpenTelemetry, to select 3 that are concerns which can be addressed through
+updates to the specification and resolve them.
```
svrnm, line 27: I assume this is the initial scope and that this SIG will continue afterwards?

svrnm (commented):
Thanks for proposing this project @tsloughter! Some feedback to the proposal

projects/convenience.md
tsloughter: Yup, will do!

projects/convenience.md
svrnm: That makes sense, so the SIG identifies those 3 items, works on them and after they are accomplished the members of that SIG can decide if they want to continue. I like that 👍

projects/convenience.md
tsloughter: Finally updated.

projects/convenience.md
tsloughter: Updated in the text.

projects/convenience.md
```diff
@@ -0,0 +1,88 @@
...
+identified problem, OTEP's will be created where appropriate before opening PR's
+to the specification.
+
+## Staffing / Help Wanted
+
+- Project Lead: Tristan Sloughter (MyDecisiveAI)
+
+- TC sponsors: 
+  - Austin Parker (Honeycomb)
+  - TBA
```
svrnm, line 60: @open-telemetry/governance-committee @open-telemetry/technical-committee can we get sponsorship for this project proposal?

yurishkuro:
I recommend the name Usability, not Convenience.

tsloughter:
I'm good with a name change if people prefer "usability". I was also considering "developer experience".

austinlparker (approved):
Overall, I support this!

projects/convenience.md
tedsuo, line 60: I will sponsor this!

tedsuo, line 60:
```suggestion
  - Ted Young (Lightstep)
```

I will sponsor this!

tsloughter:
@yurishkuro what do you think of "developer experience"?

yurishkuro:
yeah, sounds good

theletterf (commented):
This is necessary. I wonder, though, about the following:

- Why 3 problems only? Why not generating a ranked list of issues?
- Surveys can be tricky. What would the sample be? The PoV of community members might be biased by years of experience. How to reach beginners and new users?
- Have we considered running UX research interviews?

tsloughter:
@theletterf the plan is a ranked list where we pick the top 3. The 3 is just to give the project a goal to reach rather than an open ended, "fix this list". After the 3 the group can decide whether to propose continuing as a full SIG or disbanding and waiting to regroup when there are resources available again to take on more issues.

How to reach the right users will be worked on with th End User SIG. 

UX research interviews? Would that be like a user trying to instrument an application for the first time live with an interviewer watching?

julianocosta89:
+1 for "Developer Experience"and I'd like to be involved in this SIG.
Thanks @tsloughter for bringing it back to life!

projects/convenience.md
```diff
@@ -0,0 +1,91 @@
...
+OpenTelemetry.
+
+This project aims to identify developer experience issues with using
+OpenTelemetry, to select 3 that are concerns which can be addressed through
+updates to the specification and resolve them.
+
+## Deliverables
+
+The first deliverable will be an end user survey. Assuming the agreement and
+participation of the End User SIG this will be done in cooperation with them. A
```
lmolkova, line 46: I like the idea of doing end-user research, but it could be easier to start by investigating prior art and asking OTel contributors and authors of OTel instrumentation libraries about convenience they have to introduce.

To name a few:
- Java instrumentation API https://github.com/open-telemetry/oteps/pull/165
- (somewhat similar) [Micrometer Observation API](https://spring.academy/guides/microservices-observability-reactive-spring-boot-3) that intends to simplify multi-signal instrumentation
- Enriching auto-collected telemetry in user apps: https://github.com/open-telemetry/oteps/pull/207
- Reducing error noise (some error like 404/401 are not really errors) - https://github.com/open-telemetry/semantic-conventions/issues/1003

projects/convenience.md
```diff
@@ -0,0 +1,91 @@
+# Convenience Project
```
JamieDanielson, line 1:
```suggestion
# Developer Experience Project
```

projects/convenience.md
```diff
@@ -0,0 +1,91 @@
...
+
+- GC sponsors: 
+  - Austin Parker (Honeycomb)
+  
+- TC sponsors:
+  - Ted Young (Lightstep)
+  - Liudmila Molkova (Microsoft)
+
+- Needed: Engineers and maintainers/approvers for multiple languages to develop
+  solutions and do prototypes.
```
dmathieu, line 66: Based on the folks who showed interest in contributing:

```suggestion
- Maintainers, approvers and contributors:
  - Damien Mathieu (Elastic)
  - Juliano Costa (Datadog)
```

projects/convenience.md
JamieDanielson: I guess we'll want to rename this file as well to devex or developer-experience? I left the others in the Background section because it's a discussion of history which explicitly referred to this as Convenience.

svrnm: Can we make sure that the name of this proposal can clearly be distinguished from the "Contributor Experience" proposal, since contributors to the otel project (often) also are developer or see themselves as developer ?

martinjt:
More than happy to be involved.

One note from skimming through. I believe we should stay away from making developer experience functions, methods etc. Consistent across all languages.

The key issues I see right now are when we try to make something work across all languages the same when there are massive differences between the idiomatic style and flow of different languages and ecosystems.

I think what we need to do is work out how we provide spec flexibility to allow them, while ensuring standards.

projects/convenience.md
jpkrohling: Came here to say the same: perhaps "instrumentation experience"?

martinjt: Developer Experience is a generally accepted term, and making it instrumentation or any other variation will diminish the impact of the initiative publicly.

Contributor Experience is a different thing and speaks directly to the target market of that.

You could say in the subtitle "working with OpenTelemetry", but changing away from Developer Experience wouldn't have the same impact we so desperately need right now.

svrnm:
> One note from skimming through. I believe we should stay away from making developer experience functions, methods 
> etc. Consistent across all languages.
> The key issues I see right now are when we try to make something work across all languages the same when there are 
> massive differences between the idiomatic style and flow of different languages and ecosystems.

Those 2 things are not mutually exclusive by default. You might have expected me saying that already, but I say it anyways;-) -- Consistency is a good thing! So it is something worth to strive for as long as it does not compete with other goals of this SIG, which is providing a good developer experience (which I agree the "flow of a language and ecosystem" can be). The keywords for consistency I find important in the proposal are "named differently" and "experiencing different results". Especially the first one is one that [hurts us badly already](https://github.com/open-telemetry/community/issues/2165). This can be improved in the proposal, but eventually it's up to the SIG to draw a line between consistency and other criteria for accomplishing a good experience

projects/convenience.md
svrnm: +1 for keeping "Developer Experience" but _if_ we can find a way to distinguish it great, but it's not a hill I am going to die on, especially since "SIG DevEx" is catchy.

tsloughter: Could Contributor Experience use a different word than Experience maybe? I've not thougt of one that captures the goal of that project but I feel like there is one...

tsloughter:
@dmathieu @julianocosta89 @martinjt are you interested in providing prototypes for particular languages? If so I'd like to note next to your names, in addition to company, what language it is will be covered by your involvement. I'd like to have at least 2 or 3 covered before considering the project personnel needs met.

martinjt:
Happy to take .NET

dmathieu:
I can take Go

julianocosta89:
> @dmathieu @julianocosta89 @martinjt are you interested in providing prototypes for particular languages? If so I'd like to note next to your names, in addition to company, what language it is will be covered by your involvement. I'd like to have at least 2 or 3 covered before considering the project personnel needs met.

I can take Java or Rust

lmolkova (approved):

projects/convenience.md
```diff
@@ -0,0 +1,93 @@
...
+  - Austin Parker (Honeycomb)
+  
+- TC sponsors:
+  - Ted Young (Lightstep)
+  - Liudmila Molkova (Microsoft)
+
+- Maintainers, approvers and contributors:
+  - Damien Mathieu (Elastic)
+  - Juliano Costa (Datadog)
+  - Martin Thwaites (Honeycomb)
```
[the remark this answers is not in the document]
tsloughter: Makes sense, done.

tsloughter:
@martinjt @dmathieu @julianocosta89 ok, great!

I'm adding a part to the doc to cover what @lmolkova brought up about surveying SIGs about existing developer experience improvements they've done and then this should be good to merge and we can move on to slack and setting up a meeting time.

stevejgordon:
I'd be interested in helping with .NET too, if needed.

projects/convenience.md
```diff
@@ -0,0 +1,94 @@
+# Developer Experience Project
```
dmathieu, line 1: The file is still named `convenience.md`. Should it be renamed, maybe to `developer-experience.md`?

tsloughter: yup. will do.

jpkrohling (changes requested):
I'm completely in agreement with the proposal, but I think we need to have consensus about the name. We have a proposal for a "contributor experience" SIG, and it will likely be confusing to have a "developer experience" one as well, especially as a good number of contributors are developers. 

I don't have a good suggestion, but know that I'm not the only one with that concern. Therefore, I'm blocking this until we have consensus.

austinlparker:
> I'm in complete agreement with the proposal, but I think we need to have consensus about the name. We have a proposal for a "contributor experience" SIG, and it will likely be confusing to have a "developer experience" one as well, especially as a good number of contributors are developers. 
> 
> I don't have a good suggestion, but I know that I'm not the only one with that concern. Therefore, I'm blocking this until we have consensus.

"Developer Experience" and "Contributor Experience" are both terms of art, and for the uninitiated they can be explained in a sentence or two.

samsp-msft:
What I am interested in is a group dedicated to the experience for *Application & Library developers* who are consuming OpenTelemetry and using them in their applications. That is a much larger and different audience from those contributing to OpenTelemetry or writing tools such as an APM to consume and display data collected by OpenTelemetry.

## Background
On the .NET team, we have included OpenTelemetry as part of the [Aspire](https://learn.microsoft.com/en-us/dotnet/aspire/get-started/aspire-overview) experience. The Aspire project templates include the OTel libraries, setting up OTLP export and for each component referenced that includes telemetry, we enable it automatically. We added an OTLP endpoint to a dashboard that was originally intended just to show the container running state and have access to the console logs. The [dashboard](https://learn.microsoft.com/en-us/dotnet/aspire/fundamentals/dashboard/explore#structured-logs-page) gives a live view of structured logs, metrics and traces, in views conceptually similar to the more established products. The difference is that it's targeted at showing the incoming data in a live fashion. Data is held in memory, there is no storage, no trends, no alerts, no custom dashboard views a-la Grafana. It's much simpler, and more akin to a Wireshark equivalent for OTLP.

My reason for creating the OTel views in the dashboard was to aid developers with adding their own telemetry - reducing the cost and complexity of setting up the combination of Prometheus, Grafana, Yaeger and a logging viewer. The dashboard is launched automatically as part of the F5 (run) experience from Visual Studio. 

One of the (to me) unexpected lessons from adding the OTel display features of the Aspire dashboard to the Aspire app development experience is that OTel isn't just for analyzing apps in production - it's equally valuable for analyzing apps during the iterative development loop - especially for distributed apps running as a series of containers, as the same kinds of problems of connectivity, latency will show up at development time. Tracing is a great way to verify if services are interacting in the way that you expect. Logs are the way to understand why a container failed.

If you have and use telemetry as part of the application development phase, then you have a much higher chance that you'll be collecting the right telemetry at runtime. 

## Takeaways

If the OpenTelemetry community can provide better tools & defaults to developers to make it easier to use and consume telemetry as part of their regular development cycle, then adding telemetry will become second nature as its used day to day as part of their own inner loop, not just by ops in production. Ensuring good telemetry is not a tax, it enables a better local test and diagnostics cycle.

There were a couple of talks at the community day that touched on the concept of using telemetry as part of the E2E testing for applications. I can imagine collecting and analyzing the telemetry produced by a test run in a CI/CD system, comparing the telemetry from previous runs, and highlighting metrics and span differences from a baseline and then either producing a report or failing the run if it exceeds a set threshold. 

I think having a group focused on the experience for those adding telemetry to their applications will have an overall increase in the satisfaction of the developers using OTel in their solutions.

tsloughter:
I've added a section about the use of first talking to language SIG's to get their initial feedback.

projects/convenience.md
tsloughter: Agree, I added a paragraph before this one that covers this now. Let me know what you think.

tsloughter:
As for the naming, DevEx/DX is a pretty common and well known term I believe. Enough so to be abbreviated often. Looks like "contributor experience" is as well. So I'd argue they should both remain with the same names.

We could consider something like, "End User Experience", but I think that is too broad while we are focused on developers using the API/SDK.

projects/developer_experience.md
```diff
@@ -0,0 +1,104 @@
...
+improvement the next deliverable will be an end user survey. Assuming the agreement and
+participation of the End User SIG this will be done in cooperation with them. A
+write-up of the results will follow, along with the group's decision on the top 3
+issues for the group to work on.
+
+Prototypes of resolutions will be developed and, depending on the scope of the
+identified problem, OTEP's will be created where appropriate before opening PR's
+to the specification.
+
+## Staffing / Help Wanted
```
samsp-msft, line 62: Should this be organized into having reps from each language/SDK? I can see this initiative being a combination of cross-cutting and SDK specific activities/problems to be solved.

samsp-msft:
Do we have infrastructure from the CNCF etc for running surveys. devdiv@Microsoft uses SurveyMonkey as they have good infrastructure and can have org policies for PII management etc. We typically use surveys for two purposes:
- statistical & sentiment analysis. Verbatim responses are useful when categorized to understand high level problem areas
- Collecting contact details for further follow up in zoom/team calls to do a deeper interview as to the customer experience.

The main challenge is getting the survey in-front of the target audience's nose. Github doesn't have good infra for notifying users of a survey (such as having a banner). Pinned discussions depend on the user visiting that area, same for issues.

Second is going to be managing PII, this is an open group, but most participants won't want their information to be spread over the internet - any PII needs to be sandboxed and only used for the stated purpose, and have an expiry.

projects/developer_experience.md
```diff
@@ -0,0 +1,104 @@
...
+frequent request from their users. A report of the findings will be shared with
+the whole community.
+
+Utilizing the knowledge gained from the first deliverable on potential areas of
+improvement the next deliverable will be an end user survey. Assuming the agreement and
+participation of the End User SIG this will be done in cooperation with them. A
+write-up of the results will follow, along with the group's decision on the top 3
+issues for the group to work on.
+
+Prototypes of resolutions will be developed and, depending on the scope of the
```
samsp-msft, line 58: This initiative should be larger than just the API shape of the SDK. Developer experience is not just about writing code that consumes the SDK. Its about the development lifecycle - discovery/learning, tooling, debugging/diagnostics, testing, CI/CD & app deployment. 

Examples:
- How do I enable the built in telemetry to each platform & if there isn't built-in telemetry getting it built
- How do I discover the names/sources for what needs to be enabled to get telemetry and understand the scope of how much data will be generated by the source
- How do I view my telemetry at design-time
  - How do I have different sampling for the inner loop so I get all the telemetry, but have sensible defaults for production deployment
- How can I verify if the right telemetry is being emitted as part of unit tests
- Can I use telemetry as a way of monitoring system behavior as part of integration tests?
- How does OPs know what telemetry is available for the app, know what needs to be configured to enable it and what the dials are to control the volume/granularity of telemetry produced.

austinlparker:
> Do we have infrastructure from the CNCF etc for running surveys. devdiv@Microsoft uses SurveyMonkey as they have good infrastructure and can have org policies for PII management etc. We typically use surveys for two purposes:
> 
> * statistical & sentiment analysis. Verbatim responses are useful when categorized to understand high level problem areas
> * Collecting contact details for further follow up in zoom/team calls to do a deeper interview as to the customer experience.
> 
> The main challenge is getting the survey in-front of the target audience's nose. Github doesn't have good infra for notifying users of a survey (such as having a banner). Pinned discussions depend on the user visiting that area, same for issues.
> 
> Second is going to be managing PII, this is an open group, but most participants won't want their information to be spread over the internet - any PII needs to be sandboxed and only used for the stated purpose, and have an expiry.

We usually just use Google Forms. IIRC CNCF does use SurveyMonkey for stuff, there's probably a way we can get on it. To be quite honest, the biggest challenge the project faces in terms of surveys is (as you intuit) that we do not have a great way to get the survey in front of users other than social media and Slack messages. 

That said, we can work with our vendor community to help broadcast/promulgate surveys and other feedback mechanisms, which we'll probably end up doing in this case. My generic preference is that rather than doing time-limited surveys, we build out better continuous reporting mechanisms, but that's kinda neither here nor there.

Regarding PII, we don't really wind up with PII issues because we don't collect any (I'm not sure we even collect e-mails on our existing surveys).

projects/developer_experience.md
austinlparker: This is an extremely broad list, many of which would be extremely difficult to provide guidance for on a cross-project basis. I'm not saying they're bad ideas, but to take one -- for example, 'verifying if the right telemetry is being emitted as part of unit tests' -- would involve quite a bit of design and research on testing libraries, scaffolding, and patterns across many languages in order to discover the cross-product of necessary hooks that could be recommended for inclusion at a specification level.

austinlparker:
> What I am interested in is a group dedicated to the experience for _Application & Library developers_ who are consuming OpenTelemetry and using them in their applications. That is a much larger and different audience from those contributing to OpenTelemetry or writing tools such as an APM to consume and display data collected by OpenTelemetry.

While I think we are talking about similar things here, I also feel like perhaps we are not. This SIG -- at least, initially -- is designed to focus on the needs of _instrumentors_; Developers who are writing OpenTelemetry code in their applications and libraries. While the other things you mention are important, and are part of DevEx, I think it's important for us to keep the initial scope of this SIG constrained.

If this SIG is successful in these initial goals, then it would seem prudent to expand to other topics, but given the amount of backlog the project currently suffers I believe we need to find some 'quick wins' as it were to address the main ergonomic pain points of instrumentors.

tsloughter:
+1 to everything @austinlparker said.

I completely agree about the importance of the other parts of developer experience, and honestly have looked Aspire as something I want to build for Erlang/Elixir -- like I started doing with Erleans to copy another Microsoft project :) -- but I wanted to purposely keep the scope of this project small to focus the group and not spread us too thin. 

Then, things can change in a future DevEx project (or SIG) after the roaring success of this initial iteration :).

Whether to limit the scope of surveying as well can be left to discussion within the project. Maybe we'll want to still surface other areas of improvement in the initial report but limit our focus of work on the ones related to the spec.

As for discussion of how to perform the surveying I think we can leave that to after we've gotten this merged and take it to the project group?

projects/developer_experience.md
tsloughter: I don't think we can get someone from each. But there was a point to get reps from a few languages, which we have. Currently we have people involved with Java, .Net, Go and Erlang/Elixir.

danielgblanco (approved):
Thanks for the proposal!

danielgblanco:
>As for discussion of how to perform the surveying I think we can leave that to after we've gotten this merged and take it to the project group?

I agree. The End-User SIG has some basic guidance on how surveys are usually conducted https://github.com/open-telemetry/sig-end-user/tree/main/end-user-surveys.

I also think that scoping this SIG to a few actionable items makes a lot of sense. After this initial effort the SIG can re-evaluate if these efforts should continue on a more long-term basis in its current form (like a permanent SIG) or if there's a better way to tackle those issues in a different way, perhaps involving a combination/collaboration between End-User SIG and implementation SIGs directly.

lmolkova:
> I'm in complete agreement with the proposal, but I think we need to have consensus about the name. We have a proposal for a "contributor experience" SIG, and it will likely be confusing to have a "developer experience" one as well, especially as a good number of contributors are developers.

throwing in some alternative naming suggestions:

- Instrumentation Development Experience
- Instrumentation Experience
- Instrumentation Convenience

martinjt:
One of the externally facing goals of this, at least in my eyes, is to give the users of OpenTelemetry confidence that we're making it better for them.

While I can understand the desire to try and have very well segregated and defined names internally, I think it will be detrimental to that goal.

We could say "developer experience is an overarching project covering contributors to the project as well as end users", however, I think that would soften the message too much.

Developer Experience is a term that, external to OSS Contributors, is widely known as making the tool easier to use for the people use it.

In my eyes, we should absolutely stick with this being referred to as the "Developer Experience" project, and not try and segment/qualify it. That distinction can be made in the first paragraph and have the same effect, without diminishing the impact on the other experience projects.

austinlparker:
If anyone has very strong opinions that this project should _not_ be named Developer Experience, I would appreciate their feedback as soon as reasonably possible so that we can close on this and move forward.

jpkrohling:
I need to catch up with this discussion, but I know that both @tedsuo and @svrnm expressed concerns about the name last week. I like @lmolkova's suggestions, having a slight preference for Instrumentation Development Experience, followed by Instrumentation Experience.

austinlparker:
> I need to catch up with this discussion, but I know that both @tedsuo and @svrnm expressed concerns about the name last week. I like @lmolkova's suggestions, having a slight preference for Instrumentation Development Experience, followed by Instrumentation Experience.

Could @tedsuo and @svrnm post their concerns in this PR so we're not playing telephone then?

svrnm:
> > I need to catch up with this discussion, but I know that both @tedsuo and @svrnm expressed concerns about the name last week. I like @lmolkova's suggestions, having a slight preference for Instrumentation Development Experience, followed by Instrumentation Experience.
> 
> Could @tedsuo and @svrnm post their concerns in this PR so we're not playing telephone then?

Apologies for not coming back on this issue earlier, I was at KCD Munich and had some other things to follow up with. 

I do not have a strong concern with it, I just want to make sure that "Developer Experience" how I understand it and what the SIG plans to do is congruent.

A good definition I found for Developer Experience comes from [this github blog post](https://github.blog/2023-06-08-developer-experience-what-is-it-and-why-should-you-care/):

> DevEx refers to the systems, technology, process, and culture that influence the effectiveness of software development. It looks at all components of a developer’s ecosystem—from environment to workflows to tools—and asks how they are contributing to developer productivity, satisfaction, and operational impact.

Another one from the [Microsoft Engineering Fundamentals Playbook](https://microsoft.github.io/code-with-engineering-playbook/developer-experience/):

> Developer experience refers to how easy or difficult it is for a developer to perform essential tasks needed to implement a change. A positive developer experience would mean these tasks are relatively easy for the team (see measures below).
> 
> The essential tasks are identified below.
> 
>    Build - Verify that changes are free of syntax error and compile.
>    Test - Verify that all automated tests pass.
>    Start - Launch end-to-end to simulate execution in a deployed environment.
>    Debug - Attach debugger to started solution, set breakpoints, step through code, and inspect variables.
>
> If effort is invested to make these activities as easy as possible, the returns on that effort will increase the longer the 
> project runs, and the larger the team is.

If I read the proposal the focus is on the API/SDK "easy of use" or providing convenience functions to use them:

> The first deliverable will be the collection of experience in dealing with
developer experience issues by each existing language SIG. This means not only
additions to the API/SDK or libraries developed to enhance the experience for
users but any that may be planned or are being thought about because of a
frequent request from their users.

Based on what people commonly understand under "tools" for developer experience I see the following issue:

Tools is much bigger than what the language APIs/SDKs are doing, and also from the definitions I found and the understand I had, tools people have in mind are IDEs, AI assistance, Automation Tooling for Tests, Helpers for Debugging, maybe even the chair they are sitting on etc. Based on [this comment](https://github.com/open-telemetry/community/pull/2144/files#r1657736386) by @samsp-msft and the answer by @austinlparker this seems not to be in scope (initially) and there is the desire to keep the scope small. 

So the SIG needs to ask themselves if/how to manage expectations if people **assume** that all of this is part of what they are going to do? Is this SIG going to look into making OpenTelemetry easier to build, test, start and debug?

If the answers to those questions is "Yes" I have no further objections.

austinlparker:
> So the SIG needs to ask themselves if/how to manage expectations if people **assume** that all of this is part of what they are going to do? Is this SIG going to look into making OpenTelemetry easier to build, test, start and debug?
> 
> 
> 
> If the answers to those questions is "Yes" I have no further objections.

If those are what's needed to make the developer experience better, sure? Better APIs isn't just "there should be declarative span creation sugar", it could also be "APIs should be easier to test" or "SDK configuration should be easier"

The point of calling this Developer Experience is in the spirit of keeping our options open. If we said "oh this is just instrumentation convenience" and there's overwhelming feedback that the biggest problem is, like, SDK initialization or testing, then we'd be artificially limiting ourselves. 

Now, do I think that those will be the biggest pain points or the first thing being tackled? No, I suspect that the biggest DevEx problems we have right now are around the actual instrumentation API, but I don't want to presume things that aren't backed by data.

svrnm:
> > So the SIG needs to ask themselves if/how to manage expectations if people **assume** that all of this is part of what they are going to do? Is this SIG going to look into making OpenTelemetry easier to build, test, start and debug?
> > If the answers to those questions is "Yes" I have no further objections.
> 
> If those are what's needed to make the developer experience better, sure? Better APIs isn't just "there should be declarative span creation sugar", it could also be "APIs should be easier to test" or "SDK configuration should be easier"
> 
> The point of calling this Developer Experience is in the spirit of keeping our options open. If we said "oh this is just instrumentation convenience" and there's overwhelming feedback that the biggest problem is, like, SDK initialization or testing, then we'd be artificially limiting ourselves.
> 
> Now, do I think that those will be the biggest pain points or the first thing being tackled? No, I suspect that the biggest DevEx problems we have right now are around the actual instrumentation API, but I don't want to presume things that aren't backed by data.

That sounds good to me, but that's not what I read from the proposal yet. Since this already has a lot of approvals I don't think that has to be changed, and at the end I will also not insist on calling this SIG differently, I just wanted (and still want) to call out that the name "Developer Experience" has a very very broad scope and SIG members need to be prepared to push back with many requests in that domain that they find out of scope, either point in time or general. To give another example that for me is within "DevEx": what if developers think that the biggest problem is that there is no developer friendly OTLP endpoint, will something like [OTEP 230](https://github.com/open-telemetry/oteps/pull/230) be in scope?

austinlparker:
> That sounds good to me, but that's not what I read from the proposal yet. Since this already has a lot of approvals I don't think that has to be changed, and at the end I will also not insist on calling this SIG differently, I just wanted (and still want) to call out that the name "Developer Experience" has a very very broad scope and SIG members need to be prepared to push back with many requests in that domain that they find out of scope, either point in time or general. To give another example that for me is within "DevEx": what if developers think that the biggest problem is that there is no developer friendly OTLP endpoint, will something like [OTEP 230](https://github.com/open-telemetry/oteps/pull/230) be in scope? 

That wouldn't be in scope because it's quite clearly something that isn't related to the developer experience of OpenTelemetry itself. Out of scope work for the project doesn't magically become in scope because a SIG was formed. 

Respectfully, this naming discussion feels like bikeshedding. The initial charter for the SIG in the proposal is clear and concise, and the name 'Developer Experience' clearly distinguishes the goals and purview of the solution space. In the event that this SIG is wildly successful, perhaps it will continue to deliver improvements on Developer Experience in other areas of the field -- there are certainly many of those. However, I continue to see no clear rationale to change the name to something that potentially restricts the solution space and scope of this project before we've even had a chance to do discovery on what problems should be tackled.

svrnm:
> Respectfully, this naming discussion feels like bikeshedding. 

I understand that you see it that way and I apologize for getting that deep into this discussion. 

I was asked to express my concerns publicly and to unblock this PR, so I did exactly that:

> I do not have a strong concern with it, I just want to make sure that "Developer Experience" how I understand it and what the SIG plans to do is congruent.

I got my answer, so from my site there is no further need to discuss this. I approved the PR already with the initial name and so I only can say that my approval remains.

austinlparker:
Sorry, that was less directed at you and more the generic reader.

jpkrohling (approved):
I'm approving this as well: I think we had enough discussions around the topic and the naming, and there's no better name for this SIG. The potential confusion with Contributor Experience can certainly be clarified quickly when it happens.

projects/developer_experience.md
jpkrohling: Agree with @tsloughter. SIG representatives are welcome, but not required.

projects/developer_experience.md
jpkrohling: I'm happy with the current proposed scope for this SIG, and it can grow in the future depending on how the needs evolve. I'm marking this as resolved, but please open an issue if you think this needs to be discussed further.

mtwo (approved):

tsloughter:
I don't see any unresolved comments but this is still blocked on 3 reviews.

jpkrohling:
I'm merging, as I can count 5 GC approvals already and no concerns from the other 4.

PR #2184, opened 2024-07-04
Split Collector SIG meeting times

jpkrohling:
This partially implements the proposal from open-telemetry/opentelemetry-collector#7085.

cc @open-telemetry/collector-approvers 


sigs.yml
```diff
@@ -350,8 +350,7 @@
     - name: Trask Stalnaker
       github: trask
   - name: Collector
-    meeting: Every Wednesday at 09:00 PT plus monthly on first Wednesday at 00:00 PT
-      and on third Wednesday at 16:00 PT
+    meeting: "Wednesdays, rotating between the following time slots: 00:00 UTC, 12:00 UTC, 16:00 UTC"
```
svrnm, line 353: I know this is counter the reason why this is done, but right now all meetings are scheduled as PT, we should look into that from a larger perspective, but would you mind to set up the meetings as "PT" for now? This is especially important for daylight saving time periods as they are regionally very different. UTC would mean that meeting times change 2 per year for those who have daylight saving times.

jpkrohling:
I'm marking this as draft, as the official proposal is this: https://github.com/open-telemetry/opentelemetry-collector/pull/10536

sigs.yml
svrnm: I do not disagree that the way we write it down is not good for people outside the US, since I am affected as well, but it's scheduled that way right now. I am OK with this being the first UTC time, however when this gets scheduled the invite needs to be set as "UTC" as well such that now daylight saving times are respected, and we should also work towards fixing the other ones as well.

jpkrohling: I commit to setting the calendar entries to UTC, and will circle back after 6 months with the results of the "experiment". If we are happy with the results, we can start applying the same changes to other SIGs.

svrnm: thanks!

svrnm (approved):

mx-psi (approved):
I am in favor of trying this out for 6 months

songy23 (approved):

README.md
```diff
@@ -155,7 +155,7 @@ SIG-specific GitHub discussions.
...
-| Collector | Every Wednesday at 09:00 PT plus monthly on first Wednesday at 00:00 PT and on third Wednesday at 16:00 PT | [Google Doc](https://docs.google.com/document/d/1r2JC5MB7GupCE7N32EwGEXs9V_YIsPgoFiLP4VWVMkE) | [#otel-collector](https://cloud-native.slack.com/archives/C01N6P7KR6W) | [calendar-collector](https://groups.google.com/a/opentelemetry.io/g/calendar-collector) | [Juraci Paixão Kröhling](https://github.com/jpkrohling) |
+| Collector | Wednesdays, rotating between the following time slots: 00:00 UTC, 12:00 UTC, 16:00 UTC | [Google Doc](https://docs.google.com/document/d/1r2JC5MB7GupCE7N32EwGEXs9V_YIsPgoFiLP4VWVMkE) | [#otel-collector](https://cloud-native.slack.com/archives/C01N6P7KR6W) | [calendar-collector](https://groups.google.com/a/opentelemetry.io/g/calendar-collector) | [Juraci Paixão Kröhling](https://github.com/jpkrohling) |
```
arminru, line 158: As if scheduling across time zones wouldn't already bring enough headaches, you'll also have to watch out for daylight savings time switches happening on different dates. The offset between regions will differ by +/-1h for 1-3 weeks each twice a year. To switch properly, this needs to be set up accordingly in the Google calendar entry by pinning it to UTC and not PT with US DST dates as we usually have it.

arminru (approved):

README.md
jpkrohling: Yes, I'll change the entries to use UTC instead.

codeboten (approved):

MovieStoreGuy (approved):

jpkrohling:
I'm merging this as I believe we have enough buy in from the community, including Collector SIG leadership.

jpkrohling:
Calendar entries changed to reflect this: we now have three entries alternating the times and the time is set to UTC.

jpkrohling:
Apparently, I removed only my own copy of that 🤦🏽 It should be gone now from the community calendar as well.

mx-psi:
Looks good now!  :rocket:

PR #2196, opened 2024-07-08
Clarify maintainer role in github teams

trask:
I think this is the current practice but isn't documented and came up in https://cloud-native.slack.com/archives/C01NJ7V1KRC/p1720257247004479?thread_ts=1720120170.039779&cid=C01NJ7V1KRC


svrnm (commented):
Thanks! We might want to review all the teams if they follow that structure or if any need fixing

svrnm (approved):

arminru (approved):

danielgblanco (approved):

danielgblanco:
Should we merge this and then go ahead and make sure all teams are setup like that? I think needs a TC member to do it.

tsloughter:
What @danielgblanco said. For example, I don't have permission to add maintainers to `erlang-contrib-maintainers`

carlosalberto (approved):

trask:
> Should we merge this and then go ahead and make sure all teams are setup like that? I think needs a TC member to do it.

I created #2205

trask:
> I don't have permission to add maintainers to `erlang-contrib-maintainers`

@tsloughter you should have permission now

PR #2197, opened 2024-07-09
Add Java SIG APAC meeting time

trask:
(added to public calendar already)


arminru (approved):

svrnm (approved):

sigs.yml
```diff
@@ -449,7 +449,7 @@
     - name: 'Juraci Paixão Kröhling'
       github: jpkrohling
   - name: 'Java: SDK + Automatic Instrumentation'
-    meeting: Every Thursday at 09:00 PT
+    meeting: Every Thursday at 09:00 PT with an additional meeting every other Thu at 09:00 China Standard Time
```
steverao, line 452: Is it better to replace `China Standard Time` by `CST (UTC+8)` or `UTC+8` here? Because, I guest maybe some guys that come from other asia countries are more familiar with latter.

trask: thanks! I'll update it to UTC+8 (CST means US Central Standard Time in my brain 😅)

PR #2201, opened 2024-07-09
Document GitHub-hosted runners in assets.md

trask:
Opened https://cncfservicedesk.atlassian.net/servicedesk/customer/portal/1/CNCFSD-2378 to find out from the CNCF what kind of billing limits are in place. I think we can merge without this info, and I can send follow-up PR once we find out.


assets.md
```diff
@@ -36,6 +36,27 @@ Link: https://console.equinix.com/projects/6f1c9af6-0470-42da-8f22-59d0df245f6b
...
+with various computing power
+- 2-core
+- 4-core
+- 8-core
+- 16-core
+- 32-core
+- 64-core
+
+In particular, we have created a 4-core Linux ARM64 runner named `otel-linux-arm64`
+with a concurrency limit of 20 jobs that repositories can use as needed.
```
arminru, line 56:
```suggestion
In particular, we have created a 4-core Linux ARM64 runner named `otel-linux-arm64`
with a concurrency limit of 20 jobs (at the time of writing) that repositories can use as needed.
```
Or will keep this updated in case changes are applied?

assets.md
```diff
@@ -36,6 +36,27 @@ Link: https://console.equinix.com/projects/6f1c9af6-0470-42da-8f22-59d0df245f6b
   other assets on
 - Admin: Juraci Paixão Kröhling @jpkrohling and OTel TC via the mailing list address
 
+### GitHub-hosted runners
+
+We have the ability to create pay-as-you-go GitHub runners for
```
arminru, line 41:
```suggestion
We have the ability to create pay-as-you-go [GitHub runners](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners) for
```

arminru (approved):

pellared (approved):

dmathieu:
Should we remove the `docs/using-actuated.md` doc to avoid confusion and better indicate this is the recommended way of getting ARM runners?

pellared:
> Should we remove the `docs/using-actuated.md` doc to avoid confusion and better indicate this is the recommended way of getting ARM runners?

Are we not having more actuated runners available for us?

svrnm (approved):

PR #2230, opened 2024-07-19
Add meeting hours for contributor experience SIG

svrnm:
This PR adds the meeting hours for SIG Contributor Experience

cc @musingvirtual @jpkrohling @JamieDanielson @kaylareopelle @codefromthecrypt @mx-psi @theletterf @jberkus @maryliag


README.md
```diff
@@ -182,7 +182,7 @@ SIG-specific GitHub discussions.
...
-| Contributor Experience | TBA | [Google Doc](https://docs.google.com/document/d/1CTQI0p3QF8JP8reV8z_ggcs8KE5YVPpQGvAQknw4qP0) | [#otel-contributor-experience](https://cloud-native.slack.com/archives/C06TMJ2R0SK) | [calendar-contributor-experience](https://groups.google.com/a/opentelemetry.io/g/calendar-contributor-experience) | [Juraci Paixão Kröhling](https://github.com/jpkrohling) |
+| Contributor Experience | Every Monday alternating between 9:00 and 17:00 UTC | [Google Doc](https://docs.google.com/document/d/1CTQI0p3QF8JP8reV8z_ggcs8KE5YVPpQGvAQknw4qP0) | [#otel-contributor-experience](https://cloud-native.slack.com/archives/C06TMJ2R0SK) | [calendar-contributor-experience](https://groups.google.com/a/opentelemetry.io/g/calendar-contributor-experience) | [Juraci Paixão Kröhling](https://github.com/jpkrohling) |
```
trask, line 185: 17:00 UTC will conflict with maintainer meeting once daylight savings time ends, maybe worth specifying it as 10:00 PT?

trask (approved):

maryliag (approved):

README.md
mx-psi: Makes sense to me to change this to PT

mx-psi (approved):

danielgblanco (approved):

README.md
danielgblanco: I'd personally prefer UTC on everything but I understand that's a lot easier for me (UK based) than the majority of maintainers :) I'd opt for uniformity in timezones used though, with PT being the _de facto_ standard, or at least have a primary one listed in PT and then maybe local one if maintainers are based elsewhere.

svrnm: @danielgblanco my understanding is that the problem is less around UTC, but around which daylight saving time adjustment we apply to our meetings, and there are no clear winners here, we choose PT so far, because most contributors are US-based. But depending on what we pick we always get different problems:

* If we pick UTC, everyone without DST adjustments in their timezone has their meeting moved twice per year
* If we pick a specific DST adjustment (currently PT), everyone within that timezone has no changes, everyone with a different schedule has 4 changes, and everyone without an adjusment in their timezone has moved their meeting twice per year. 

We can now choose to pick "PT" for the "US-friendly" meetings and a different time zone for "*-friendly" meetings

https://en.wikipedia.org/wiki/Daylight_saving_time_by_country has a nice picture to illustrate that mess.

arminru (approved):

austinlparker (approved):

README.md
trask: > We can now choose to pick "PT" for the "US-friendly" meetings and a different time zone for "*-friendly" meetings

this is what we did for Java: #2200

JamieDanielson (approved):

README.md
codefromthecrypt: I completely agree with the UTC+8. For example, I'm going to the LLM Sig tomorrow and I had to do a double-take as it is 7am not 8am at the moment. Ack there are some pragmatic impacts to this wrt attendance from west coast folks, just it is easier to get people to attend in the native time slot if they know what time it is without calculating backwards from california.

svrnm: https://github.com/open-telemetry/community/issues/2237

svrnm:
Merging this now and fixing the timezones later based on the outcome of https://github.com/open-telemetry/community/issues/2237 (because this also means I need to update the google invites)

PR #2236, opened 2024-07-25
Update LLM SemConv Asia/Europe WG meeting time

drewby:
LLM Semconv is rescheduling series from 4pm to 11pm each other Wednesday starting from August 7th to accommodate more time zones in Asia.

cc @nirga @lmolkova @cartermp


lmolkova (approved):

sigs.yml
```diff
@@ -150,7 +150,7 @@
     - name: Severin Neumann
       github: svrnm
   - name: 'Semantic Conventions: LLM'
-    meeting: Every Wednesday alternating between 10:00 and 16:00 PT
+    meeting: Every Wednesday alternating between 10:00 and 23:00 PT
```
svrnm, line 153: We had this discussion now for a few other SIGs as well, would it make more sense to express the Asia/Europe meeting time in a different time zone that suits the people attending that meeting? PT means that daylight saving time is adjusted for people in PT, but people in areas without DST or a different DST will have their meeting times changed 2 (or 4) times.

svrnm (approved):
my comment on timezones is non-blocking, since there are other entries that need to be reviewed and updated.

arminru (approved):

drewby:
> my comment on timezones is non-blocking, since there are other entries that need to be reviewed and updated.

Thanks! I'm happy to update it later if this is the agreed practice.

svrnm:
> > my comment on timezones is non-blocking, since there are other entries that need to be reviewed and updated.
> 
> Thanks! I'm happy to update it later if this is the agreed practice.

Feel free to take a look at https://github.com/open-telemetry/community/issues/2237 in the meantime, I can only express my perspective as someone who lives in CET-timezone and works a lot with PT-timezone, so my suggestion might be wrong for people in different timezones.

Issue #2264, opened 2024-08-07
Update which Collector meeting times are pinned to the Pacific Time zone 


evan-bradley:
This is a follow up to https://github.com/open-telemetry/community/pull/2249.

After this PR, the Collector has three meeting times:
* 09:00 PT, which is 12:00 ET, 18:00 CET, and 00:00 or 01:00 UTC+8 depending on DST. This meeting is primarily US-centric, but also sees frequent attendance from European Collector leads and community members.
* 17:00 PT, which is 20:00 ET, 02:00 CET, and 08:00 or 09:00 UTC+8 depending on DST. This meeting is intended for meeting between those in PT and UTC+8, as it is outside normal working hours for those in ET and CET.
* 20:00 UTC+8, which is 08:00 ET, 14:00 CET, and 05:00 PT during DST. This meeting is primarily intended to be Europe-friendly, but also sees frequent attendance from Collector leads and community members in ET.

Since the 20:00 UTC+8 time is outside of normal APAC working hours and is intended for those in ET and CET, I would like to propose that we pin it to PT. PT has daylight savings time and will make the time stable for those in ET and CET, outside any brief differences in when DST is in effect.

cc @open-telemetry/collector-triagers @open-telemetry/collector-contrib-triagers

svrnm:
@evan-bradley, thanks for clarification. Can you provide a PR to revert the change I applied, I can also do it, just let me know

evan-bradley:
Appreciate you taking a look, I opened a PR with the way I think we should define the times.

PR #2271, opened 2024-08-09
Pin all Collector SIGs to PT

evan-bradley:
This ensures the times stay the same for the places where most contributors are from. In particular, for the 05:00 PT meeting, this keeps the time consistent for those in the Eastern and European time zones.

I've also reordered the times so they match the rotation order.

cc @svrnm @open-telemetry/collector-triagers @open-telemetry/collector-contrib-triagers 

Resolves https://github.com/open-telemetry/community/issues/2264


svrnm (approved):

songy23 (approved):

arminru (approved):

trask:
@evan-bradley it looks like the timezone on the public calendar events needs to be updated as well

PR #2297, opened 2024-08-26
Update GC liaison for CI/CD semconv SIG

trask:


danielgblanco (approved):
Thanks Trask. I'm happy to be GC liaison for the CI/CD semconv SIG.

arminru (approved):

PR #2392, opened 2024-10-10
Remove "Candidate Name" from the list of candidates, don't think they qualify.

bogdandrutu:


jack-berg (approved):

svrnm (approved):

alolita (approved):
+1

elections/2024/governance-committee-candidates.md
```diff
@@ -4,7 +4,6 @@
 
 In alphabetical order:
 
-- [Candidate Name](#candidate-name)
```
danielgblanco, line 7: There's also a block below that would have to be removed too (it was the template)

bogdandrutu: Are nomination done? I can remove that as well

tedsuo (approved):

elections/2024/governance-committee-candidates.md
danielgblanco: Yeah, I think this would have to removed eventually, but if we remove the link to the template here we may as well remove the template body below? I think it should already be understood what's needed considering other candidates. Especially considering we have the [guidance](https://github.com/open-telemetry/community/blob/main/elections/2024/governance-committee-election.md#nominations).

danielgblanco: Nominations finish on Friday at midnight.

PR #2409, opened 2024-10-24
Project Proposal: Audit Logging SIG

mlenkeit:
This PR contains a project proposal for an Audit Logging SIG as [discussed on Slack](https://cloud-native.slack.com/archives/CJFCJHG4Q/p1719840258022029).

We are aware that the project proposal still has several _tbd_'s especially with regard to staffing and timeline that need to be defined before the SIG can start working.

We will approach other vendors directly with this proposal to identify additional contributors. Of course, anyone who comes across this proposal here on GitHub is invited to contribute.

While we do have some ideas about a potential timeline for semantic conventions, OTEL SDK/API and collector adjustments respectively, we would like to align this with other contributors first before publishing.

Any feedback from the community on the proposed scope of the SIG is highly appreciated!

### Open topics

The following items reference topics from the PR discussion that are still open:

- immutability, tamper-proof logs, signing - in scope?
  - https://github.com/open-telemetry/community/pull/2409#issuecomment-2438469592
  - https://github.com/open-telemetry/community/pull/2409#issuecomment-2485820262
  - https://github.com/open-telemetry/community/pull/2409#issuecomment-2486409314


projects/audit-logging.md
```diff
@@ -0,0 +1,68 @@
...
+- permission changes (e.g. of a service account or application user)
+- modification of data
+- accessing sensitive information
+- failed login attempts
+
+### Current challenges
+
+Audit Logging is currently not within the scope of OpenTelemetry
+
+- no semantic conventions for audit logs in OTEL
```
mtwo, line 19:
```suggestion
- There aren't currently any semantic conventions designed specifically for audit logs in OTEL
```

mtwo:
Are there any requirements around signing logs / detecting tampering? I've heard that mentioned before in the context of audit logs, but I don't know how common of a requirement it is

projects/audit-logging.md
```diff
@@ -0,0 +1,68 @@
+# Audit Logging
+
+## Background and description
+
+Audit logging describes the capability of capturing audit-trail relevant events of a system to meet compliance requirements. Such events may originate from the infrastructure (e.g. a Kubernetes cluster) up to the application-level. It is a capability that is particularly relevant for providers of enterprise software.
+
+Unlike regular application logs, audit logs are usually subject to long retention periods and software providers must guarantee their completeness (i.e. guarantee of delivery).
```
reyang, line 7: Good points! In addition, these are something we might want to consider:
1. Audit logs might be considered as a critical part of the business, which could result in a different API design strategy - for example, audit logging might require a different API behavior, if the information provided by the caller is invalid, the API might throw exception instead of failing silently and move on.
2. Audit log might require some sensitive information without redaction due to the regulation requirements (e.g. user identity and client IP address).
3. The data path could require higher level of access control or privilege.

projects/audit-logging.md
```diff
@@ -0,0 +1,68 @@
...
+
+Examples of audit logs include:
+- permission changes (e.g. of a service account or application user)
+- modification of data
+- accessing sensitive information
+- failed login attempts
+
+### Current challenges
+
+Audit Logging is currently not within the scope of OpenTelemetry
```
reyang, line 17: I have a different perspective here, I think OpenTelemetry doesn't have a good solution for audit logging, but audit logging is definitely within the current scope (it is still a type of logs / timestamped events).

projects/audit-logging.md
```diff
@@ -0,0 +1,68 @@
...
+
+- no semantic conventions for audit logs in OTEL
+- OTEL APIs/SDKs do not provide feedback to the application level whether data (in particular logs) have been successfully delivered to a remote endpoint. To guarantee delivery, either the SDK has to give those guarantees, or provide feedback to the application so that it can take care of guaranteed delivery itself.
```
reyang, line 20: https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification#project-naming

```suggestion
- OTel APIs/SDKs do not provide feedback to the application level whether data (in particular logs) have been successfully delivered to a remote endpoint. To guarantee delivery, either the SDK has to give those guarantees, or provide feedback to the application so that it can take care of guaranteed delivery itself.
```

projects/audit-logging.md
```diff
@@ -0,0 +1,68 @@
...
+## Deliverables
+
+- semantic convention for audit logs
+- extend OTEL APIs/SDKs for audit logging purposes (in collaboration with the respective SIG)
+- extend OTEL collector for audit logging purposes (in collaboration with the respective SIG)
+
+## Staffing / Help Wanted
+
+The following vendors are interested in improving this area:
+- SAP
```
reyang, line 41: You can put me here, I'm from Microsoft, I'm currently a Technical Committee member of OpenTelemetry, and I work on audit logs as part of my job (I don't own the Microsoft audit logging story, but I work with the team closely).

projects/audit-logging.md
```diff
@@ -0,0 +1,68 @@
...
+
+Other vendors are invited to join the discussion.
+
+### Required staffing
+
+* Project lead: SAP (name tbd)
+* Sponsors: tbd
+* GC liaison: tbd
+* Engineers:
+  * SAP will provide a prototype in two languages (tbd; likely two of Java, JavaScript, Go)
```
reyang, line 51: I think we need prototype in two parts:
1. API/SDK - this is where we need three programming languages IIRC.
2. OTel Collector - higher guarantee on data delivery (completeness, integrity, latency, etc.), data path security.

projects/audit-logging.md
```diff
@@ -0,0 +1,68 @@
...
+
+The following vendors are interested in improving this area:
+- SAP
+
+Other vendors are invited to join the discussion.
+
+### Required staffing
+
+* Project lead: SAP (name tbd)
+* Sponsors: tbd
```
reyang, line 48: I can sponsor this.

projects/audit-logging.md
```diff
@@ -0,0 +1,68 @@
...
+- OTEL APIs/SDKs do not provide feedback to the application level whether data (in particular logs) have been successfully delivered to a remote endpoint. To guarantee delivery, either the SDK has to give those guarantees, or provide feedback to the application so that it can take care of guaranteed delivery itself.
+- OTEL collectors may lose audit logs in transit (i.e. no guarantee of delivery)
+
+### Goals, objectives, and requirements
+
+The goal of this project is to make OTEL fit for audit logging purposes that meet compliance requirements of enterprise software providers, in particular:
+
+- REQ-CONV-01: Semantic conventions for application-level audit logs are defined
+- REQ-CONV-02: Semantic conventions for infrastructure-level audit logs are defined
+- REQ-APPL-01: Guaranteed delivery of audit logs exported via OpenTelemetry SDK.
```
reyang, line 29: nit: remove the "." or add "." to the other items

projects/audit-logging.md
```diff
@@ -0,0 +1,68 @@
...
+- REQ-PIPE-01: OTEL collector must provide guaranteed delivery of audit logs, including when its process is interrupted
```
reyang, line 30: I think we should clarify/define "guaranteed delivery".

projects/audit-logging.md
```diff
@@ -0,0 +1,68 @@
...
+- accessing sensitive information
+- failed login attempts
+
+### Current challenges
+
+Audit Logging is currently not within the scope of OpenTelemetry
+
+- no semantic conventions for audit logs in OTEL
+- OTEL APIs/SDKs do not provide feedback to the application level whether data (in particular logs) have been successfully delivered to a remote endpoint. To guarantee delivery, either the SDK has to give those guarantees, or provide feedback to the application so that it can take care of guaranteed delivery itself.
+- OTEL collectors may lose audit logs in transit (i.e. no guarantee of delivery)
```
mlenkeit, line 21:
```suggestion
- OTel collectors may lose audit logs in transit (i.e. no guarantee of delivery)
```

projects/audit-logging.md
```diff
@@ -0,0 +1,68 @@
...
+
+### Goals, objectives, and requirements
+
+The goal of this project is to make OTEL fit for audit logging purposes that meet compliance requirements of enterprise software providers, in particular:
```
mlenkeit, line 25:
```suggestion
The goal of this project is to make OTel fit for audit logging purposes that meet compliance requirements of enterprise software providers, in particular:
```

projects/audit-logging.md
```diff
@@ -0,0 +1,68 @@
...
+- REQ-CONV-01: Semantic conventions for application-level audit logs are defined
+- REQ-CONV-02: Semantic conventions for infrastructure-level audit logs are defined
+- REQ-APPL-01: Guaranteed delivery of audit logs exported via OpenTelemetry SDK.
+- REQ-PIPE-01: OTEL collector must provide guaranteed delivery of audit logs, including when its process is interrupted
+
+## Deliverables
+
+- semantic convention for audit logs
+- extend OTEL APIs/SDKs for audit logging purposes (in collaboration with the respective SIG)
+- extend OTEL collector for audit logging purposes (in collaboration with the respective SIG)
```
mlenkeit, line 36:
```suggestion
- extend OTel APIs/SDKs for audit logging purposes (in collaboration with the respective SIG)
- extend OTel collector for audit logging purposes (in collaboration with the respective SIG)
```

projects/audit-logging.md
mlenkeit, line 19:
```suggestion
- no semantic conventions for audit logs in OTel
```

projects/audit-logging.md
mlenkeit: @reyang thanks for mentioning these points.

Especially the API behavior is something that we had thought about initially. However, when we first pitched audit logging [on Slack](https://cloud-native.slack.com/archives/CJFCJHG4Q/p1719840258022029), we received the following comment from Ted Young:

> As a rule, the OpenTelemetry API never throws an exception. I understand why you might want this, though it is not present in many audit logging systems, which use regular loggers. So a strong case would have to be made on this particular point.

Based on this initial feedback, we decided to file this SIG proposal without proposing API changes.

mlenkeit:
> Are there any requirements around signing logs / detecting tampering? I've heard that mentioned before in the context of audit logs, but I don't know how common of a requirement it is

@mtwo for all I know, immutability of audit logs is a common requirement although not all audit logging systems/use cases that I've seen address this requirement with technical measures but sometimes also organizational measures. However, given the flexibility of OTel processing queues (i.e. different topologies of collectors), having a technical solution in OTel would be favorable.

@reyang what is your opinion on this?

projects/audit-logging.md
mlenkeit: Thanks for pointing this out! It's clear to us, but I'll make this clearer in the doc...

projects/audit-logging.md
reyang: 👍

projects/audit-logging.md
```diff
@@ -0,0 +1,71 @@
...
+
+Other vendors are invited to join the discussion.
+
+### Required staffing
+
+* Project lead: SAP (name tbd)
+* Sponsors:
+  - Reiley Yang (@reyang)
+  - tbd
+* GC liaison: tbd
```
svrnm, line 52: I volunteer for being the liaison if this is going to become a SIG

reyang:
> > Are there any requirements around signing logs / detecting tampering? I've heard that mentioned before in the context of audit logs, but I don't know how common of a requirement it is
> 
> @mtwo for all I know, immutability of audit logs is a common requirement although not all audit logging systems/use cases that I've seen address this requirement with technical measures but sometimes also organizational measures. However, given the flexibility of OTel processing queues (i.e. different topologies of collectors), having a technical solution in OTel would be favorable.
> 
> @reyang what is your opinion on this?

@mlenkeit I think this can be achieved as long as OpenTelemetry is designed to allow additive changes, doesn't have to be there in the first place. I personally haven't seen people signing logs, and I've seen lots of cases where immutable data path is used.

projects/audit-logging.md
mlenkeit: Thank you! I've updated the doc accordingly https://github.com/open-telemetry/community/pull/2409/commits/3876a315eaad74539bc52110f8a430a985acdb55.

projects/audit-logging.md
```diff
@@ -0,0 +1,71 @@
...
+- accessing sensitive information
+- failed login attempts
+
+### Current challenges
+
+Audit Logging is currently not within the scope of OpenTelemetry
+
+- no semantic conventions for audit logs in OTel
+- OTel APIs/SDKs do not provide feedback to the application level whether data (in particular logs) have been successfully delivered to a remote endpoint. To guarantee delivery, either the SDK has to give those guarantees, or provide feedback to the application so that it can take care of guaranteed delivery itself.
+- OTel collectors may lose audit logs in transit (i.e. no guarantee of delivery)
```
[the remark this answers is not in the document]
mlenkeit: @svrnm I've added a chapter to the appendix on guarantee of delivery in the context of audit logging: https://github.com/open-telemetry/community/pull/2409/files#diff-736e6b0ae9ae655b78d9ba007d08592071abb6cc1ef64d7893ff81642c8ec734R79-R113.

I think this capability should be built such that it can be enabled for audit logs if required while other signals continue to be processed/delivered just like they are today.

projects/audit-logging.md
mlenkeit: @reyang I've added a chapter to the appendix on guarantee of delivery in the context of audit logging: https://github.com/open-telemetry/community/pull/2409/files#diff-736e6b0ae9ae655b78d9ba007d08592071abb6cc1ef64d7893ff81642c8ec734R79-R113.

projects/audit-logging.md
```diff
@@ -0,0 +1,71 @@
...
+- permission changes (e.g. of a service account or application user)
+- modification of data
+- accessing sensitive information
+- failed login attempts
+
+### Current challenges
+
+Audit Logging is currently not within the scope of OpenTelemetry
+
+- no semantic conventions for audit logs in OTel
```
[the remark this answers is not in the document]
mlenkeit: @svrnm our experience has shown that in order to analyze audit logs at scale, it is important to define an (extensible) event catalog. The event catalog standardizes audit log events across workloads/produces. For example, our internal event catalog currently consists of 50+ such events. Ideally, such a catalog would be part of semantic conventions.

To make this more tangible, I've added some examples to the appendix of the document:
https://github.com/open-telemetry/community/pull/2409/files#diff-736e6b0ae9ae655b78d9ba007d08592071abb6cc1ef64d7893ff81642c8ec734R115-R192

projects/audit-logging.md
```diff
@@ -0,0 +1,196 @@
...
+
+### Appendix A: Guarantee of Delivery
+
+In the context of this document, guarantee of delivery describes the ability of delivering audit logs from source to destination through OTel means while ensuring that all such signals arrive at the destination and/or providing the source with a means to handle failed delivery.
+
+Messaging protocols that support different levels of delivery guarantees may refer to this behavior as _at least once_ or _exactly once_, as opposed to _at most once_.
+
+We assume that every component that is involved in the delivery of audit logs from source to destination must support guarantee of delivery individually, rather than assuming that this ability can be provided by e.g. only the collector or SDK.
+
+The implications of guarantee of delivery can be illustrated with an example consisting of a workload, an OTel collector, and a durable storage. The workload acts as the source and produces audit logs via the OTel SDK/API. It writes the data via OTLP to the OTel collector. The OTel collector is configured to export audit logs to a durable storage that acts as the destination such as an S3 bucket.
```
reyang, line 87: Minor, it is a bit weird to see "SDK/API", I think folks normally use "API/SDK".

projects/audit-logging.md
```diff
@@ -0,0 +1,196 @@
...
+- accessing sensitive information
+- modification of data
+
+### Current challenges
+
+OpenTelemetry does not have a good solution for audit logging
+
+- no semantic conventions for audit logs in OTel
+- OTel APIs/SDKs do not provide feedback to the application level whether data (in particular logs) have been successfully delivered to a remote endpoint. To guarantee delivery, either the SDK has to give those guarantees, or provide feedback to the application so that it can take care of guaranteed delivery itself.
+- OTel collectors may lose audit logs in transit (i.e. no guarantee of delivery)
```
reyang, line 21: Does "OTel collectors" mean the "OpenTelemetry Collector" (https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification#project-naming) or "any collectors that can handle OpenTelemetry data (whether OTLP or something else)"?

projects/audit-logging.md
```diff
@@ -0,0 +1,196 @@
...
+
+The following implications would apply:
+
+- workload produces an audit-relevant event:
+
+  The workload emits the event via the OTel SDK/API. It may wait for acknowledgement of receipt from the collector before proceeding. If the event is rejected or receipt is not acknowledged in time, the workload or SDK may act accordingly, e.g. retry, rollback a database transaction, inform the user, etc.
+
+- OTel collector receives the event:
+
+  To ensure that the event is not lost even if the collector process is terminated or crashes, the collector may need to persist the event before acknowledging receipt to the workload or SDK. If the event cannot be persisted, receipt must be rejected.
```
reyang, line 97: What is the expectation if the collector instance disappeared (e.g., the machine running the collector exploded / was stolen)?

projects/audit-logging.md
```diff
@@ -0,0 +1,196 @@
...
+
+- OTel collector exports the event:
+
+  Once the event is exported and the target (i.e. S3) acknowledges receipt, the event can dropped from the collector's persistence.
+
+- S3 receives the event:
```
reyang, line 103: S3 -> the target (i.e. S3)

projects/audit-logging.md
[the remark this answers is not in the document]
renewelches: another examples from the security world is https://github.com/ocsf.

renewelches:
> > > Are there any requirements around signing logs / detecting tampering? I've heard that mentioned before in the context of audit logs, but I don't know how common of a requirement it is
> > 
> > 
> > @mtwo for all I know, immutability of audit logs is a common requirement although not all audit logging systems/use cases that I've seen address this requirement with technical measures but sometimes also organizational measures. However, given the flexibility of OTel processing queues (i.e. different topologies of collectors), having a technical solution in OTel would be favorable.
> > @reyang what is your opinion on this?
> 
> @mlenkeit I think this can be achieved as long as OpenTelemetry is designed to allow additive changes, doesn't have to be there in the first place. I personally haven't seen people signing logs, and I've seen lots of cases where immutable data path is used.

I think the key parts are encryption at REST e.g. when a buffer writes to disk and encryption at TRANSPORT. 
Where it is getting tricky is when we have to separate  audit log data and meta/transport data. In some cases this might lead to duplication. E.g the K8s cluster name that triggered the audit event could be part of the audit log message. In this case it must be immutable. But it is also an OTLP attribute and as such could be changed. Or to phrase it differently, is immutability required for the whole signal or just the message?

projects/audit-logging.md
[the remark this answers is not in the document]
svrnm: thanks @mlenkeit. Makes it much clearer

The `metadata` looks like attributes that would be covered by other semantic conventions (e.g. there is a [`log.record.id`](https://opentelemetry.io/docs/specs/semconv/attributes-registry/log/#log-record-attributes) for the `metadata.id`, the timestamp of course and some of the other ones (e.g. for `k8sCluster` we have [`k8s.cluster.name`](https://opentelemetry.io/docs/specs/semconv/attributes-registry/k8s/#kubernetes-attributes). So I would assume here it is more about re-using and extending certain other domains that are not unique to "audit logs"

For the `event` and `data` examples you gave, I would argue that they are not "semantic conventions for audit logs" but "semantic conventions for log types that typically require the strict requirements of auditing". What do I mean by that: if we talk about "semantic conventions for audit logs" I think about a namespace called `audit.` that holds attributes that are specific to the business logic of audit logging, like a signature that helps to tamperproof the log line, or maybe even meta information under which regulation this log is required to be an "audit log"
In contrast "semantic conventions for log types that typically require the strict requirements of auditing" are their own namespaces like the "UserLoginFailure" example would fall into a "authentication" or "auth" namespace, with "auth.login.method" or "auth.login.failureReason" as potential attributes, `event.name` being set to `auth.login.failure` or something.

I am just making those things up to exemplify the difference, they will probably take a different form or shape eventually, so to make a long story short, here is a suggestion to rephrase:

```suggestion
- no semantic conventions for audit logs in OTel
- no semantic conventions for log types that typically rquire the strict requirements of auditing, like authentication, authorization and data changes
```

@renewelches thanks for calling out OCSF, if I remember correctly there were conversations in the past between OTel and OCSF, cc @lmolkova

projects/audit-logging.md
svrnm: I think this is the most tricky part, or to put it in a question: do we need guarantee of delivery between 2 components (workload->collector,collector->S3) or end-to-end (workload->S3)?

I would assume "end-to-end" except the collector can guarantee that data is persisted according the auditing requirements

projects/audit-logging.md
mlenkeit: @reyang This is indeed supposed to refer to the OpenTelemetry Collector. I've rephrased this particular sentence to  "OTel Collector instances may lose audit logs in transit" and adjusted other occurrences of _OTel collectors_ as well. https://github.com/open-telemetry/community/pull/2409/commits/066501b92566ce770eb79f5633d2c1e5166aa9a5

projects/audit-logging.md
mlenkeit: good point, makes it more consistent. https://github.com/open-telemetry/community/pull/2409/commits/066501b92566ce770eb79f5633d2c1e5166aa9a5

projects/audit-logging.md
```diff
@@ -0,0 +1,196 @@
...
+
+  Note that this is outside the scope of the OTel. More general, when using OTel for audit logging purposes, it's the users (e.g. Ops) responsibility to configure a suitable export target.
+
+Note that this example may contain implementation details for illustration purposes. The actual implementation may differ as long as the requirements are met.
+
+The example is kept simple for illustration purposes. Many edge cases need to be discussed by the SIG, such as batch-sending of signals or handling of multiple export targets.
+
+It may turn out that all OTel receivers, processors, or exporters can be made compatible with guarantee of delivery for audit logging purposes.
+
+### Appendix B: Sample Audit Log Events
```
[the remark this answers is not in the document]
mlenkeit: prime example of how words can lead to completely different expectations depending on one's background 😄  I've changed it to "Examples of Audit Log Events". https://github.com/open-telemetry/community/pull/2409/commits/066501b92566ce770eb79f5633d2c1e5166aa9a5

projects/audit-logging.md
mlenkeit: > (e.g., the machine running the collector exploded / was stolen)
_If_ the solution for audit logging with OTel meant that the OTel Collector had an own persistence, I would argue that theft/explosion/etc. are rather in the responsibility of Operations in terms of configuring said persistence such that it is resilient "enough".
Or to make this more concrete: _if_ for example something such as the [storage extension](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/storage) was used, Operations would need to make sure that the database/file/redis storage runs in an HA mode.
I'm stressing the _if_ here, because I think is a detail that the SIG should work out. Or do you think that's something that should rather be clarified upfront?

projects/audit-logging.md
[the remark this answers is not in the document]
mlenkeit: Regarding `metadata`, I fully agree: Most of these attributes are already covered by semconv. We may identify additional attributes in SIG meetings though, depending on the experience/requirement of other contributors/companies.

I understand how "semantic conventions for audit logs" can be misleading. To me, the suggestion that you made has a notion of particularly describe logs that are "already there" (e.g. events emitted by a K8s cluster) and can be considered relevant for audit purposes. Especially in enterprise software, it's common that applications produce logs that are specifically mean to be audit logs (and nothing else). To me, it' s important that we find wording that covers these two types that we do have.

How about the following?

```suggestion
- no semantic conventions for representing and identifying audit trail-relevant events in OTel (like authentication, authorization or modification of

projects/audit-logging.md
reyang: > I'm stressing the _if_ here, because I think is a detail that the SIG should work out. Or do you think that's something that should rather be clarified upfront?

I suggest that we leave this for the SIG to figure out. In the OTEP, I suggest that we avoid "guaranteed delivery" and use something like "certain degree/level of data delivery guarantee". Not a blocker for this PR though (I'm good with the current version).

svrnm: +1 for what @reyang wrote. I think it is good to have this in the appendix and some wording around it, since there is many people (including myself) who have a superficial knowledge around audit logs, so it helps to contextualize and understand what this is all about

projects/audit-logging.md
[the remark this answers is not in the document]
tigrannajaryan: +1 to looking into OCSF for security events and borrowing relevant semantic conventions from there.

svrnm:
any updates / progress on this?

mlenkeit:
> any updates / progress on this?

@svrnm we presented the proposal in the spec SIG on Dec 3, 2024 and sparked a lively discussion. Together with @reyang, we decided to collect some more early feedback from different SIGs especially on the aspect of delivery guarantees. Over the next three weeks, I'm presenting the topic in a few language-specific SIGs and in the Collector SIG. I'll share updates here afterwards.

projects/audit-logging.md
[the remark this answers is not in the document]
trisch-me: @mlenkeit security semconv SIG is also working on introducing security related attributes to the semconv, we have currently discussion around [user authentication](https://github.com/open-telemetry/semantic-conventions/pull/1456). I feel that this SIG and security SIG might have some common tasks/discussions

mlenkeit:
Status update:
1. we've presented the proposal in the Collector, Java and JS SIG to collect more general feedback; this was rather positive
1. we are in the process of filling the staffing gaps in the SIG proposal
    - we're trying to fill the open engineering positions from the interested vendors
    - for maintainers/approvers, we'll reach out to community members directly starting with those who engaged in the discussions from 1)

mlenkeit:
We're proposing @hilmarf as the project lead; SIG proposal has been updated accordingly. Still working on getting names for additional engineers and maintainers/approvers.

mlenkeit:
Following the discussions at KubeCon London, we've reconsidered our approach and are now following a phased approach:
- In Phase 1 (_in progress_), we are **building an end-to-end prototype** to refine the challenges and requirements for audit logging in OTel and to showcase potential solutions. This is time-boxed until end of September 2025. We are set up to run this without a formal OTel project sign-off. We consider this truly as a proof of concept, i.e. we don't expect that the OTel modifications from the PoC will be accepted as-is and we are prepared to discard them if necessary.
- In Phase 2, we intend to **contribute functional extensions upstream** back to OTel. We will work towards signing off this project proposal and either join existing SIGs or form a separate one. The results from Phase 1 should help us in the discussions with the maintainers to make our proposed OTel extensions/changes more tangible.
- In Phase 3, we plan to **work on semantic conventions** for audit logging.

We have just started Phase 1 with @hilmarf and additional contributors from our side. While there isn't much yet, all our prototype efforts will be available at [apeirora/audit-log-poc-for-otel](https://github.com/apeirora/audit-log-poc-for-otel).

Towards the end of Phase 1, we'll reach out to the respective SIGs to demonstrate the results of the prototype and get additional support for Phase 2 and this project proposal.

We'll update the project proposal with more specific deliverables for Phase 2 as we gain more insights during Phase 1.

paulmbw:
For those interested, I'm building a new startup in this space, specifically LLM observability for audit and compliance. Read more here: https://traceprompt-web.pages.dev/

svrnm:
@mlenkeit can you provide an update on this proposal, if you'd like to proceed with it or where thing stand on your end?

hilmarf:
Hi @svrnm,

Yes — we would like to proceed.

We created a small proof-of-concept repository: https://github.com/apeirora/audit-log-poc-for-otel containing tests and initial findings.

Status and focus
- @mlenkeit added an overview of possible deployment topologies (see [PR](https://github.com/apeirora/audit-log-poc-for-otel/pull/60)/[asset](https://raw.githubusercontent.com/apeirora/audit-log-poc-for-otel/11874892462480e0c4271a73e202411058b5fe58/docs/assets/otel-deployment-topologies.drawio.svg)). We plan to focus on Type 6 and Type 7a because they support persistence and help avoid lost logs.
- Client-side proposals:
  - A named `ContextKey` to keep optional error callbacks traceable ([PR](https://github.com/apeirora/opentelemetry-java/pull/3)).
  - A `LogRecordProcessor` that uses an `AuditLogStore` interface for local persistence until the collector confirms receipt ([PR](https://github.com/apeirora/opentelemetry-java/pull/2)).
- Receiver-side work:
  - PRs ( https://github.com/apeirora/opentelemetry-collector-contrib/pull/2 and https://github.com/apeirora/opentelemetry-collector/pull/3 ) address persistence on the receiver, which is responsible for confirming delivery to clients.

We’re still iterating and would like to kick-start a SIG. Contributions, feedback, and reviews are very welcome — please comment on the linked PRs or open issues/PRs in the repository.

Thanks!

svrnm:
Thanks for sharing, @hilmarf, I'll need some time to take a closer look!

svrnm:
Hi @mlenkeit, @hilmarf, 

at the last two meetings of the Governance Commitee meetings ([recording 1](https://zoom.us/rec/share/OiVyi70Hf9yK_ySmEf2N8MgsUQ0HN-UWYIZAjJkH7OUW4JoamUHq-TGabfMZpI_-.jhZ7xAu3pauas1uR), [recording2](https://zoom.us/rec/share/wzEkX0Ou4UCdUmshekq100izBebty3B0sZYsxW-gdHUrr-5CjH0XnfaDfoH3KHYm._ih-ocAE2J9DQQF8)) we talked about open project proposals and how they align with the current goals and priorities of the project. And while "audit logging" is in scope of our project, we are currently not able to accept this proposal&formation of a new SIG, which requires resources to be assigned from our end as well. We are currently trying to focus on wrapping up and stabilizing current commitments before accepting new projects, we work on a blog post related to that, which draft you can read here: https://github.com/open-telemetry/opentelemetry.io/pull/8208

From our point of view the best way forward is, that you run this project stand alone, in close contact with us, and when things have evolved, come back joining the efforts.

We hope you understand where we are coming from and apologies for saying "no" to your valuable idea. We will keep this issue open for a little longer to answer any questions you might have.

Thanks,
Severin

hilmarf:
Hi @svrnm,

are you aware of another workstream which focuses more on the reliable delivery?

We've noticed the work around the new 'new exporter helper' and it seems that others are also interested in guaranteed delivery of logs (for auditing)/metrics (for billing).

See: https://github.com/open-telemetry/opentelemetry-collector/issues/8122#issuecomment-3474821727

Maybe we can join forces there?

Thanks,
Hilmar

svrnm:
First of all, apologies for the delay, I was on PTO and then on KubeCon last week.

> are you aware of another workstream which focuses more on the reliable delivery?

Not to my personal knowledge, but if there is some work being done in the collector SIG, @open-telemetry/collector-approvers would know best.

Not creating the SIG right now, does not mean that there is not an option to accomplish certain goals and requirements within the project that satisfy (some) of your needs. If there are workstreams in the collector that support audit logging requirements, then yes, joining forces is a great starting point!

mx-psi:
@hilmarf Speaking as a member of the Collector SIG, we would be happy to join forces with you on this front, specially now . I see you upvoted https://github.com/open-telemetry/opentelemetry-collector/issues/8122#issuecomment-3474821727, it would be interesting to discuss more about how you think we can adress this.

I am happy to chat more via DM and we can prepare something that you can present to the wider Collector SIG in one of our SIG meetings.

svrnm:
@hilmarf @mlenkeit did you have a chance to chat with @mx-psi and the collector SIG? I think getting the audit work started there is a great way of making some traction, as from what I understand a lot of the work that is required needs to be done in the collector.

svrnm:
@hilmarf @mlenkeit have you had a chance to look at @mx-psi's comment? Thanks!

darasok85az-lang (changes requested):
Hello can you guys like help me figure something out I don't know how to work this platform if any sort of funds or coming into this account can you guys show me how to link an account onto here because I have never linked an account into here somebody had hacked into my old account so I don't know if they're link in their account to it so they're the money goes into their account or what but I need some help

hilmarf:
Component donation: signingprocessor - https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/50079

MarkovianProtocol:
We published a pinned canonical form and hash for OTel spans with conformance vectors: https://github.com/MarkovianProtocol/otel-canonical-profile at 4f8c4c5. An outside implementer ran both validators and got 18/18 in Python and Node. Spans rather than log records, so not a drop-in — I can produce the equivalent vectors for the fields the processor canonicalizes if that's useful.

Levaj2000:
A datapoint for the open scope question ("immutability, tamper-proof logs, signing — in scope?"): OCSF 1.9.0, released last week, ships the same primitive that this proposal's draft data model and the `signingprocessor` donation (open-telemetry/opentelemetry-collector-contrib#50079) describe — a signature over a canonical form of the record, plus a hash link to the previous record. Disclosure: I contributed the OCSF side (the `attestation` object and `record_integrity` profile).

Many pipelines will end up touching both — OTel as transport, OCSF as the envelope in the SIEM — so here is the field mapping while both vocabularies are still cheap to align:

| Audit Logging draft | OCSF 1.9.0 |
|---|---|
| canonical form: RFC 8785 (JCS) | `digital_signature.serialization_id` (enum: JCS, JWS, COSE, DSSE) |
| `audit.integrity.algorithm` | `digital_signature.algorithm_id` / `algorithm` |
| `audit.integrity.certificate` | `digital_signature.certificate` |
| `audit.integrity.value` (signature bytes) | no landed equivalent yet — proposed in ocsf/ocsf-schema#1709 |
| `audit.sequence.prev_hash` | `attestation.prev_event.fingerprint` |
| `audit.sequence.number` | no direct equivalent (`prev_event.uid` links by id, not counter) |
| — | `attestation.chain_uid`, `attestation.authority_uid` — no draft equivalent |

Two observations from having shipped the OCSF shape, offered as input rather than objections:

1. `sequence.number` + `prev_hash` give continuity within a stream, but nothing names the stream or says who vouches for it. We added `chain_uid`/`authority_uid` after hitting exactly this in multi-tenant verification: records signed under different keys interleave in one pipeline, and the verifier needs to know which chain a record claims membership in before it can check the link.
2. Worth deciding early where the signature is produced. A Collector-side processor attests custody from the pipeline onward; producer-side signing attests from the point of action. Both are legitimate tiers — but the record should say which one a consumer is getting.

Happy to contribute the mapping as a doc, or conformance vectors in either direction, if the SIG forms and wants them.

hilmarf:
> Two observations from having shipped the OCSF shape...

Thanks a lot @Levaj2000 for your inputs and experience sharing with OCSF. I've tried to incorporate them with https://github.com/apeirora/opentelemetry-specification/pull/6/changes/578b930a8958612bdb5795f52ffde54e9efc6576. Happy to get your feedback.

Given your hands-on experience with OCSF, it would be great to have you as a supporter of this SIG — your perspective on the intersection of OCSF and OTel would be really valuable as we shape the audit logging signal. Would you be open to joining us?

Levaj2000:
Thanks @hilmarf — happy to join as a supporter; please add me. The corner I can usefully hold is the OCSF↔OTel intersection: we run the OCSF side of this in production (hash-chained, per-event-signed audit exports under the record_integrity profile that landed in OCSF 1.9 via [ocsf/ocsf-schema#1661](https://github.com/ocsf/ocsf-schema/issues/1661)), and my interest is that the two specs end up mutually verifiable rather than parallel.

I read through 578b930 — both additions are the right calls. audit.sequence.stream_id resolves the demultiplexing problem cleanly, and the audit.integritysigner language that producer and collector signatures "MUST NOT be treated interchangeably" is exactly the discipline this needs: point-of-action attestation and custody attestation answer different questions, and we learned to keep them separate the slow way.

A few notes from having shipped this shape, in roughly descending order of how much round-trip they might save you:

A chain pointer needs a resolvable record identifier, not only a hash. A predecessor hash proves linkage but can't locate the predecessor across storage, sharding, or retention boundaries. OCSF's merged shape ended up requiring every event to carry metadata.uid and the previous-record pointer to reference it (prev_event.uid), with the hash binding the reference to content. I'd recommend the data model require a per-record id within the stream and have the chain pointer carry both id and hash.

Define genesis explicitly: the first record of a stream omits the previous-record pointer entirely — never a sentinel. We shipped the sentinel version once (a literal "GENESIS" string inside a hash-valued field) and have it documented as an anti-pattern: a hash field whose value isn't a hash breaks every schema-driven verifier. One sentence of spec text here saves every implementer that round trip.

Signatures and chain hashes are only third-party-verifiable if the record declares what bytes were hashed. I'd recommend a required companion attribute naming the canonicalization for audit.integrity.value — RFC 8785/JCS, flat bytes, or a named producer scheme. OCSF landed on an enum plus a free-text sibling precisely so producers whose serialization isn't JCS can say so honestly instead of misclaiming; without it, "verify the signature" quietly becomes "trust the producer."

Offer: a written OCSF↔OTel crosswalk. The mappings look near-lossless from here — audit.sequence.stream_id ↔ attestation.chain_uid, the producer/collector signer split ↔ producer signatures vs custody attestations with authority_uid naming the attesting party, the chain pointer ↔ prev_event. If we keep that mapping tight, one producer can emit both shapes (or transform between them) without divergent integrity constructs. Happy to draft it as a doc for the spec repo if useful.

We can also contribute test vectors for the verification-procedures section: a production OCSF export (236-event chain, per-event ECDSA-P256 signatures verifiable against a public JWKS, no secrets required) and a stdlib-only worked sample with fully recomputable fingerprints — both public:

https://github.com/levaj2000/ai-identity/tree/main/docs/cosai-ws4-ocsf-mapping/ocsf-log-reference-bundle
https://github.com/levaj2000/ai-identity/tree/main/docs/cosai-ws4-ocsf-mapping/trust-base-inventory-sample

Looking forward to it. Cheers

atoulme:
> Offer: a written OCSF↔OTel crosswalk. The mappings look near-lossless from here — audit.sequence.stream_id ↔ attestation.chain_uid, the producer/collector signer split ↔ producer signatures vs custody attestations with authority_uid naming the attesting party, the chain pointer ↔ prev_event. If we keep that mapping tight, one producer can emit both shapes (or transform between them) without divergent integrity constructs. Happy to draft it as a doc for the spec repo if useful.

Interested in this. I dabbled into mapping logs to OCSF events here: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47461

surpradhan:
I maintain [otel-agent-audit](https://github.com/surpradhan/otel-agent-audit), a signed-audit Collector component for AI agent traces, listed in the OTel Registry: canonical serialization, hash chaining, Ed25519 signing, and a standalone verifier. Most of the draft data model is something I've implemented and had to operate, so I'd rather contribute that experience than duplicate the work.

Three findings that bear on the draft.

**Arrival order can't be the chain order.** Collector batching and concurrency mean arrival order isn't reproducible. I order by a content-derived key before hashing, so a verifier can re-derive the same chain from the same records. Anything chaining on arrival sequence is reproducible only for the process that wrote it.

**Sealing needs an explicit completeness boundary.** I chain per trace, and a trace seals when its root span arrives. If the root arrives before its children, those children are dropped: the chain is internally valid but incomplete. I needed a separate buffering processor upstream to hold this together. Discrete events may not hit this, but anything with a parent/child or session boundary will.

**Rotation is the gap I'd most like the spec to close.** With `audit.sequence.prev_hash` and now `stream_id`, a chain crossing a key rotation is linked structurally, but that link is signed by the incoming key. Nothing signed by the outgoing key attests to the handover, so a legitimate rotation and a truncation with a rotation-shaped hole are hard to tell apart. I filed it against my own implementation [here](https://github.com/surpradhan/otel-agent-audit/issues/19) and don't have a good answer yet. It seems worth settling at spec level rather than per implementation.

Two things I could contribute if useful: signed checkpoints over chain tips, for cross-stream completeness rather than only per-record continuity; and verifier tooling, since for an audit format the verifier is arguably the product and it's the thinner half right now.

@hilmarf @svrnm I'd like to join as a supporter if you're still taking them. The corner I can usefully hold is AI agent traces, including the bridge to the GenAI SIG's in-flight guardrail work (open-telemetry/semantic-conventions-genai#262, open-telemetry/semantic-conventions-genai#427), where a policy decision is exactly the kind of event that should land in a verifiable record.

Levaj2000:
@atoulme 

 Great — the draft is already underway: [OCSF↔OTel audit crosswalk + derived test vectors](https://github.com/levaj2000/ai-identity/tree/main/docs/otel-ocsf-audit-crosswalk). Field-level mapping between the audit data model draft and OCSF 1.9's record_integrity shape, the places it isn't 1:1 with what a transform must do there, and OTel-shaped vectors derived from our production OCSF export — chain linkage re-verifies across the transform (235/235 links on the 236-event export), and the signatures survive byte-for-byte, verifiable against the same public JWKS.

The connector angle is exactly why it's worth writing down: with the mapping explicit, [collector-contrib#47461](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47461)-style translation becomes a mechanical transform instead of a design exercise, and the integrity constructs survive rather than being rebuilt divergently on each side. One finding worth flagging for a connector specifically: a signature is bound to the canonical bytes of its origin shape — a transform can carry the proof (origin digest + declared canonicalization) but can never re-sign, which is what "lossless" can and can't mean here.

Would value your review against what the connector actually needed. @hilmarf — the doc ends with four one-attribute spec asks that would make the mapping fully lossless (three OTel-side, one already filed on the OCSF side); happy to bring it into the spec repo in whatever form suits.

Cheers, Jeff

hilmarf:
@surpradhan Thanks a lot for your inputs!

I've added the rotation and sealing problems as open questions in the OTEP. https://github.com/apeirora/opentelemetry-specification/pull/6/changes/165dd2ba1e6eccea37ffd1f137e33cfbeea9e6d8

Regarding arrival order: I think `stream_id` (relatively new - https://github.com/apeirora/opentelemetry-specification/commit/578b930a8958612bdb5795f52ffde54e9efc6576) + `sequence.number` should be sufficient to reconstruct the chain deterministically, regardless of arrival order. What do you think?

When multiple SDK instances fan-in (different stream_ids) - there's no global ordering across streams, true. Is this something you ran into?

Unfortunately this SIG hasn't started yet, due to other higher priorities from OTel. The more I'm happy to have you onboard with your valuable experience.

surpradhan:
Thanks @hilmarf, and good to see the sealing and rotation points land as open questions in the OTEP.

**stream_id and sequence.number are enough to reconstruct the chain for source-emitted streams.** When a single AuditLogger assigns the sequence and owns the stream_id, arrival order at the Collector doesn't matter: a verifier demuxes by stream_id, sorts by sequence.number, and checks prev_hash. The caveat I raised only applies when records are derived downstream from telemetry that arrived out of order, which is my case: building records from agent spans in the Collector. There's no source-assigned counter there, so I order by a content-derived key (start_time_unix_nano, span_id) before hashing, and a verifier re-derives the same chain from the same records. For that derived case, the order has to come from record content rather than arrival order, or two honest rebuilds can disagree. That's probably its own profile rather than something the base signal needs, since a source-emitted stream never encounters it.

**A dropped stream leaves no trace in the streams that survive.** Within a stream, sequence.number catches gaps, but a whole stream that's lost or truncated leaves every other stream intact, so per-record continuity can't detect it. What I've leaned on is periodic signed checkpoints that form their own hash chain: a stream's tip {stream_id, tip hash, entry count} is committed when it seals, batched every N seals, with each checkpoint linking to the previous one. A verifier replays that chain and re-derives every committed tip from the log, so a stream that reached a checkpoint can't be dropped or truncated without causing a tip mismatch or a break in the checkpoint chain. Honest bounds: a stream dropped before its checkpoint is written, or one that never seals, isn't covered. So this narrows the completeness window to the checkpoint interval and also serves as a batched seal record, complementary to the terminal-record idea rather than replacing it. 

**Only a co-signed transition record makes the outgoing key attest to a rotation.** Of the three options, that's the one that actually closes the gap. A per-record key_id and a JWKS pointer let a verifier map records to keys, but neither has the outgoing key sign the handover, so a legitimate rotation still looks like a rotation-shaped truncation. I don't have this solved in my own implementation yet, but that's the direction I'd back, and I'm happy to prototype it against a live chain and report what breaks.

hilmarf:
> ... the doc ends with four one-attribute spec asks that would make the mapping fully lossless (three OTel-side, one already filed on the OCSF side)

Hi @Levaj2000, I hope this (https://github.com/apeirora/opentelemetry-specification/pull/6/commits/7cac2e1dcbaa028bf7941308413df0b7035972a8) does address the three open points for OTel. Happy to hear your thoughts.

MarkovianProtocol:
Rotation and the checkpoint idea are the same gap. `sequence.number`, `prev_hash` and the signature all come from one key, so a producer can keep two sets of books under one `stream_id` — one chain served to you, another to whoever checks later — and every record in both verifies. What separates them isn't in the record. It's a second observer.

So the checkpoint signature has to come from outside the producer. Cosigners that remember the last tip they saw for a stream, and refuse a new one inconsistent with it, make the second book uncollectable — one fork can't get signed. Rotation falls out of the same move: a cosigner holding the pre-rotation tip won't sign across the boundary unless the outgoing key signed the handover.

It attests the stream is singular and append-only, not that any record is true, and covers nothing before a stream's first checkpoint. Running live: `curl https://log.markovianprotocol.com/checkpoint` — RFC 6962, tree size 7897, eight independent witnesses.

Happy to write it against the OTEP's `audit.sequence` attributes, with vectors.

Levaj2000:
Thanks @hilmarf — read through 7cac2e1 and the follow-ups on the auditing branch. All three asks land:

1. `audit.integrity.canonicalization` — jcs as the assumed default plus the explicit-set recommendation is the right split. Verifiers no longer guess, and the section text names the failure modes honestly.
2. The resolvable chain pointer — and since `audit.record.id` is MUST, the locator half rests on a required field, which is exactly what a transform across storage, sharding, and retention boundaries needs.
3. Genesis by omission — this came out stronger than my ask: MUST omit, and receivers MUST NOT require a magic constant. Good.

One note for the thread: I reviewed at the branch tip, where these now read `previous_hash` / `previous_record_id` after the clarity renames — flagging in case anyone lands on this comment from 7cac2e1's diff and wonders about the names.

Also noticed `audit.sequence.end` landing today — that's a graceful answer to the sealing question @surpradhan raised: a defined completeness boundary for clean shutdowns, with crash truncation honestly "unknown" rather than silently passing. I'll check whether the crosswalk needs an end-signal row and update the doc.

One more observation on the open rotation question below: the thread is converging usefully — @surpradhan's outgoing-key-signed handover plus @MarkovianProtocol's framing of rotation and checkpoints as one fork-resistance gap, with the checkpoint cosigner sitting *outside* the producer. That last part is the load-bearing move: one producer's keys can sign both books. Happy to review the OTEP text when it lands.

Cheers, Jeff

MarkovianProtocol:
Two defects, with vectors: https://github.com/MarkovianProtocol/otel-audit-vectors — `python3 verify.py`, stdlib only, offline.

`audit.sequence.previous_hash` is defined twice. The table says SHA-256 *of* the predecessor's `IntegrityHash`; the prose says it *is* that hash. Case A passes under one reading, fails under the other, and the failure surfaces as a critical integrity violation. Two implementers reject each other's chains.

The second one has both halves working as written. `IntegrityHash` covers the record as persisted, and `audit.integrity.value.1` is in that record — so a collector countersignature changes every hash, and every producer-written `previous_hash` stops matching. Case C: signatures PASS, `prev_hash` FAIL on the whole stream. Chain over the form that already excludes `audit.integrity.*` and it goes away.

Case B is the fork from my earlier comment, runnable: same `stream_id`, divergent tail, every per-record check passes.

surpradhan:
@MarkovianProtocol, you're right about my checkpoints. The same Ed25519 key signs the entries and the checkpoints, so a producer holding it can serve two consistent chains and both verify. My [threat model §2](https://github.com/surpradhan/otel-agent-audit/blob/main/docs/threat-model.md) already lists split-view as undefended with witnesses as the upgrade path, so it isn't news on my side, but better said here than letting the checkpoint idea sound stronger than it is. The narrower thing they do buy, and only against an adversary who doesn't hold the key, is detecting a whole stream going missing, which per-record continuity can't, since a dropped stream leaves every surviving stream intact. Singularity needs the outside observer.

Agreed too that a cosigner won't sign across a rotation unless the outgoing key signed the handover. The part I'd add is that the dependency only runs one way. The attacker in [my rotation issue](https://github.com/surpradhan/otel-agent-audit/issues/19) holds the incoming key and not the outgoing one, and cuts across a boundary so it looks like a routine rotation. A co-signed transition record stops that on its own, with no witness anywhere in the picture. So option (a) under open question 8 can be settled now on its own merits, rather than waiting on whatever checkpoint and witness design lands later, and it makes that design easier when it arrives.

On your span profile: it pins a different canonical form from mine by design, so running its vectors against my form as-is would just disagree on every byte. What's actually useful is implementing your profile alongside my own serialization and checking I reproduce your hashes and your reject cases, then writing up where the two forms deliberately differ. Your choice to string-encode nanosecond timestamps is the same 2^53 boundary I've just filed against the `signingprocessor` as open-telemetry/opentelemetry-collector-contrib#50881, where nanosecond timestamps and int64 attributes are currently bare JSON numbers going through JCS. Happy to do that and report back.

MarkovianProtocol:
@surpradhan

You're right about the direction of the dependency, and it changes the ordering.
The attacker in your issue 19 holds the incoming key and not the outgoing one, so
a transition record signed by the outgoing key refuses that case on its own, with
no witness anywhere in it. Option (a) under open question 8 stands on its own
merits and does not need to wait on a checkpoint design.

Your narrowing of what checkpoints buy is the accurate one. Against an adversary
who does not hold the signing key, a checkpoint over `{stream_id, count, tip}`
detects a whole stream going missing, which per-record continuity cannot, because
a dropped stream leaves every surviving stream intact. Non-equivocation is the
part that needs the second observer.

On running our profile alongside your serialization: spec, pass and reject
vectors, and two dependency-free validators (Python stdlib, Node) are at
https://github.com/MarkovianProtocol/otel-canonical-profile. The places the two
forms will disagree are deliberate, so they are worth naming before you spend
time deriving them:

- `start_time_unix_nano` and `end_time_unix_nano` are decimal strings, and a JSON
  number there is a reject (`TIMESTAMP_NOT_STRING`) rather than a coercion. Same
  2^53 boundary you filed as collector-contrib#50881.
- int64 attributes are decimal strings inside a typed wrapper, `{"i": "-12"}`.
  Every attribute value carries a one-key type tag, so `{"i":"1"}` and `{"s":"1"}`
  are different bytes instead of relying on JSON's own types to separate them.
- Attributes are folded in behind a second-level digest, `attributes_sha256`, so
  a verifier whose store rewrote attribute bytes can still check the span core.
  It also means our two forms can disagree on attributes without disagreeing on
  the core.
- Span fields are an allowlist; `status.message` is outside hash scope.
- Duplicate attribute keys, NaN and +/-Infinity doubles, and uppercase or
  wrong-length ids are rejects, not normalizations.

One practical note so it doesn't read as a hole in the vector set:
`NONFINITE_DOUBLE` has no JSON vector, because NaN and Infinity are not
representable in JSON transport. It is reachable only through protobuf ingestion,
and the spec says implementations must still enforce it there.

Worth having the write-up cover the cases where our form is the one that is
wrong, not only where they differ.

*Drafted with AI assistance.*

PR #2423, opened 2024-11-06
Remove actuated docs

trask:
See https://github.com/open-telemetry/community/issues/2135#issuecomment-2438011230

(also fixes CI broken link on https://gist.github.com/alexellis/1f33e581c75e11e161fe613c46180771)

PR #2424, opened 2024-11-06
Fix google group used for zoom bombing doc

trask:
It looks like I updated the permissions on the doc already, but forgot to update it here.


arminru (approved):

danielgblanco (approved):

trask:
> hmm... it looks like the gist in https://github.com/open-telemetry/community/blob/main/docs/using-actuated.md#right-sizing-vms was removed by the author.

sent fix in #2423

PR #2437, opened 2024-11-13
GenAI SIG: Update alternate meeting to 16:00 PST

drewby:
We are moving the alternating meeting to 16:00 PST to get better attendance. The current time is not getting a quorum to make decisions.

@lmolkova @tedsuo @nirga


arminru (approved):

codefromthecrypt (commented):
While I'm personally ok with this, I feel it helpful to recap some points last discussion about timezones which led to this time in SIGs such as this.

1. folks with family duties in UTC+8 may struggle with this time
2. Having not in UTC+8 will likely lead to folks in the timezone missing due to misunderstanding US timezone impact
3. This makes places like india quite late or early

drewby:
> While I'm personally ok with this, I feel it helpful to recap some points last discussion about timezones which led to this time in SIGs such as this.
> 
> 1. folks with family duties in UTC+8 may struggle with this time
> 2. Having not in UTC+8 will likely lead to folks in the timezone missing due to misunderstanding US timezone impact
> 3. This makes places like india quite late or early

Yes, this was much the motivation to move the meeting the first time, but we did not get the attendance. 

I'm OK to change to UTC time but wanted to be consistent with what was already there. Should we change both?

codefromthecrypt:
> Yes, this was much the motivation to move the meeting the first time, but we did not get the attendance.

One suggestion I had when this first started was to give more time to allow things to settle before canceling again (some SIG cancel then resume). I hinted to someone I forget, at least 6 months. Personally, I wouldn't propose this change or 👍 it, at best +0. We are raising awareness of a new group composure, and a new product. This has a good chance of drawing in more people, and ratcheting back the timezones will add barriers to entry. The other thing, is while I do agree we aren't there, yet, we do want a group to eventually be more diverse globally. When all meetings have to be US timezone friendly, it counters that point. These are the same reasons for the original timezone layout.

> I'm OK to change to UTC time but wanted to be consistent with what was already there. Should we change both?

I think how it was before, one in pacific one in UTC+8 is fine as it was about the audiences involved. Mainly not anchoring UTC+8 to US timezone related things. If folks outside the US want the former normalized to UTC, that's ok with me.

lzchen (approved):

lmolkova:
Discussed at the GenAI SIG call:
- we'd like to have a weekly meeting (with a quorum) - currently de-facto we meet every other week. So adding another 4PM (or 1-2 hours later) PT slot would be great - we had better engagement there.  
- 4+ PM PT MAY work for some folks in APAC (8am JST, SGT, CST), but excludes India.  
- we should be less meeting-focused (capture discussions in the meeting notes and on issues/PRs)
- we can keep APAC slot open, but we kindly ask people who would want to join APAC call to add their topics to the agenda beforehand so that others know and can plan accordingly

@patcher9 and @AmanAgarwal041 we'd like to check with you folks - it seems this change would be the most problematic for you folks. I wonder what your thoughts are and if you're planning to attend current APAC call  (14:00 UTC).

I'll also post in the GenAI instrumentation channel.

codefromthecrypt:
I wonder if we are being too narrow with the impact of this, thinking mostly about india and openlit specifically. Doesn't this also affect israel and europe? While participation has indeed been sporadic, it seems we should consider if this closes the SIG to folks who we should be pulling in more often. Many folks I work with are in europe and I'm a little surprised no feedback on this. Thoughts?

lmolkova:
> Adding my 2 cents from Elastic: we are looking to increase our attendance at the SIG meeting and many of us are in Europe. So we would very much appreciate a meeting at a EU-friendly time. Thanks!

Given that we have plenty of active participants from Europe, I'd prefer to focus on extra slot in Europe-friendly time.

Proposed slot is 1am CET (4pm PDT) is impossible to attend in Europe. 

Our current slots might, however, work for some contributors there:
- 7pm CET (10am PDT)
- 8am CET (11pm PDT)

@estolfo do you think either of those can work?

If not, I wonder if we can find the slot that works for most people in Europe - other OTel projects use 4pm-7pm CET (7am-10am PDT) which mostly work for the US too, but there are a lot of conflicts for folks who participate in multiple otel projects. 4pm CET/7am PDT would probably be the least crowded.

codefromthecrypt:
The current time is 14:00 UTC+8 which is 7am berlin time, and this still allows folks in UTC+8 to attend. We need to be conscious of cutting folks like me out with proposals, at least by mentioning it.

Other SIGs are allowing different people to start calls, also, where it isn't always the case where one person can attend all meetings. For example, if we want another model, we can get feedback from @mx-psi and @jpkrohling who have been working in a different 3 tier model.

The scope of this SIG is very wide, and I doubt it will be sustainable over time to cut out regions and/or assume a single person can attend all meetings. My 2p albeit unsolicited.

codefromthecrypt:
Personally, I  mentioned that I had wasn't able to attend meetings for nearly a month (due to medical which also affected my normal job). I also mentioned that many I knew were tied up with kubecon related busy stuff which not only affected meetings but other sorts of sig participation.

Despite that, if we look at the logs, folks have been joining in all existing timezones. https://docs.google.com/document/d/1EKIeDgBGXQPGehUigIRLwAUpRGa7-1kXB736EaYuJ2M/edit?tab=t.0#heading=h.ylazl6464n0c

Maybe we can focus on what we feel is a failure at the moment. For example, what number of attendance warrants re-doing the schedule and how persistent that pattern should be in order to change things.

When we change it, it affects people. I saw chat negotiation with OpenLit who are in india, who were ok being unable to attend anymore due to a switch to earlier UTC+8. However, cutting UTC+8 out is a tricky move. It undermines a lot of work I've been doing in the community to increase interest in otel. In fact, I'm doing yet another presentation on this next week in singapore (last month it was kubeday melbourne).

TL'DR' this is very tricky, but please let's not undermine a very hard to engage timezone which has a huge potential and was the reason of the original schedule in the first place.

codefromthecrypt:
updated above comments as I misunderstood the last suggestion. I think IIUC was to create another timeslot (so 3), which is far better than choosing which region to black hold. I also added notes about what quorum means, and personally believe we can find folks who can step up into approver role who aren't clustered in one region.

jpkrohling:
In the Collector SIG, we've been having three slots for quite some time. The result is that most people are able to attend two thirds of the calls, meaning that we all know that SIG calls are more like "office hours" than a meeting where we decide things. Decisions are made via GitHub only, and what we discuss via SIG calls and Slack is summarized in GitHub as well.

While I recognize that our APAC-centric slot has lower attendance, the fact that we have prominent contributors there (hi @MovieStoreGuy!) means that good discussion can happen there even if other maintainers can't join (ie, no quorum).

danielgblanco (approved):

estolfo:
Hi @lmolkova is there any decision about this? Would 4pm CET/7am PDT be possible as one of the LLM SIG time slots?

Cirilla-zmh:
> Regardless of what we do, let's not cut out UTC+8 or make it hard for folks to join.

Our group are also working on Gen AI observability and are also very willing to contribute to the community. However, since I am in China, the time of 16:00 PST is indeed not very convenient for us. Regular online meetings at UTC+8 would be very meaningful for us.

If we cannot always find a time that suits everyone, why not keep two different times? Like what we are running for another SIG? ;)

Issue #1809, opened 2023-11-16
Many meeting notes still reference recordings being stored on YouTube


trask:
Need to go through meeting notes (https://github.com/open-telemetry/community#special-interest-groups) and update them to reference the new storage location: https://docs.google.com/spreadsheets/d/1SYKfjYhZdm2Wh2Cl6KVQalKg_m4NhTPZqq-8SzEVO6s

mtwo:
This was completed last year, closing

PR #2449, opened 2024-11-21
Add href anchors for each sig entry in README tables

trask:
As part of cleaning up meeting docs in https://github.com/open-telemetry/community/issues/1809, it would be nice to be able to link directly to each SIG entry in the community readme.


scripts/update-sig-tables.py
```diff
@@ -91,9 +96,9 @@ def format_chat(chat):
             calendar = f"[{invites}](https://groups.google.com/a/opentelemetry.io/g/{invites})"
         
         if group_name == "Specification SIGs":
-            markdown_content += f"| {name} | {meeting} | {notes} | {chats} | {calendar} | {tc_sponsors} | {gc_liaison} | \n"
+            markdown_content += f"| <a href=\"{short_name}\"></a>{name} | {meeting} | {notes} | {chats} | {calendar} | {tc_sponsors} | {gc_liaison} | \n"
```
svrnm, line 99:
```suggestion
            markdown_content += f"| <a id=\"{short_name}\"></a>{name} | {meeting} | {notes} | {chats} | {calendar} | {tc_sponsors} | {gc_liaison} | \n"
```

Based on [fragment navigation](https://html.spec.whatwg.org/multipage/browsing-the-web.html#scroll-to-fragid), set this to id or name.

scripts/update-sig-tables.py
```diff
@@ -91,9 +96,9 @@ def format_chat(chat):
...
         else:
-            markdown_content += f"| {name} | {meeting} | {notes} | {chats} | {calendar} | {gc_liaison} |\n"
+            markdown_content += f"| <a href=\"{short_name}\"></a>{name} | {meeting} | {notes} | {chats} | {calendar} | {gc_liaison} |\n"
```
svrnm, line 101:
```suggestion
            markdown_content += f"| <a id=\"{short_name}\"></a>{name} | {meeting} | {notes} | {chats} | {calendar} | {gc_liaison} |\n"
```

scripts/update-sig-tables.py
```diff
@@ -65,6 +65,11 @@ def format_chat(chat):
             if chat.get('name') and chat.get('type')]
         )
 
+        short_name = None
+        for chat in sig.get('chat', []):
+            if chat.get('type') == 'slack':
+                short_name = chat.get('name').replace('#otel-', '')
```
svrnm, line 71:
```suggestion
                short_name = chat.get('name').replace('#', '')
```

I would keep a kind of prefix for this, otherwise if we ever have a section called "Maintainers" or "Collector", etc we create a conflict. My preference would be "sig" but since we have some of them already maybe stick with otel?

Or we "urlify" the name?

svrnm (changes requested):
I love this! I ran into this problem so often that I wanted to link to a SIG and didn't have an anchor for that, so this is great.

To put this out for discussion, although I see advantages in both solutions, we could also turn this table into a list of sections, so instead of each SIG having a row we could go 

```
### Specification: General

* Meeting Time: ...
* Meeting Notes: ... 
* Slack Channel: ...
``` 

That way we could even have a quick description above.

Although this would make the document muuuuuch longer.

scripts/update-sig-tables.py
trask: > My preference would be "sig"

makes sense to me, I added the `sig-` prefix (where it wasn't already prefixed by `sig-`

svrnm (approved):

mx-psi (approved):

PR #2492, opened 2024-12-16
[DO NOT REVIEW] Add Otel/STEF project

tigrannajaryan:
Otel/STEF (Sequential Tabular Encoding Format) is a new data format and network protocol for OpenTelemetry data.

For the target use-cases Otel/STEF outperforms both OTLP and Otel Arrow (phase 1): Otel/STEF is smaller and/or faster.

See stef.md for details: benchmarks and comparisons to other formats, links to prototypes and description of project goals.

### DIFF
--- a/projects/stef.md
+++ b/projects/stef.md
@@ -35,6 +35,8 @@ The project goals are the following:
   Arrow.
 - Refine and stabilize Otel/STEF specification and encourage its implementation
   inside or outside Otel.
+- Evaluate additional use-cases that can be served by Otel/STEF, for example
+  direct-export from SDKs to a local file.
 
 Project non-goals:
 
--- a/projects/stef.md
+++ b/projects/stef.md
@@ -38,7 +38,7 @@ The design principles are:
 Design non-goals are:
 
 - Random access to data by seeking to an arbitrary point in the stream.
-- Ability to keep the entire data batch in memory efficient and traverse it for 
+- Ability to keep the entire data batch in memory efficiently and traverse it for 
   processing purposes. Otel/STEF is not a replacement for Protobuf-based OTLP.
 - Ability to query (search) data by some criteria. Otel/STEF does not target Parquet's 
   niche.
@@ -54,6 +54,9 @@ The project goals are the following:
 - Add Otel/STEF exporter and Otel/STEF receiver to Otel Collector.
 - Publish benchmark-justified guidelines on applicability of Otel/STEF vs OTLP vs Otel
   Arrow.
+- Explore the limitations of Otel/STEF, the impact of its statefulness and 
+  persistent connection on load balancers and servers that accept a large number of 
+  connections.
 - Refine and stabilize Otel/STEF specification and encourage its implementation
   inside or outside Otel.
 - Evaluate additional use-cases that can be served by Otel/STEF, for example
--- a/projects/stef.md
+++ b/projects/stef.md
@@ -19,8 +19,29 @@ Otel/STEF supports options that allow trading speed for size. At the highest
 speed setting it can be faster than OTLP (_including_ conversion to/from pdata), while 
 producing significantly smaller payloads.
 
-A draft specification and a prototype implementation of Otel/STEF is attached to thi
+A draft specification and a prototype implementation of Otel/STEF is attached to this
 proposal.
+
+### Design Principles
+
+Otel/STEF's design is optimized for fast serialization and compact representation at the 
+expense of other usage patterns. It chooses speed+size over other capabilities 
+that formats may have.
+
+The design principles are:
+
+- Fast sequential record-by-record reading/writing.
+- Minimal allocations during serialization/deserialization.
+- Stateful codec, optimized for avoiding repetitive work and duplicate data when used 
+  as a network protocol.
+
+Design non-goals are:
+
+- Random access to data by seeking to an arbitrary point in the stream.
+- Ability to keep the entire data batch in memory efficient and traverse it for 
+  processing purposes. Otel/STEF is not a replacement for Protobuf-based OTLP.
+- Ability to query (search) data by some criteria. Otel/STEF does not target Parquet's 
+  niche.
 
 ### Goals, objectives, and requirements
 
@@ -69,13 +90,13 @@ The benchmarks are run on the highest compression / slowest speed option.
 ### Wire Size
 
 The following benchmarks show zstd-compressed wire sizes of OTLP vs Otel/STEF vs Otel 
-Arrow. Benchmarks are run using
+Arrow Streaming Mode. Benchmarks are run using
 [Otel Arrow bench](https://github.com/open-telemetry/otel-arrow), with Otel/STEF
 prototype implementation added for comparison.
 
 Dataset `hipstershop_metrics.pb`:
 
-| BATCH SIZE         | OTLP                    | Otel/STEF                       | OTEL ARROW                      |
+| BATCH SIZE         | OTLP                    | Otel/STEF                       | OTEL ARROW + Stream Mode        |
 |--------------------|-------------------------|---------------------------------|---------------------------------|
 | Compressed (bytes) |                         |                                 |                                 |
 | batch_size: 128    | 4579  (total: 1.6 MB)   | 947 (x  4.83) (total: 323 kB)   | 2067 (x  2.22) (total: 705 kB)  |
@@ -86,7 +107,7 @@ Dataset `hipstershop_metrics.pb`:
 
 Dataset `multivariate-metrics.pb`:
 
-| BATCH SIZE         | OTLP                     | Otel/STEF                      | OTEL ARROW                       |
+| BATCH SIZE         | OTLP                     | Otel/STEF                      | OTEL ARROW + Stream Mode         |
 |--------------------|--------------------------|--------------------------------|----------------------------------|
 | Compressed (bytes) |                          |                                |                                  |
 | batch_size: 128    | 20355  (total: 1.6 MB)   | 1453 (x 14.00) (total: 115 kB) | 4114 (x  4.95) (total: 325 kB)   |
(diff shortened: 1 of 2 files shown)


projects/stef.md
```diff
@@ -0,0 +1,330 @@
...
+savings significant.
+
+Otel/STEF supports options that allow trading speed for size. At the highest 
+speed setting it can be faster than OTLP (_including_ conversion to/from pdata), while 
+producing significantly smaller payloads.
+
+A draft specification and a prototype implementation of Otel/STEF is attached to thi
+proposal.
+
+### Goals, objectives, and requirements
```
jsuereth, line 25: I think there should be two additional goals here (possibly one if you phrase it well).

- We should evaluating an efficient file-based protocol for direct-export from API with out-of-band "Collector-style" SDK.  That is, imagine an SDK implementation that can serialize events out-of-band quickly and efficiently, and offers more resilience on process-death.   This is what I was working my towards with https://github.com/jsuereth/otlp-mmap/ and I think if we invest in a stateless protocol, this should be a use case it can support.
- Providing guidance to eBPF based telemetry extraction.  If we are able to define structures and buffers and efficient stateful communication, we might be able to provide a good set of primitives for eBPF based event-extraction (complementary to my first bullet point).   This is an avenue I think may be worth exploring.

projects/stef.md
```diff
@@ -0,0 +1,330 @@
...
+- Add support for traces and logs (the prototype implements metrics only).
+- Add Otel/STEF exporter and Otel/STEF receiver to Otel Collector.
+- Publish benchmark-justified guidelines on applicability of Otel/STEF vs OTLP vs Otel
+  Arrow.
+- Refine and stabilize Otel/STEF specification and encourage its implementation
+  inside or outside Otel.
+
+Project non-goals:
+
+- We do not plan to offer Otel/STEF as a general-purpose replacement for OTLP or for Otel
```
jsuereth, line 41: I think I disagree with this. I understand the concern around limiting investment and engineering resources.   However - I think you need to better clarify the target use case in this situation.   Is this something *only* the Go SDK would be able to use with a collector?  Is this something for just Collector->Collector communication?   

That isn't really answered in this proposal, and I think it's critical.

My $.02 is that if we invest in a stateless protocol It should be an Open Standard (e.g. Arrow/parquet) *or* have an OpenTelemetry use case net well served by existing open standards and OTLP.

Here, you demonstrate that STEF can outperform arrow in efficiency for transmitting telemetry data.  **Where in open-telemetry do we use that?**

TL;DR; this should have a targeted set of use cases for the protocol.

jsuereth (commented):
I believe there is a legitimate need in OpenTelemetry for a stateful protocol (or even something as simple as adding dictionaries to OTLP itself).

As discussed before, I'm supportive of investigating STEF and bringing it to a usable state.

However - I think we need to understand our end-game here.  When we evalaute "why not Arrow" or "Why STEF" or "Why not OTLP", I think this proposal is still lacking our primary goal/scope.  It has projects, but not implications of delivery of those projects.

I'd like to make sure we align on where this will be used.  I called out two areas I think could dramatically improve from some low-level, stateful, efficient protocols.  I don't think these are the only targets, but I'm also not sure these were on your radar either.

Let's confirm we agree on end-state then I'm in.

projects/stef.md
tigrannajaryan: >We should evaluating an efficient file-based protocol for direct-export from API with out-of-band "Collector-style" SDK.

That's certainly something we can extend STEF to do. A mmap-ed ring buffer of STEF frames can be that. Since STEF optionally allows full state resets between frames, it essentially has a stateless mode built-in (with resets happening every frame).

>Providing guidance to eBPF based telemetry extraction.

I am not sure I understand this one.

projects/stef.md
tigrannajaryan: >have an OpenTelemetry use case net well served by existing open standards and OTLP.

This is my motivation. OTLP is inefficient size-wise. Otel Arrow Phase 1 is better size-wise, but even better is possible (as the benchmarks show) and is also quite expensive cpu-wise. STEF significantly advances the performance of our network protocols on multiple dimensions, with relatively small investment.

The first use case is as is described in the proposal: smaller wire size (network cost savings), less cpu consumption (compute cost savings) by the Collector. 

Do you think additional explanation is needed for this use case? Or do you want the additional use cases to be described (e.g. the mmap-ed direct-export)?

projects/stef.md
tigrannajaryan: >I think there should be two additional goals here (possibly one if you phrase it well).

I added a more generic goal that says that the project should evaluate additional use-cases.

projects/stef.md
```diff
@@ -0,0 +1,332 @@
...
+
+## Project Board
+
+Project board link to be added here.
+
+## SIG Meetings and Other Info
+
+SIG meeting to be scheduled once the project is approved.
+
+## FAQ
```
yurishkuro, line 276: Has there been a comparative analysis done of other existing protocols besides Arrow?

projects/stef.md
```diff
@@ -0,0 +1,332 @@
...
+
```
yurishkuro, line 277: I the Spec I did not see a list of design principles for STEF protocol. Does such list exist? What is the protocol optimized for? For instance, this doc illustrates speed and size benchmarks, but do they come with trade-offs? What about memory layout, zero-copy capabilities, ability to append data, efficiency of query execution against the data, etc.?

yurishkuro (commented):
I am supportive, but would like to see more clarity on the goals, design principles, and trade-offs. For example, this could be explicitly limited to and optimized for a wire transmission protocol at the expense of other usage patterns, but I did not find these goals clearly stated.

projects/stef.md
```diff
@@ -0,0 +1,332 @@
...
+metric data exists in Go. It demonstrates wire-size and performance improvements over 
+OTLP and Otel Arrow.
+
+A small subset of existing benchmarks is reproduced in this section. For more detailed 
+speed and size tests and comparison to other protocols see
+[benchmarks](https://github.com/tigrannajaryan/stef/tree/main/benchmarks).
+
+The benchmarks are run on the highest compression / slowest speed option.
+
+### Wire Size
```
brancz, line 69: It's unclear to me whether this used the otel-arrow unary or stream mode (where only dictionary portions that are unknown to the server are sent).

brancz (commented):
I'm a bit sceptic, Arrow has extension types, how much have they been explored? It sounds like this format is in philosophy 90% identical to Arrow. This is intended as a mostly internal protocol (since only Go is expected to have an implementation), so even the weirder things we've seen in Arrow would be doable (eg. use a binary array and store custom bytes in them).

Much like @yurishkuro, I'd like to see more on tradeoffs.

projects/stef.md
tigrannajaryan: Benchmarks include comparison with OTLP, Parquet and Otel Arrow. If you think there are other interesting formats to compare to we can add it to the project goals.

projects/stef.md
tigrannajaryan: Stream mode for Otel Arrow.

tigrannajaryan:
Added some more benchmark results in an appendix https://github.com/open-telemetry/community/blob/24716f9c027d51c793bea404354454434579f89d/projects/stef.md#appendix-a-benchmark-results

tigrannajaryan:
> I'm a bit sceptic, Arrow has extension types, how much have they been explored? 

I think this is a question for Otel Arrow SIG.

>It sounds like this format is in philosophy 90% identical to Arrow. This is intended as a mostly internal protocol (since only Go is expected to have an implementation), so even the weirder things we've seen in Arrow would be doable (eg. use a binary array and store custom bytes in them).

Correct, in many ways it is similar to Arrow.

> Much like @yurishkuro, I'd like to see more on tradeoffs.

Added a section with design principles that explains the tradeoffs.

projects/stef.md
tigrannajaryan: Added a label to the table heading to make it clear.

projects/stef.md
tigrannajaryan: Let me know if you would more details.

projects/stef.md
jsuereth: I think we should showcase where stateful protocols are viable (e.g. how does it work through a load balancer?).

While I think stateful protocols have a lot of good use cases, from a broad sense - I have concerns that we may be optimising network overhead at the expense of inflexible network architecture and memory overhead in storing dictionaries from N clients.

I.e. for small, contextual cases this protocol is amazing and should be optimised for such cases.  In broad, highly distributed, cases, we may still need to invest in OTLP optimisations.

To be clear - I think STEF shows a lot of potential and is worth investing in.  I want to be explicit in this proposal where we think we see the biggest benefit and where the trade-offs fall off.  Let's have a target architecture in mind for benchmarks, comparison and 'success'.  I think you have that in your head (and were able to elucide when we talked about this), but it's just not written in the proposal.

tigrannajaryan: @jsuereth I completely agree with you, we need answers to that. I think this line in the list of goals touches that:

>Publish benchmark-justified guidelines on applicability of Otel/STEF vs OTLP vs Otel Arrow.

If this is not enough I can call it out more explicitly.

tigrannajaryan: Added more in https://github.com/open-telemetry/community/pull/2492/commits/bd6c8982dbc3e2e98ef4d6d0315d12fda89931ab

projects/stef.md
```diff
@@ -0,0 +1,471 @@
...
+Dataset `hipstershop_metrics.pb`:
+
+| BATCH SIZE         | OTLP                    | Otel/STEF                       | OTEL ARROW + Stream Mode        |
+|--------------------|-------------------------|---------------------------------|---------------------------------|
+| Compressed (bytes) |                         |                                 |                                 |
+| batch_size: 128    | 4579  (total: 1.6 MB)   | 947 (x  4.83) (total: 323 kB)   | 2067 (x  2.22) (total: 705 kB)  |
+| batch_size: 1024   | 24764  (total: 1.1 MB)  | 5622 (x  4.40) (total: 242 kB)  | 10773 (x  2.30) (total: 463 kB) |
+| batch_size: 2048   | 39325  (total: 865 kB)  | 9209 (x  4.27) (total: 203 kB)  | 17808 (x  2.21) (total: 392 kB) |
+| batch_size: 4096   | 64824  (total: 713 kB)  | 15501 (x  4.18) (total: 170 kB) | 29421 (x  2.20) (total: 324 kB) |
+| batch_size: 16384  | 196877  (total: 591 kB) | 38376 (x  5.13) (total: 115 kB) | 86299 (x  2.28) (total: 259 kB) |
```
jack-berg, line 109: What do you think about updating these to show the result of using OTLP + various compression algorithms. gzip is the default, but I [wrote benchmarks](https://github.com/open-telemetry/opentelemetry-collector/tree/main/config/configgrpc) for a variety of others and had good results. I wonder how Otel/STEF And OTEL Arrow + Stream Mode stack up.

austinlparker:
Heya, super interesting stuff. I have a couple of questions as a GC member --

1. Could you help me understand how this project fits into the project goals, overall? I appreciate that 'stateful OTLP' is useful for many reasons, and I do see that this offers improvements vs. OTLP/Arrow, but given the balance of work that we're trying to tackle in 2025 I'm trying to understand where this fits in.
2. Is the existence of this a result of OTLP/Arrow not being suitable? I agree with some points upthread (e.g., having a stateful 'collector-like' sdk buffer), but it also feels like this being scoped down to just Go/Collector would make it more difficult to achieve that goal.

Appreciate any responses to these points.

jpkrohling:
> but given the balance of work that we're trying to tackle in 2025 I'm trying to understand where this fits in.

This is also a concern of mine: why, and why now? We are constantly saying that we have too many things on our plates: what are we dropping in favor of STEF? Given the Collector's focus on v1, I'd like to get a SIG Collector collective "stamp" before committing to more work involving the Collector (code owners, maintenance, ...). 

I think the tech is really cool and the compression case is very compelling, but I see this as an optimization to be done after we have v1 out of the door.

tigrannajaryan:
Thanks all for comments. I am converting this to a draft for now until I have more clarity on the answers.

tigrannajaryan:
After consulting a number of people we will approach this project somewhat differently. STEF will be a separate OSS project, outside Otel. Otel Collector will merely use it to implement just another receiver/exporter for now and we may do more with STEF in the future.

STEF is now published by Splunk under Apache 2.0 License here: https://github.com/splunk/stef

If you are interested in contributing to STEF please ping me. We will be adding contributing rules, etc soon.

PR #2548, opened 2025-02-03
Add otelbot GitHub App to assets

trask:
Related to #1549 and #2127.

This is more secure than using @opentelemetrybot (a repo can't grant more permissions that the app requests), and can be managed directly in the org (instead of as a separate github account with separate credentials).

May need other GitHub App "bots" with more permissions that are isolated to specific repos.

Opening as draft while I test things out in the Java instrumentation repo...

PR #2569, opened 2025-02-13
Simplify glob pattern

trask:
Follow-up to @chalin's https://github.com/open-telemetry/community/pull/2548#discussion_r1952444032

### DIFF
--- a/assets.md
+++ b/assets.md
@@ -336,11 +336,11 @@ This GitHub App addresses two common issues:
    ```
 
    It is recommended to push to branch names that start with `otelbot/`, and to add a branch protection
-   rule for `otelbot/**/**` with the same setup as documented for
-   [`dependabot/**/**`](docs/how-to-configure-new-repository.md#branch-protection-rule-dependabot).
+   rule for `otelbot/**/*` with the same setup as documented for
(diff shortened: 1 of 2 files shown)


chalin (commented):
Actually, what I was alluding to earlier is that `foo/**/**` could be simplified to `foo/**` to match the branches that I think we want to match. 

It worked in the OTel.io settings:

> image

Note the match given the pattern `opentelemetrybot/**`.

According to the GH docs, https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-rulesets/creating-rulesets-for-a-repository#using-fnmatch-syntax. The base docs of fnmatch are https://ruby-doc.org/core-2.5.1/File.html#method-c-fnmatch.

According to Copilot:

> When applied to branch names, the `**` pattern in fnmatch matches substrings recursively, treating slashes as ordinary characters.

The quote above supports my claim that `**` should be enough.  WDYT?

trask:
I tried `otelbot/**` but it didn't allow me to create a branch named `otelbot/one/two`

chalin (approved):
👍🏻

danielgblanco (approved):

arminru (approved):

PR #2666, opened 2025-04-14
Add Arrow SIG to the README

trask:
@jmacd @lquerel let me know which week you want to start the meetings, and I'll create the calendar invite

PR #2668, opened 2025-04-14
Modify Sampling SIG to every other Thursday

jmacd:
This adjusts the Sampling SIG schedule to alternate with the OTel-Arrow schedule.
This has been discussed in the Sampling SIG with the intent to cancel this week's 4/17/25 meeting, with the next Sampling SIG at 4/24/25.


trask:
merged into #2666

PR #2683, opened 2025-04-16
Move messaging semconv stability SIG to "currently inactive" status

trask:
We put the messaging semconv stability SIG on hold at the beginning of March to focus on wrapping up Database semconv stability.
 
I discussed this week with @lmolkova and we are planning to focus our efforts on RPC semconv stability once Database semconv is marked stable at the end of this month.

Unfortunately, I don't think that leaves bandwidth for the messaging semconv stability SIG at this time, so I think better to officially move it to inactive status and attempt to reboot it once RPC semconv is stable.


lmolkova (approved):
😢

arminru (approved):

mx-psi (approved):

PR #2722, opened 2025-05-09
Add a second OTel-Arrow meeting

jmacd:
A poll was taken in Slack, not much response. Given the current members are mostly in the PT timezone, adding in the late afternoon feels like a good balance.

The first Tuesday meeting will be May 20, after next week's regular Thursday slot.


projects/otelarrow.md
```diff
@@ -74,7 +74,11 @@ Project approvers
 
 ## Meeting times
 
-Every other Thursday at 8AM PT.
+Bi-weekly Thursday at 8AM PT.
+
+Bi-weekly Tuesday at 4PM PT.
+
+See the [Calendar](../README.md#calendar) to find upcoming meetings.
```
cijothomas, line 82: https://github.com/open-telemetry/community?tab=readme-ov-file#implementation-sigs Can you update the table here too?

trask (approved):

PR #2744, opened 2025-05-14
Simplify repo branch protection rule setup

trask:
Removes the `**/**` branch protection rule.

This rule complicates life a lot, as it forces us to create special override rules for `renovate/**/*`, `dependabot/**/*`, and `gh-readonly-queue/main/**`.

I think the key is that we should make sure all "long-term" branches have a branch protection rule similar to `main`, which is an issue I'm tracking and will work on in https://github.com/open-telemetry/sig-security/issues/118.

### DIFF
--- a/assets.md
+++ b/assets.md
@@ -370,13 +370,6 @@ This GitHub App addresses two common issues:
    git config user.email 197425009+otelbot@users.noreply.github.com
    ```
 
-   It is recommended to push to branch names that start with `otelbot/`, and to add a branch protection
-   rule for `otelbot/**/*` with the same setup as documented for
-   [`dependabot/**/*`](docs/how-to-configure-new-repository.md#branch-protection-rule-dependabot).
-
-   > [!WARNING]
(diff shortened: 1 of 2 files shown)


assets.md
```diff
@@ -370,12 +370,8 @@ This GitHub App addresses two common issues:
    git config user.email 197425009+otelbot@users.noreply.github.com
    ```
 
-   It is recommended to push to branch names that start with `otelbot/`, and to add a branch protection
-   rule for `otelbot/**/*` with the same setup as documented for
-   [`dependabot/**/*`](docs/how-to-configure-new-repository.md#branch-protection-rule-dependabot).
-
-   > [!WARNING]
-   > Branch protection rule **ordering** matters, so you will need to delete the `**/**` branch protection rule temporarily, then add the `otelbot/**/*` branch protection rule, then add back the `**/**` branch protection rule.
```
trask, line 378: this is one of the reasons that the `**/**` branch protection rule complicates life

rulesets resolve this complication, but we can't move to rulesets because of https://github.com/orgs/community/discussions/61107#discussioncomment-6569780

trask:
(abandoning this)

PR #2751, opened 2025-05-15
Project Proposal: Browser Phase 1

tedsuo:
OpenTelemetry is rebooting our telemetry efforts for the browser as a sequence of small, highly focused projects. The goal of this first project will be to create instrumentation for the browser runtime and a small initial set of browser libraries, plus the prerequisite API and data modeling work needed to create this instrumentation.

Currently, the proposed meeting time is Thursdays at 8am Pacific. Comment on the doc as to whether this time works for you.

Still needed:

- [x] SIG Sponsor
- [x] Approval for meeting time
- [x] Project Board


mtwo (approved):

projects/browser-phase-1.md
```diff
@@ -0,0 +1,155 @@
...
+
+OpenTelemetry is rebooting telemetry for the Browser as a sequence of small, highly focused projects. The first project will focus on creating instrumentation for the browser runtime and a small set of browser libraries, plus the prerequisite API and data modeling work needed to create this instrumentation.
+
+## Current Challenges
+
+OpenTelemetry currently has a NodeJS-focused javascript implementation that is also capable of running in the browser. However, the requirements for a successful browser observability solution differ enough from NodeJS that a specialized solution is required.
+
+* Loading and unloading in the browser are specialized tasks that are very different from booting and shutting down a NodeJS application. Unloading in particular is under strict time and resource constraints.
+* Memory, cpu, and networking resources are severely constrained, especially on mobile devices.
+* A lack of compression and gRPC in browsers
```
svrnm, line 13:
```suggestion
* A lack of compression and gRPC in browsers
* Sending and offloading telemetry data requires additional considerations, since browser windows can be closed at any time
```

Maybe worth to call this out additionally as well? Or is [beacon](https://caniuse.com/beacon) good enough these days so there is nothing to worry about anymore?

svrnm (approved):

martinkuba-sn:
8am Pacific on Thursdays is in general fine with me 👍

martinkuba:
8am Pacific on Thursdays is in general fine with me 👍

projects/browser-phase-1.md
tedsuo: This is listed above under "Loading and unloading"

svrnm: ah, then I missed that, thanks!

david-luna (approved):

projects/browser-phase-1.md
```diff
@@ -0,0 +1,155 @@
...
+
+We also have a goal of taking an incremental approach, focused on obtaining the highest initial value for the least amount of work. Among participating companies with the OpenTelemetry community, there are already multiple solutions for browser observability. And the current OTel JS SDK does run in the browser.
+
+Rather than start from scratch with a new SDK, we would like to focus our browser work on instrumentation.  OTel instrumentation has only one dependency, the OTel instrumentation API. There is no need to do SDK work in order to stabilize the instrumentation packages we want to provide for our community.
+
+We propose that the various existing browser clients bind to the new OTel Browser API, and we delay work on providing an optimized SDK solution until after we have stabilized a key set of instrumentation packages. This will allow the community to begin receiving value from our OTel browser work much faster than if we were to start by optimizing the SDK.
```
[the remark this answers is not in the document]
tedsuo: One core issue to investigate is whether the current API packages work well in the browser, and what can be done to reduce their package sizes. There's also places where the browser is missing features which are present in NodeJS.

I will clarify that the goal is not to cause the APIs to diverge from the spec, but to make them load better in the browser.

projects/browser-phase-1.md
```diff
@@ -0,0 +1,155 @@
...
+Data Modelling:
+* Sessions
+* Resources / Entities
+* Navigation
+* Anonymous user ids (Manager)
+
+### Support
+
+We will also decide our compatibility story.
+* Which versions of which browsers do we plan to officially support?
```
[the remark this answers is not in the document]
tedsuo: Thanks David. If a feature based approach is better we can explore that. I think it would still be good to publish the resulting version matrix.

projects/browser-phase-1.md
```diff
@@ -0,0 +1,155 @@
...
+* Will we aim support electron at this time?
+
+### Step 2: Core Instrumentation
+
+Once we are clear on what API packages we plan on using, we plan to implement an initial set of instrumentation packages for high priority browser libraries. The goal of this step is to provide enough of an initial ecosystem that end users and vendors can validate the decisions made in Step 1, along with the Semantic Conventions, so that we can stabilize these decisions based on real world feedback.
+
+During this phase, the SIG will not be focused on implementing a browser-optimized SDK. Instead, we will continue to use the existing OTel JS SDK as our reference implementation. Third party SDKs may also choose to bind to the OTel Browser API at this time.
+
+Instrumentation for the browser runtime:
```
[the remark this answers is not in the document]
tedsuo: That is a good question. I think it will need to be on a case by case basis, depending on whether the new instrumentation will be unstable or need to be packaged differently.

tedsuo:
Thanks for the feedback everyone, I've updated the proposal in response.

* Clarified intentions around the API packages
* Clarified our goals relating to code management and existing instrumentation
* Added more members.

I also updated the meeting time to 8:30, with a 30 min meeting being the goal. We are going to experiment with shorter meetings to see if that allows us to move faster and with clearer intentions.

I am OOO until the 27th, I will respond to further feedback at that time. If we have enough approvals, I would like to start the SIG the following week, with June 5th as the kick off meeting.

projects/browser-phase-1.md
```diff
@@ -0,0 +1,161 @@
...
+### Step 1: Browser Fundamentals
+
+Review the current Tracing, Metrics, and Logs API packages. Identify what, if anything, would need to be changed to make these packages optimal for the browser. Discuss the pros and cons of implementing these changes by sharing the existing packages with NodeJS, or forking them to have a separate set of API packages just for the Browser. Please note that the goal of this review is not to deviate from the OpenTelemetry API specification, but to evaluate the practical limitations that the current API packages may place on browser instrumentation.
+
+OpenTelemetry API review:
+* Package size
+* Dependencies
+* Other browser requirements that differ from NodeJS
+
+Data Modelling:
```
danielgblanco, line 41: Would this work need cross-SIG alignment with @open-telemetry/semconv-client-approvers? Or is this more focused towards the implementation of common semconv?

We are trying to decide if the current Client-Side SIG needs to continue existing as a SIG, or if those conversations about cross-SIG semconv can be folded onto the general semconv meeting + Slack channel for async conversations.

For now, we've decided to keep the Client-Side SIG meetings and move it to a fortnightly meeting after the Browser SIG starts.

jpkrohling (changes requested):
I'm supportive of this but I think it should have the same requirements as any other new project. I'm specifically looking for this, from the project template:

> Industry outreach
> Who (people, companies) in the industry should be aware of this effort? Was there an attempt to get them onboard? What did they say?

projects/browser-phase-1.md
```diff
@@ -0,0 +1,161 @@
...
+**Implementation Engineers:**
+* Martin Kuba
+* Purvi Kanal
+* Nev Wylie (Microsoft)
+* Karlie L (Microsoft)
+
+	
+**NodeJS Maintainers and Approvers:**
+
+When bootstrapping our browser work, we want to make sure that the existing OpenTelemetry Javascript community is involved. Existing OTel JS maintainers and approvers should participate in this project to help ensure that it is successful.
```
vigneshshanmugam, line 112: Whats the path to add more folks involved? Is it similar to the existing process?

projects/browser-phase-1.md
```diff
@@ -0,0 +1,161 @@
...
+
+OpenTelemetry is rebooting telemetry for the Browser as a sequence of small, highly focused projects. The first project will focus on creating instrumentation for the browser runtime and a small set of browser libraries, plus the prerequisite API and data modeling work needed to create this instrumentation.
+
+## Current Challenges
+
+OpenTelemetry currently has a NodeJS-focused javascript implementation that is also capable of running in the browser. However, the requirements for a successful browser observability solution differ enough from NodeJS that a specialized solution is required.
+
+* Loading and unloading in the browser are specialized tasks that are very different from booting and shutting down a NodeJS application. Unloading in particular is under strict time and resource constraints.
+* Memory, cpu, and networking resources are severely constrained, especially on mobile devices.
+* A lack of compression and gRPC in browsers
```
vigneshshanmugam, line 13: Just a note - Compression spec is already available in most of the browsers https://developer.mozilla.org/en-US/docs/Web/API/Compression_Streams_API, also leveraged by existing RUM tools.

projects/browser-phase-1.md
```diff
@@ -0,0 +1,161 @@
...
+* Vue.js
+
+### Next Steps
+
+Once we are satisfied that we have achieved the above goals, this SIG will review the remaining work and available resources and choose a new project for the SIG to work on. 
+
+Options include:
+* Continue implementing instrumentation packages.
+* Implement a browser-specific SDK.
+* Implement a more efficient client protocol.
```
vigneshshanmugam, line 87: Not sure if we need it to explicit, just for clarity - what does this mean in practise? There are existing techniques that are used in the browsers that works by shortening the field names, but now with the compression stream spec - the need for that is reduced. These techniques were done without shipping a GRPC specific package to the client as that would influence the package size shipped to users.

danielgblanco: We're talking about OTLP here. For example, currently, OTLP requires different exporters and connections per signals (i.e. one for traces, one for logs, etc). These connections would be coalesced if they go to the same origin, but regardless, certain aspects like _Resource_ attributes need to be duplicated in batches exported for each of those signals, whereas if we had multi-signal OTLP we could do it in one batch. This is just one example, but I expect there will be more things that can be further optimised.

vigneshshanmugam: Thanks for the extra context, makes sense 👍🏽

projects/browser-phase-1.md
tedsuo: Thanks, good to know!

projects/browser-phase-1.md
tedsuo: There are definitely some cross-platform client issues still open. Sessions and user interactions come to mind. Maybe these discussions will pick up speed once the browser SIG starts, but for now it feels like shutting down the Client SIG (or moving to a monthly meeting) is the right choice, as there does not appear to be much active discussion going on.

projects/browser-phase-1.md
tedsuo: Is there specific work you are interested in taking on?

projects/browser-phase-1.md
danielgblanco: Okay, I will work with current client-side SIG leads and update the project doc to reflect this. I'd be happy with the project re-shaping to have a much smaller footprint and maybe something like `Stabilise session semconv` can be an idea. The GH teams and sponsors can remain the same, but meeting at a lower cadence. I won't pollute this project proposal with a thread on what the current client-side SIG will do. Just an understanding that these cross-cutting aspects need a more general place to be discussed, and I think we have that understanding. I will resolve this.

tedsuo:
@open-telemetry/technical-committee this project needs a TC sponsor in order to be approved. 

I believe that that a guiding sponsor would be ideal, but I know that the TC is short staffed at the moment. An escalating sponsor would potentially be sufficient, as the SIG does not intend to make cross-language spec changes, protocol proposals, or other work outside of browser-specific instrumentation and semantic conventions. 

The one area that this SIG overlaps with open spec work is prototyping the EntityProvider, which has an active SIG that we would be collaborating with.

Given that, please advise on how you would like to proceed.

projects/browser-phase-1.md
```diff
@@ -0,0 +1,161 @@
...
+
+### Required Staffing
+
+**Project Lead:** Ted Young
+
+**GC Liaison:** Daniel Gomez Blanco
+
+**Sponsors:**
+* Ted Young
+* SPONSOR NEEDED
```
dyladan: I can sponsor this

```suggestion
* Daniel Dyla
```

tedsuo:
> I'm supportive of this but I think it should have the same requirements as any other new project. I'm specifically looking for this, from the project template:
> 
> > Industry outreach
> > Who (people, companies) in the industry should be aware of this effort? Was there an attempt to get them onboard? What did they say?

Thanks @jpkrohling.  I updated the proposal to include this section.

tedsuo:
> I think this is long overdue and would be happy to sponsor. As a JS maintainer I think I am uniquely qualified to help evaluate what can be changed in the existing JS implementation vs what would need to be reimplemented.

Great! I added you as a sponsor. Given that much of this SIG is specifically focused on thorny issues related to differences between the NodeJS and Browser runtime, I agree that a sponsor such as yourself is critical to the success of this project.

I've also added @carlosalberto as an escalating sponsor to fulfill the TC requirement.

projects/browser-phase-1.md
```diff
@@ -90,15 +90,20 @@ Options include:
 
 ## Staffing / Help Wanted
 
+### Industry Outreach
+
+There are a number of existing RUM/Client observability implementations. Representatives from a wide selection of companies that have experience with browser observability are present in the SIG. This includes Microsoft, New Relic, DynaTrace, HoneyComb, Grafana Labs, and Cisco, among others. We plan for further outreach among end users once we have working code for them to review.
```
jpkrohling, line 98: Some clear leaders in the area are missing. My concern is that we'll, once again, design something without hearing opinions from people we **know** to be relevant in the market.

I realize that some of those voices, like Sentry, have not been exactly friendly towards OTel in the past, which is yet another reason to reach out to them right now: they clearly have opinions on how improve what we have right now in the browser use-case.

I know nothing about this world of RUM, but a quick search also shows me that some of the apparent leaders in the area aren't in the list above. Companies like Datadog, Sematext, PostHog, Akamai, ... I'm confident you know who's relevant in the market, given how close you are to the SIG and this area, so I'd be interested in hearing what they (and others) said when asked to join this effort.

projects/browser-phase-1.md
```diff
@@ -90,15 +90,20 @@ Options include:
...
 **GC Liaison:** Daniel Gomez Blanco
 
 **Sponsors:**
-* Ted Young
-* SPONSOR NEEDED
+* Ted Young - Spec Maintainer Sponsor
+* Daniel Dyla - JS Maintainer Sponsor
+* Carlos Alberto Cortez – TC Escalating Sponsor
 
 **Implementation Engineers:**
```
jpkrohling, line 111: I understand that Martin is from ServiceNow and Purvi from Honeycomb. Didn't the other companies in your list commit to help in the implementation? I see Marco (Grafana) and David (Elastic) in the list of NodeJS maintainers and approvers here (apparently, it doesn't match with the list I see on opentelemetry-js?): are they also committing to work on the implementation?

jpkrohling (changes requested):
Sorry for not approving this yet, I want to make sure the chances of success for this project are higher this time.

projects/browser-phase-1.md
```diff
@@ -0,0 +1,166 @@
...
+
+### Required Staffing
+
+**Project Lead:** Ted Young
+
+**GC Liaison:** Daniel Gomez Blanco
+
+**Sponsors:**
+* Ted Young - Spec Maintainer Sponsor
+* Daniel Dyla - JS Maintainer Sponsor
```
dyladan: I don't think implementation SIG maintainers are necessarily allowed to sponsor other SIGs. I'm offering in my capacity as spec sponsor. I just think being the JS maintainer makes me particularly relevant for this role.

```suggestion
* Daniel Dyla - Spec Sponsor
```

tedsuo: Gotcha!

projects/browser-phase-1.md
tedsuo: @vigneshshanmugam Thank you! Would you be willing to be listed as an implementation engineer? Meaning that you will make some time to write code in addition to reviewing designs?

projects/browser-phase-1.md
tedsuo: Ok, I can review the implementation engineers section with the client SIG and update it with more commitments.

I'm not sure what you mean about NodeJS approvers section? Everyone listed in that section is part of the javascirpt approvers team. Can you clarify?

projects/browser-phase-1.md
```diff
@@ -0,0 +1,166 @@
...
+* A lack of compression and gRPC in browsers
+* Package and dependency sizes
+* All activity occurring in the program relates to a single user instead of many independent transactions. So far this has led to a number of spec developments, such as the need for mutable resources for session management, a better logging model for recording events, and may lead to more browser specific design requirements.
+* Sessions in particular are a very important lifecycle for clients which are not present as a concept in server-side programs.
+* Sessions persist across multiple page loads, another difficult problem to design a solution for that is browser specific.
+* NodeJS has built in facilities for tracking context across async boundaries, but browsers do not have an equivalent concept.
+
+## Goals, Objectives, and Requirements
+
+OpenTelemetry needs a javascript implementation optimized for the browser. Sharing code with NodeJS is helpful when possible. But going forward, we want to prioritize running well in the browser over code reuse with the NodeJS project. If there are places where code reuse with NodeJS is a hard requirement, those places should be identified explicitly.
```
reyang, line 22: This is vague, we should be explicit about the technical direction of API package - which approach will this SIG take?

* The SIG aims to produce a shared OpenTelemetry API package that can be used by both web browsers and Node.js services.
* The SIG aims to produce two separate OpenTelemetry API packages, one for web browser and one for Node.js services.
* The SIG doesn't know the answer yet and would expect to figure this out later.

There are many libraries (e.g., NPM packages) that target both web browsers and Node.js services, these libraries will need an OpenTelemetry API package for instrumentation.

tedsuo: As stated later in the doc, the SIG needs to review this situation. That review is the first deliverable.

projects/browser-phase-1.md
danielgblanco: In my opinion, this is why I think the scope of this "phase 1" is key, and in particular the aspect around not pushing for current vendors to abandon their SDK implementations, but rather allow for current implementations to provide a bridge to standard APIs.

This approach not only solves a very immediate problem for end-users (i.e. having a standard API for custom telemetry on top of vendor SDKs), but I also believe that having success scenarios where instrumentation libs are adopted can help ease industry outreach.

Ultimately, I believe that with the contributors listed below we have SMEs that can help steer this in the right direction, and be an example for the rest of the industry.

projects/browser-phase-1.md
jpkrohling: I believe a subsequent commit fixed the inconsistencies I've seen. Thanks for handling that!

projects/browser-phase-1.md
jpkrohling: There has been a discussion in the GC about this, and the outcome as I understand is that we (GC) believe that enough people in the industry know about this effort. 

Personally, I would have wished for an informal, friendly ping to make them aware of this, but I also understand that the GC doesn't see this as a requirement for new projects.

I'm marking this as resolved, and approving the PR.

mx-psi (approved):

projects/browser-phase-1.md
```diff
@@ -0,0 +1,169 @@
...
+* Purvi Kanal (Honeycomb)
+* Hector Hernandez (Microsoft)
+* David Luna (Elastic)
+
+### Additional SIG Members
+
+**Design review:**
+* Ram Thiru (Microsoft)
+* Santosh Cheler (Cisco / Splunk)
+* Marco Schäfer (Grafana Labs)
```
danielgblanco:
```suggestion
* Marco Schäfer (Grafana Labs)
* Jordan Porter (New Relic)
```
cc @metal-messiah

projects/browser-phase-1.md
dyladan: My personal opinion is that a separate web and node API is a non-starter for exactly this reason:

> There are many libraries (e.g., NPM packages) that target both web browsers and Node.js services, these libraries will need an OpenTelemetry API package for instrumentation.

I would much prefer to improve the existing API package even if that means a 2.0

dyladan: we can discuss this in the SIG though

projects/browser-phase-1.md
```diff
@@ -0,0 +1,169 @@
...
+* Carlos Alberto Cortez – TC Escalating Sponsor
+
+**Implementation Engineers:**
+* Martin Kuba (ServiceNow)
+* Purvi Kanal (Honeycomb)
+* Karlie L (Microsoft)
+* Ken Rimple (Honeycomb)
+* Jared Freeze (Embrace)
+* Joaquín Díaz (Embrace)
+
```
Abinet18, line 119: Can I be added here , Abinet Debele(Cisco)

projects/browser-phase-1.md
```diff
@@ -0,0 +1,169 @@
...
+* Daniel Dyla - Spec Maintainer Sponsor (Dynatrace)
+* Carlos Alberto Cortez – TC Escalating Sponsor
+
+**Implementation Engineers:**
+* Martin Kuba (ServiceNow)
+* Purvi Kanal (Honeycomb)
+* Karlie L (Microsoft)
+* Ken Rimple (Honeycomb)
+* Jared Freeze (Embrace)
+* Joaquín Díaz (Embrace)
```
reyang: We should add GitHub aliases, similar to what we've been doing in the past (e.g. https://github.com/open-telemetry/community/blob/main/projects/config.md#required-staffing)

projects/browser-phase-1.md
```diff
@@ -0,0 +1,169 @@
+# Browser Instrumentation (Phase 1) Proposal
+
+## Background and Description
+
+OpenTelemetry is rebooting telemetry for the Browser as a sequence of small, highly focused projects. The first project will focus on creating instrumentation for the browser runtime and a small set of browser libraries, plus the prerequisite API and data modeling work needed to create this instrumentation.
+
+## Current Challenges
+
+OpenTelemetry currently has a NodeJS-focused javascript implementation that is also capable of running in the browser. However, the requirements for a successful browser observability solution differ enough from NodeJS that a specialized solution is required.
```
reyang, line 9: I don't see Web Assembly being mentioned, maybe that'll be Phase 3.

I want to call out that the browser observability story isn't just about JavaScript. In other words, if this group designed some new APIs just for JavaScript in this Browser Phase 1, we might soon realize that we need to redesign these APIs for web assemblies.

projects/browser-phase-1.md
tedsuo: Great!

projects/browser-phase-1.md
tedsuo: Correct, we are not working on web assembly during this phase of development. I disagree that web assembly would have any bearing on the OpenTelemetry APIs we are designing here for javascript applications running in the browser. Can you explain why it would be relevant?

reyang: It is probably a question regarding "is web assembly part of the Browser project or not?". I can see three answers:
1. It is. Web assembly will be considered in a later phase, not Browser Phase 1 though.
2. It is not. Web assembly is a totally separate topic, and we are not going to consider it as part of the Browser support (although web assembly works for web browsers).
3. We don't know the answer yet, we just want to not think/worry about it for now.

projects/browser-phase-1.md
tedsuo: Thanks!

projects/browser-phase-1.md
```diff
@@ -0,0 +1,171 @@
...
+* @hectorhdzg (Microsoft)
+* @david-luna (Elastic)
+
+### Additional SIG Members
+
+**Design review:**
+* @ramthi (Microsoft)
+* @scheler (Cisco / Splunk)
+* @codecapitano (Grafana Labs)
+* metal-messiah (New Relic)
```
reyang, line 135:
```suggestion
* @metal-messiah (New Relic)
```

reyang (approved):

projects/browser-phase-1.md
tedsuo: Ok, thank you for the clarification. I believe that web assembly would be an entirely separate topic. The issue there is that instrumented code compiled with web assembly would use the OpenTelemetry APIs in that language. For example, a C++ library would be instrumented with Otel C++, not with Otel Browser JS.

On its own, a complete web assembly application using OpenTelemetry would probably work. However, it is not clear to me how context propagation would work between separate libraries compiled using web assembly and then mixed together. Not to mention, sharing an SDK. There are probably a lot of issues with this approach. Most likely, this would be it's own specialized implementation.

I will list it as a potential future topic, and as something to consider when we review the API design. But I don't think that we should focus on it as an implementation target at this time. In general, we are trying to keep this project tightly scoped. For example, even an SDK for the browser is out of scope.

carlosalberto: Is this resolved @reyang ?

projects/browser-phase-1.md
tedsuo: Resolving this comment so that we can merge the PR, but I'm looking forward to working with you @vigneshshanmugam!

vigneshshanmugam: Missed this comment, yes happy to get involved in both the implementation and design. Thanks for the poke.

Issue #2623, opened 2025-03-20
[Donation Proposal]: (Mini project) Transcripting


moxious:
### Description

At Grafana in developer advocacy we use a simple set of scripts to fetch all YouTube transcripts and check the textual transcripts of what's in videos into a repo.  [You can see that here](https://github.com/grafana/developer-advocacy/blob/main/transcripts/README.md)

We do this for a number of reasons:
1. We use it to draft early first drafts when we're working on some new docs; basically, interview engineers for an hour, and then use the rich back & forth as grist for a blog post, a new docs page, whatever
2. We have "expert Q&A LLM answer agents" which can be fed everything that the technical experts said on video.  This expands the reach of automatic Q&A techniques [like this](https://community.grafana.com/t/grafana-alloy-for-sending-windows-event-logs/143337/3).

It would not be that hard to contribute this code to OTel for its YouTube channel if that's desirable.  What you'd get is a small set of python scripts & instructions, plus a directory of the community's choosing where all the transcripts would be checked in (effectively the output of the scripts).

This makes "video greppable" for the community, and enables any subsequent downstream LLM approaches community may want to use.

## Mini POC 

I took [this really cool OTel video](https://www.youtube.com/watch?v=qn4x0DgG5SI&list=PLVYDBkQ1TdyzFwkcit__AcBnMJ8lgF7L1) (thanks @reese-lee!) and pulled its transcript

```
$ cat otel/2023-09-28T04\:00\:36Z-otel-end-user-discussions-amer-january-2023.md | aichat --prompt "Please create a list of questions that are raised in this video transcript; omitting the answers"

1. How do you deal with helping in languages you're not an expert in?
2. How do you define an enabler in a company?
3. Have other companies experienced this struggle with language expertise?
4. Do you create champions in those languages you're not familiar with?
5. Have those people you worked with come back with questions or seek information on their own?
6. How long did it take for individuals to get up to speed with observability concepts?
7. Is auto-instrumentation a good approach to help people get started with OpenTelemetry?
8. What strategies can improve understanding of the purpose and value of OpenTelemetry within a team?
9. Would having real-world case documentation around OpenTelemetry be helpful?
10. Should there be a periodic forum for pointed Q&A with experts in the community?
11. How are you dealing with clock/time drift in data?
12. Is it useful to create a processor that detects data points from the future?
13. What are best practices for bifurcating data in a pipeline using filters and conditions?
14. What would be the advantage of using a router solution versus a filter processor for data separation?
[...]

PR #2777, opened 2025-05-23
WIP: Transcripting code donation

moxious:
Associated Issue:  https://github.com/open-telemetry/community/issues/2623#issuecomment-2903180203

tl;dr what is this?  It's a small python script with instructions that fetches YouTube transcripts and summarizes them into nice Markdown files. The intent here is to store public text associated with those videos.  This is nice by itself, but when combined with this: https://github.com/open-telemetry/opentelemetry.io/pull/6769 it gets better.  Kapa can be trained on these, and OTel has a sustainable way to do Q&A on the website _based on video_.

Core of this PR is the python code which isn't that big.  Most of the line changes are actual markdown files which are _the output of the python code_.


moxious:
Current known limitations:  this works by pulling a raw youtube transcript and then summarizing/cleaning up.  So when the raw youtube transcript is imperfect (which it often is with names) errors do happen.  And so "Reese Lee" becomes sometimes "Ree Lee" and "Adriana Villela" becomes "Adriana Villa".  Both the "nice cleaned up version" and the "very messy YouTube original" are included for comparison (and also so  it's harder for OpenAI to fool me)

svrnm:
> As this is aimed at YouTube transcripts, and I see how it can be really useful for the content the End-User SIG publishes, would it make more sense if this PR is opened against [open-telemetry/sig-end-user](https://github.com/open-telemetry/sig-end-user?rgh-link-date=2025-05-27T13%3A51%3A40Z) ?
> 
> cc @avillela @reese-lee

Not all recordings are from End User SIG right? I think we can start with community and later see if there is better places to have them

danielgblanco:
You're right. We do have YouTube videos that come from Comms SIG. However, as those tends to refer to documentation, do we think this tool is equally useful there? My thinking of putting this in a repo that's not `community` is that it'd make permissions easier to maintain those scripts.

trask:
hi @moxious, can you send this PR to https://github.com/open-telemetry/sig-end-user instead? thanks

moxious:
yep.  Traveling now, but will do early next week.

PR #2810, opened 2025-06-11
Clarify required sections of the project proposal template

mx-psi:
Clarifies what sections from the project template are required, and which are to be filled after approval

### DIFF
--- a/project-template.md
+++ b/project-template.md
@@ -22,7 +22,7 @@ In general, OTEPs are not accepted unless they come with working prototypes avai
 
 Who is currently planning to work on the project? If a project requires specialized domain expertise, please list it here. If a project is missing a critical mass of people in order to begin work, please clarify.
 
-### Industry outreach
+### Industry outreach (Optional)
 
 Who (people, companies) in the industry should be aware of this effort? Was there an attempt to get them onboard? What did they say?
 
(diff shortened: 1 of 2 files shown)


project-template.md
```diff
@@ -22,7 +22,7 @@ In general, OTEPs are not accepted unless they come with working prototypes avai
 
 Who is currently planning to work on the project? If a project requires specialized domain expertise, please list it here. If a project is missing a critical mass of people in order to begin work, please clarify.
 
-### Industry outreach
+### Industry outreach (Optional)
```
danielgblanco, line 25: Should we add these as markdown comments underneath the title?

danielgblanco (approved):

project-template.md
mx-psi: Do you mean in addition to adding it to the title or instead of adding it to the title?

svrnm: I still struggle with making this optional, I would rather like to see a structured process like:

* before start of the project, list who is aware already and who should be aware
* (optional) list whom you have reached out to so far/whom you plan to reach out to before the project starts
* (post approval, mandatory) work with SIG communications to make an announcement of the project via blog and social media

cc @open-telemetry/docs-approvers

mx-psi: > I still struggle with making this optional, I would rather like to see a structured process like:

I agree with you, but I would rather have this discussion in a separate PR. As it stands today, this is optional based on GC discussion. We can change that, but this PR tries to solve the previous problem: clarifying which parts are required and clarifying the relationship between the project management doc (which seems to be authoritative) and the template (which does not seem to be authoritative)

mtwo: I agree with @svrnm's comments, I wouldn't make this optional, we should just clarify what's required

mx-psi: I mean, the reality is that this is optional today, right? We can change it, but I just want to reflect the reality today

jpkrohling: I agree with @mx-psi.

I'd love to reignite the discussion about what we are expected to do as leaders of the project in ensuring that our community is not only accepting new contributors, but actively seeks to hear the voices of those who are known in the space where new projects are being proposed, which is what the current wording attempts to convey.

But the reality is that, today, we are not required to do this when proposing a new project. We assume people know about this new project because we believe we are big enough for people to follow closely what we are doing, and they'll naturally join. This PR marking this field as optional reflects this reality.

jpkrohling (approved):

trask (approved):

mtwo (approved):

project-template.md
mx-psi: Given the approval from Morgan I am going to mark this as resolved

mx-psi:
Counting myself we have 5 GC members on board with this, so I am merging

PR #2824, opened 2025-06-22
Add missing token permissions

trask:


.github/workflows/reusable-markdown-link-check.yml
trask, line 1: I inlined this into the caller, didn't seem worth a separate workflow

PR #2859, opened 2025-07-10
Add github admin process

### LINKED ISSUE #2849: Establish process for making project-wide Github security changes
There have recently been three security-related changes recently that, while well intentioned and positive for the security of the project as a whole, appear to have broken the Collector and Go workflows without prior warning and without reversal, in two cases affecting the last Collector release:
1. #2747 affected the Collector release process, despite me having left a comment on us using this process.
2. #2793 affected the Collector and Go workflow for requesting reviews on contrib
3. https://github.com/open-telemetry/community/issues/2805#issuecomment-2968174990 (going for an org secret instead of the global Docker token) affected our release process in a different way  (I am not 100% sure if this is SIG Security related since this was not documented on a separate issue)
I think the changes are positive, but the way of implementing them has been a bit disruptive. 

Unless there is an active security incident/it affects graduation or something like it, I would like to propose that we document and stick to something like the following rules:
1. (For @open-telemetry/admins) The default for 'best practices gone wrong' should be to revert the change until we fix the issues.
2. (For @open-telemetry/sig-security-maintainers) SIG Security should give guidance to SIGs on how to align with new rules if existing SIG practices cannot be continued.
3. (Collaboration between SIGs and @open-telemetry/sig-security-maintainers / @open-telemetry/admins) We also need to have an agreed upon timeline and participation from SIGs in the decision making process in case we need changes (or we find we need changes). The split I would propose is as follows:
    1. SIG Security decides on the security practices, while taking into account the input from individual SIGs
    2. @open-telemetry/admins implements these changes and document any escape hatches if there are any
    3. If applying a particular security change proves to be disruptive, individual SIGs commit to a plan and timeline to make the necessary changes to override a particular claim. SIGs can't just react immediately to issues so they need to be given time.

trask:
Resolves #2849

### DIFF
--- a/guides/maintainer/github-admin-processes.md
+++ b/guides/maintainer/github-admin-processes.md
@@ -1,0 +1,43 @@
+# GitHub Administration Process
+
+This document outlines how the GitHub administration process is handled within the OpenTelemetry organization.
+
+## Principles
+
+1. **Transparency**: Document and communicate all changes
+2. **Collaboration**: Involve affected SIGs in decision-making processes
+3. **Responsiveness**: Move quickly on maintainer requests
+
+## Guidelines
+
+When possible, changes should be implemented by sending a PR to [open-telemetry/admin](https://github.com/open-telemetry/admin) (updating the appropriate Terraform files).
+
+These changes include:
+
+- Repository settings (general, collaborators and teams, branch protections)
+- Basic GitHub team information (name, description, parent team)
+
+Other changes should be requested or documented via a [community issue](https://github.com/open-telemetry/community/issues).
+
+These include:
+
+- Installing apps
+- Repository secrets and variables
+- Organization settings
+
+**Note**: GitHub team memberships can be updated directly by maintainers and do not require a community issue.
+
+### Organization-Level Changes
+
+For changes that affect all repositories in the organization, the admin PR or community issue should be left open for **a minimum of one week** to allow for community feedback before implementation.
+
+Potentially disruptive changes should also be posted to the [#otel-maintainers](https://cloud-native.slack.com/archives/C01NJ7V1KRC) Slack channel when the community issue or admin PR is opened.
+
+### Single Repository Changes
+
+For changes affecting a single repository:
+- **If a repository maintainer** opens the admin PR or community issue, the change will be implemented as soon as possible
+- **If a GitHub admin** opens the admin PR or community issue and a repository maintainer approves it (including via positive emoji on the issue), the change will be implemented as soon as possible
+- **If a GitHub admin** opens the admin PR or community issue and no maintainer replies, the PR or issue should be left open for **a minimum of one week** before implementation
+- **If someone else** opens the community issue and a repository maintainer approves it (including via positive emoji on the issue), the change will be implemented as soon as possible
+- **If another maintainer objects after implementation** and asks to revert the change, it will be reverted until there is agreement on moving forward (following the [conflict resolution process](conflict-resolution.md) if necessary)


guides/maintainer/github-admin-processes.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Principles
+
+1. **Transparency**: Document and communicate all changes
+2. **Collaboration**: Involve affected SIGs in decision-making processes
+3. **Responsiveness**: Move quickly on maintainer requests
+
+## Guidelines
+
+When possible, changes should be implemented by sending a PR to [open-telemetry/admin](https://github.com/open-telemetry/admin) (updating the appropriate Terraform files).
```
mx-psi: Maybe an obvious comment but we could say that we will restrict permissions to manually change settings whenever possible to have a single source of truth on open-telemetry/admin?

guides/maintainer/github-admin-processes.md
```diff
@@ -0,0 +1,43 @@
...
+Potentially disruptive changes should also be posted to the [#otel-maintainers](https://cloud-native.slack.com/archives/C01NJ7V1KRC) Slack channel when the community issue or admin PR is opened.
+
+### Single Repository Changes
+
+For changes affecting a single repository:
+- **If a repository maintainer** opens the admin PR or community issue, the change will be implemented as soon as possible
+- **If a GitHub admin** opens the admin PR or community issue and a repository maintainer approves it (including via positive emoji on the issue), the change will be implemented as soon as possible
+- **If a GitHub admin** opens the admin PR or community issue and no maintainer replies, the PR or issue should be left open for **a minimum of one week** before implementation
+- **If someone else** opens the community issue and a repository maintainer approves it (including via positive emoji on the issue), the change will be implemented as soon as possible
+- **If another maintainer objects after implementation** and asks to revert the change, it will be reverted until there is agreement on moving forward (following the [conflict resolution process](conflict-resolution.md) if necessary)
```
mx-psi: I think it wouldn't hurt to encourage SIGs to commit to a timeline if they ask for a revert (or to make some specific change if known), something like "we would like to revert this change; we will work on fixing this problem within the next two weeks"

guides/maintainer/github-admin-processes.md
```diff
@@ -0,0 +1,43 @@
...
+
+Other changes should be requested or documented via a [community issue](https://github.com/open-telemetry/community/issues).
+
+These include:
+
+- Installing apps
+- Repository secrets and variables
+- Organization settings
+
+**Note**: GitHub team memberships can be updated directly by maintainers and do not require a community issue.
```
arminru:
```suggestion
**Note**: Repository-specific (e.g., approvers) GitHub team memberships can be updated directly by maintainers and do not require a community issue.
```

guides/maintainer/github-admin-processes.md
svrnm: Comment on the same line: We should call out here, that the repo is private but accessible by all maintainers, so the "transparency" is project-internal and not public

reyang: > Maybe an obvious comment but we could say that we will restrict permissions to manually change settings whenever possible to have a single source of truth on open-telemetry/admin?

I've sent a series of PRs (only OpenTelemetry project maintainers can access these links) to remove admin permission from the maintainers.

* https://github.com/open-telemetry/admin/pull/115
* https://github.com/open-telemetry/admin/pull/116
* https://github.com/open-telemetry/admin/pull/117
* https://github.com/open-telemetry/admin/pull/118
* https://github.com/open-telemetry/admin/pull/119
* https://github.com/open-telemetry/admin/pull/120
* https://github.com/open-telemetry/admin/pull/121
* https://github.com/open-telemetry/admin/pull/122

trask: > Maybe an obvious comment but we could say that we will restrict permissions to manually change settings whenever possible to have a single source of truth on open-telemetry/admin?

It's a bit more than this even, if you update something manually, the next time terraform runs your change will get wiped out.

trask: > We should call out here, that the repo is private but accessible by all maintainers

good point, added this

guides/maintainer/github-admin-processes.md
trask: 👍

PR #2875, opened 2025-07-20
Add injector project

atoulme:
This adds the injector project to the list of implementation SIGs. I have offered a time to meet weekly, 9:30am PT on Mondays.

Multiple items are missing to complete this PR:
* [x] Google doc for the meeting notes (to be created by the GC or TC?)
* [x] calendar-injector google group
* [x] Name and github handle of GC liaison

Fixes https://github.com/open-telemetry/opentelemetry-injector/issues/14

PR #2901, opened 2025-07-31
Add repositories to sigs.yml

trask:
This was requested as part of the CNCF graduation work.

A couple of repos don't have a SIG

- https://github.com/open-telemetry/opentelemetry-injector - SIG being added in #2875

- https://github.com/open-telemetry/opentelemetry-helm-charts - @open-telemetry/helm-maintainers I think we should add you to https://github.com/open-telemetry/community#implementation-sigs?

and these repos which probably make sense not to have a SIG:

- https://github.com/open-telemetry/community
- https://github.com/open-telemetry/.github
- https://github.com/open-telemetry/admin

### DIFF
--- a/sigs.yml
+++ b/sigs.yml
@@ -371,6 +371,9 @@
       - https://github.com/open-telemetry/opentelemetry-collector
       - https://github.com/open-telemetry/opentelemetry-collector-contrib
       - https://github.com/open-telemetry/opentelemetry-collector-releases
+      - https://github.com/open-telemetry/opentelemetry-go-build-tools
+      - https://github.com/open-telemetry/opentelemetry-go-vanityurls
+      - https://github.com/open-telemetry/govanityurls
   - name: 'C++: SDK'
     meeting: Alternating between Monday at 13:00 PT and Wednesday at 09:00 PT
     notes:


sigs.yml
```diff
@@ -567,6 +638,9 @@
     gcLiaison:
     - name: Ted Young
       github: tedsuo
+    repositories:
+      - https://github.com/open-telemetry/opentelemetry-network
+      - https://github.com/open-telemetry/opentelemetry-network-build-tools
```
trask: looks like this is about to be archived

@open-telemetry/network-maintainers do you want me to go ahead and archive it?

```suggestion
```

sigs.yml
```diff
@@ -692,6 +780,10 @@
     gcLiaison:
       - name: 'Austin Parker'
         github: austinlparker
+    repositories:
+      - https://github.com/open-telemetry/sig-project-infra
+      - https://github.com/open-telemetry/changelog.opentelemetry.io
+      - https://github.com/open-telemetry/assign-reviewers-action
```
trask, line 788: @austinlparker @dyladan ok to group these two under the project infra SIG?

sigs.yml
```diff
@@ -418,6 +453,13 @@
     gcLiaison:
     - name: Ted Young
       github: tedsuo
+    repositories:
+      - https://github.com/open-telemetry/opentelemetry-go
+      - https://github.com/open-telemetry/opentelemetry-go-contrib
+      - https://github.com/open-telemetry/opentelemetry-go-build-tools
+      - https://github.com/open-telemetry/opentelemetry-go-vanityurls
+      - https://github.com/open-telemetry/opentelemetry-proto-go
+      - https://github.com/open-telemetry/govanityurls
```
trask, line 465: I think these are co-owned with the Collector SIG:

- opentelemetry-go-build-tools
- opentelemetry-go-vanityurls
- govanityurls

@austinlparker is it ok to list a repository under two different SIGs?

reyang (approved):

sigs.yml
yonch: Amazing catch! Was going to add a ticket next week. We (belatedly) realized it would be easier for everybody to fold the build tools into the main repo..

Yes please 🙏 .

svrnm (approved):

jaronoff97:
@trask for the helm-charts, i think implementation makes sense. I can open a PR.

sigs.yml
trask: I've gone ahead and added these repos also to the Collector SIG as that seems to make the most sense for us at least

arminru (approved):

PR #2902, opened 2025-08-01
[docs/guides/maintainers] Clarify admin workflow for repositories owned by multiple SIGs

mx-psi:
### What does this PR do?

Requires approval from one representative from each SIG for admin changes for repositories owned by multiple SIGs.

Follows #2859.

Relates to open-telemetry/admin/pull/143.

### DIFF
--- a/guides/maintainer/github-admin-processes.md
+++ b/guides/maintainer/github-admin-processes.md
@@ -43,7 +43,7 @@ community issue or admin PR is opened.
 
 ### Single Repository Changes
 
-For changes affecting a single repository:
+For changes affecting a single repository owned by a single SIG:
 - **If a repository maintainer** opens the admin PR or community issue, the change will be
   implemented as soon as possible
 - **If a GitHub admin** opens the admin PR or community issue and a repository maintainer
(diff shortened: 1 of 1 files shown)


danielgblanco (approved):

trask (approved):

mx-psi:
cc @open-telemetry/go-maintainers @open-telemetry/collector-contrib-maintainers

songy23 (approved):

edmocosta (approved):

MrAlias (commented):
Shouldn't this be decided for each repository?

I do not think we should require this in https://github.com/open-telemetry/opentelemetry-go-build-tools given there are tools there only used by one SIG and not the other. Requiring both approve a PR there will likely never be followed or hinder development.

This also seems to overstep for simpler PRs like dependency updates. If a repository has a well defined CI system and can verify dependency change PRs I don't think requiring approval from multiple SIG representatives is needed. This similarly applies to other "trivial" type PRs.

MrAlias:
How does this get applied to https://github.com/open-telemetry/opentelemetry-proto-go where there are PRs for the profiling signal changes that are not Go or Collector SIG specific?

trask:
@MrAlias I believe this is only be scoped to repo administration (including PRs to https://github.com/open-telemetry/admin, but excluding PRs to the actual repos themselves)

MrAlias:
@trask thanks for the clarification 👍

mx-psi:
Sorry, my description wasn't the clearest, fixed now. I do not believe this requires GC majority and we already have approval from the affected SIGs maintainers, so I will wait for a few more days and merge this if there are no further comments.

PR #2904, opened 2025-08-01
Add Helm SIG

jaronoff97:
Adds the helm sig to the readme and sigs.yml let me know if i missed anything! Created from #2901

### DIFF
--- a/sigs.yml
+++ b/sigs.yml
@@ -12,6 +12,10 @@
     invites: calendar-spec-general
     sponsors: []
     gcLiaison: []
+    repositories:
+      - https://github.com/open-telemetry/opentelemetry-specification
+      - https://github.com/open-telemetry/opentelemetry-proto
+      - https://github.com/open-telemetry/build-tools
   - name: 'Specification: Sampling'
     meeting: Thursday at 08:00 PT
     notes:
@@ -46,6 +50,8 @@
     gcLiaison:
     - name: Daniel Gomez Blanco
       github: danielgblanco
+    repositories:
+      - https://github.com/open-telemetry/opentelemetry-configuration
   - name: 'Specification: Logs'
     meeting: Tuesday at 10:00 PT
     notes:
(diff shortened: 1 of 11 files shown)


sigs.yml
```diff
@@ -595,6 +595,16 @@
...
+  - name: Kubernetes Helm Charts
+    meeting: ''
+    notes:
+      type: none
+    chat:
+      - type: slack
+        name: '#otel-helm'
+        id: C03HVLM8LAH
+    sponsors: []
+    gcLiaison: []
```
danielgblanco: I'm happy to volunteer as GC liaison for this SIG.

sigs.yml
```diff
@@ -595,6 +595,16 @@
     gcLiaison:
     - name: 'Juraci Paixão Kröhling'
       github: jpkrohling
+  - name: Kubernetes Helm Charts
+    meeting: ''
```
danielgblanco: Is the SIG operating fully async, i.e. no meetings?

jaronoff97: yep, that's the current state. I sometimes discuss helm things in the operator sig

jaronoff97:
@danielgblanco thank you for being the liaison! Updated to reflect that. 🙇

sigs.yml
trask: we could say something like "Meets during Kubernetes Operator SIG"

there's some precedent for this: Semantic Conventions: Security | Meets during Semantic Conventions: General

trask (approved):

sigs.yml
```diff
@@ -595,6 +595,20 @@
...
+    meeting: "Meets during OpenTelemetry Operator"
+    notes:
+      type: gDoc
+      value: 1Unbs2qp_j5kp8FfL_lRH-ld7i5EOQpsq0I4djkOOSL4
+    chat:
+      - type: slack
+        name: '#otel-helm'
+        id: C03HVLM8LAH
+    invites: calendar-k8s-operator
+    sponsors: []
```
danielgblanco, line 688: cc @open-telemetry/technical-committee

jaronoff97:
Talked over this with @TylerHelmuth and he mentioned the collector SIG also fields helm questions, so i'm going to add that in too.

TylerHelmuth (approved):

sigs.yml
trask: I checked and the TC is ok with sorting out sponsorship later since this SIG has been around for a while already

PR #2954, opened 2025-08-28
add anchor links to SIGs

svrnm:
Minor change to the listing of SIGs in the README, which builds on top of @trask's addition to have an anchor per SIG.

I had the situation a few times now that I wanted to reference a SIG from that list directly, so I wanted to have that anchor, but to get it, I either have to type it out or go to the page source view and copy it out. This adds a small `#` at the side of each SIG name to have it accessible.

Note that I tried different variants, but the markdown renderer does not allow the usage of classes or styles as well as the usage of `<small>`, so the super script seemed to be the best solution


README.md
```diff
@@ -130,76 +130,76 @@ SIG-specific GitHub discussions.
...
-| <a id="sig-client-side-telemetry"></a>Client Instrumentation | Every other Tuesday at 09:00 PT | [Google Doc](https://docs.google.com/document/d/16Vsdh-DM72AfMg_FIt9yT9ExEWF4A_vRbQ3jRNBe09w) | [#otel-client-side-telemetry](https://cloud-native.slack.com/archives/C0239SYARD2) | [calendar-client-side](https://groups.google.com/a/opentelemetry.io/g/calendar-client-side) |  | [Daniel Gomez Blanco](https://github.com/danielgblanco) | 
+| Specification: General + OTel Maintainers Sync<a id="sig-specification" href="#sig-specification"><sup>#</sup></a> | Tuesday at 08:00 PT | [Google Doc](https://docs.google.com/document/d/1pdvPeKjA8v8w_fGKAN68JjWBmVJtPCpqdi9IZrd6eEo) | [#otel-specification](https://cloud-native.slack.com/archives/C01N7PP1THC) | [calendar-spec-general](https://groups.google.com/a/opentelemetry.io/g/calendar-spec-general) |  |  |
```
arminru: Nit: I think it would be more nicely rendered with a space before the `#` (might also be better for screen readers and translation tools?). Now it looks a bit odd to me, but either should be ok.

image

svrnm: > Nit: I think it would be more nicely rendered with a space before the `#` (might also be better for screen readers and translation tools?). Now it looks a bit odd to me, but either should be ok.

I think the problem with that was that it sometimes did a line break, but I could give a `&nbsp;` a try

trask: maybe an icon, e.g. <a href="">:link:</a>?

it's too bad we can't suppress the link underlining

svrnm: I can give the icon a try!

svrnm: done, PTAL!

scripts/update-sig-tables.py
```diff
@@ -96,9 +96,9 @@ def format_chat(chat):
             calendar = f"[{invites}](https://groups.google.com/a/opentelemetry.io/g/{invites})"
         
         if group_name == "Specification SIGs":
-            markdown_content += f"| <a id=\"{short_name}\"></a>{name} | {meeting} | {notes} | {chats} | {calendar} | {tc_sponsors} | {gc_liaison} | \n"
+            markdown_content += f"| {name}<a id=\"{short_name}\" href=\"#{short_name}\">&nbsp;<sup>🔗</sup></a> | {meeting} | {notes} | {chats} | {calendar} | {tc_sponsors} | {gc_liaison} | \n"
```
trask: (so it doesn't underline the space)
```suggestion
            markdown_content += f"| {name}&nbsp;<a id=\"{short_name}\" href=\"#{short_name}\"><sup>🔗</sup></a> | {meeting} | {notes} | {chats} | {calendar} | {tc_sponsors} | {gc_liaison} | \n"
```

scripts/update-sig-tables.py
```diff
@@ -96,9 +96,9 @@ def format_chat(chat):
...
         else:
-            markdown_content += f"| <a id=\"{short_name}\"></a>{name} | {meeting} | {notes} | {chats} | {calendar} | {gc_liaison} |\n"
+            markdown_content += f"| {name}<a id=\"{short_name}\" href=\"#{short_name}\">&nbsp;<sup>🔗</sup></a> | {meeting} | {notes} | {chats} | {calendar} | {gc_liaison} |\n"
```
trask:
```suggestion
            markdown_content += f"| {name}&nbsp;<a id=\"{short_name}\" href=\"#{short_name}\"><sup>🔗</sup></a> | {meeting} | {notes} | {chats} | {calendar} | {gc_liaison} |\n"
```

PR #2957, opened 2025-08-29
Add RPC SIG

trask:


sigs.yml
```diff
@@ -174,6 +174,24 @@
         github: danielgblanco
     roadmapProjectIDs:
       - 79
+  - name: 'Semantic Conventions: RPC'
+    meeting: Every Wednesday at 05:00 PT
+    notes:
+      type: gDoc
+      value: 1zTi_Z4WBisytPnXTpwtnGOK-4Gv5oSTuMHvIQ8--5hg
```
lmolkova: it points to DB stabilization google doc, I don't mind reusing it, just wanted to make sure it's intentional

sigs.yml
```diff
@@ -174,6 +174,24 @@
...
+        id: C09D3TRKMED
+    invites: calendar-semconv
+    sponsors:
+      - name: Liudmila Molkova
+        github: lmolkova
+    gcLiaison:
+      - name: Trask Stalnaker
+        github: trask
+    roadmapProjectIDs:
+      - TODO
```
lmolkova: created https://github.com/orgs/open-telemetry/projects/161/ along with start/target dates update

sigs.yml
trask: nope, fixed!

sigs.yml
```diff
@@ -174,6 +174,24 @@
         github: danielgblanco
     roadmapProjectIDs:
       - 79
+  - name: 'Semantic Conventions: RPC'
+    meeting: Every Wednesday at 17:00 PT
+    notes:
+      type: gDoc
+      value: 11AtPWuaNGl2MfhRwgWqNbnjAPtK06jMKWh5-_YbWuHE
```
lmolkova, line 181: Please share with write access (it doesn't seem to be shared at all)

trask: done, can you check?

lmolkova: yay, it works now :)

PR #3009, opened 2025-09-24
Use opentelemetry group for election roll

danielgblanco:
The current election voter roll script uses the `All` repository group when querying for voter eligibility (taken from [this view](https://opentelemetry.devstats.cncf.io/d/9/developer-activity-counts-by-repository-group-table?orgId=1&var-period_name=Last%20year&var-metric=contributions&var-repogroup_name=All&var-country_name=All)). This used to return all contributors, including those with less than 20 contributions.

The current behaviour seen is that this same filter now stops at a certain spot depending on the range of the query, and when using the `Last year` filter, it stops at 25.

The change in this PR changes the `All` repository group to `opentelemetry`, which still contains contributions to all repositories, to a lowest value of 13 (see [this view](https://opentelemetry.devstats.cncf.io/d/9/developer-activity-counts-by-repository-group-table?orgId=1&var-period_name=Last%20year&var-metric=contributions&var-repogroup_name=open-telemetry&var-country_name=All)).

While a better approach is needed for next year (especially considering how things are moving to LFX Insights) I believe this can solve this issue for this year's election.

### DIFF
--- a/scripts/gc-elections/generate-voters-roll.py
+++ b/scripts/gc-elections/generate-voters-roll.py
@@ -47,7 +47,7 @@ def get_users_and_contributions():
                     "uid": "P172949F98CB31475",
                     "type": "postgres"
                 },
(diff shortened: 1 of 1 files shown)


danielgblanco:
I've also raised https://github.com/cncf/devstats/issues/108 to help clarify if there's been an intended change here we should be aware of.

trask (approved):

mtwo (approved):

PR #3010, opened 2025-09-24
Remove TC Chair position

jsuereth:
The Technical committee has not operated with a chair since its founding. We have recently set up a rotation schedule for running both the Specification and TC meetings which has been performing the duties of the TC chair in our charter.

This PR simply removes the TC Chair position and denotes this rotational responsibility.

If we decide we want/need  chair position going forward we can add it back later.


lmolkova (approved):

svrnm (approved):

carlosalberto (approved):

mx-psi:
We have GC majority (5/9) and TC majority (5/8, counting author), merging

PR #3053, opened 2025-10-06
[elections/2025] Remove previous account from GC voter list

kamphaus:
After leaving Jemmic I no longer have access to the @christophe-kamphaus-jemmic GitHub account and I'm only using @kamphaus.


mx-psi (approved):
Would be great to get an approval from the other account but I guess that's impossible given your PR's description :sweat_smile:

PR #3071, opened 2025-10-14
Self-nomination: Ted Young for Governance Committee 2025

tedsuo:
Self nomination for GC election 2025


svrnm:
I support.

austinlparker:
I support

PR #3081, opened 2025-10-21
Add Develocity instance to assets.md

trask:


reyang (approved):

reyang:
Looks like https://github.com/open-telemetry/community/blob/main/.cspell.yaml needs to be updated with the new term "Develocity".

PR #3002, opened 2025-09-24
2025 GC Election Page

mtwo:
Adds the main page for the 2025 GC Election

PR #3082, opened 2025-10-22
Correct number of candidates voters can select

mx-psi:
Correct the number of candidates voters can select from four to five.

Overlooked at #3002

Thank you @andrzej-stencel for reporting this.


andrzej-stencel (approved):
❤️

PR #3097, opened 2025-10-27
Injector project file

tedsuo:
Looks like we accepted the donation of the [OpenTelemetry Injector](https://github.com/open-telemetry/opentelemetry-injector), but we never added a project file to represent what the SIG is aiming to achieve.

This PR aims to fix that. It clarifies the scope of the SIG, the initial set of goals, and the timeline for initial deliverables.

### DIFF
--- a/projects/injector.md
+++ b/projects/injector.md
@@ -10,7 +10,7 @@ System operators are end users who have access to deployment and management tool
 for their organization's services, but do not have access to the source code or
 command line invocations for those services.
 
-Currently, system operators have no mechanism for installing OpenTelemetry clients.
+Currently, system operators have very limited choices for installing OpenTelemetry clients.
 There are several language-specific installation efforts supported by the
 OpenTelemetry Operator, providing coverage for a limited number of use cases. Using
 LD_PRELOAD, we can extend this set of use cases to six languages, and remove the


projects/injector.md
```diff
@@ -0,0 +1,185 @@
...
+client consists of a language-specific OpenTelemetry SDK plus a set of
+instrumentation packages that matches the set of libraries used by the target
+application. The correct set of SDK plugins also need to be installed and
+configured to match the correct settings for their deployment environment.
+
+System operators are end users who have access to deployment and management tools
+for their organization's services, but do not have access to the source code or
+command line invocations for those services.
+
+Currently, system operators have no mechanism for installing OpenTelemetry clients.
```
jack-berg: #nit: This sentence is too pessimistic of a framing. As you note below, the operator provides this capability for select set of languages running on k8s.

projects/injector.md
```diff
@@ -0,0 +1,185 @@
...
+requirement.
+
+Note: This SIG is concerned with the OpenTelemetry Injector mechanism itself. We
+will provide a proof of concept of how the injector can be delivered via system
+packages. We do not believe that this SIG should maintain any system packages beyond
+the scope needed for demonstrating the utility of the Injector. For the actual
+packaging of OpenTelemetry, best practices should be established based on community
+feedback. For example, we believe that deciding how to use the OpenTelemetry
+Injector with Kubernetes should be discussed as part of the OpenTelemetry Operator
+SIG.
```
jack-berg, line 27: The way I framed this when I was doing the [TC due diligence](https://docs.google.com/document/d/1BqH4xFuyiWMXbaWoTUXhjZp0v2NINPXQiFsM2xfwyUY/edit?tab=t.0#heading=h.mohuxdmv1nw2) for the injector donation was that in an ideal world, the injector is a utility that can be used in a stand alone capacity for people running on bare linux, and is also a dependency of the operator. This project should collaborate with the operator SIG in whatever capacity is needed to make sure that the injector can end up being integrated into the operator, and doesn't end up as a niche tool only for bare linux users.

projects/injector.md
```diff
@@ -0,0 +1,185 @@
...
+
+## Background and description
+
+LD_PRELOAD-based injection of instrumentation into application runtimes is a
+technique that has been widely in use in the APM industry. It requires no elevated
+security privileges (LD_PRELOAD shared objects run with the same privileges of their
+host process) and is known to work in a variety of runtimes: Java, Node.js, .NET,
+Python, Ruby, Elixir and PHP. LD_PRELOAD-based injection is generally not applicable
```
jack-berg, line 35: May be worth mentioning that the reason LD_PRELOAD works with these language runtimes is that LD_PRELOAD allows a program to highjack `getenv` system calls, and all these language runtimes have environment variable based mechanisms for injecting instrumentation.

The secret sauce of LD_PRELOAD is that it allows a program to read ELF headers and evaluate whether the application links `glibc` or `musl` and select the appropriate instrumentation artifact accordingly.

projects/injector.md
```diff
@@ -0,0 +1,185 @@
...
+
+System operators are end users who have access to deployment and management tools
+for their organization's services, but do not have access to the source code or
+command line invocations for those services.
+
+Currently, system operators have no mechanism for installing OpenTelemetry clients.
+There are several language-specific installation efforts supported by the
+OpenTelemetry Operator, providing coverage for a limited number of use cases. Using
+LD_PRELOAD, we can extend this set of use cases to six languages, and remove the
+requirement that the users be running kubernetes. Running linux will still be a
```
mx-psi:
```suggestion
requirement that the users be running Kubernetes. Running Linux will still be a
```

trask:
cc @open-telemetry/injector-approvers

projects/injector.md
tedsuo: I think that's what we're saying in this paragraph? It's definitely what we intend to be saying. We're creating the tool in this SIG, but we'll work through the Operator SIG to do the packaging for Kubernetes.

projects/injector.md
tedsuo: good point!

maryliag (approved):

Issue #3060, opened 2025-10-08
Change time of K8s SemConv SIG meeting


ChrsMark:
There is a collision of the [K8s SemConv SIG](https://github.com/open-telemetry/community/blob/main/projects/k8s-semconv.md) meeting with the TC meeting where @dashpole is a required participant. In this, we will need to look for a new time.

Looking into the [calendar](https://github.com/open-telemetry/community?tab=readme-ov-file#calendar), I see 2 options:

- bring the K8s SIG meeting 30 mins earlier
- move the K8s SIG meeting on Tuesdays same time

@open-telemetry/semconv-k8s-approvers could you provide your feedback on this please? Feel free to provide any other alternatives.

dashpole:
I can't make 30 minutes earlier.  Tuesday I can do the 30 minutes earlier time, otherwise this would conflict with the spec sig.

skl:
I'm not an approver, but Tuesday 30 min earlier also works for me! 😄

ChrsMark:
@jinja2 @dmitryax since you are attending the sig meetings could you share your preference/availability?

jinja2:
I can do Tuesdays too

dmitryax:
Tuesday the same time works for me

ChrsMark:
> Tuesday I can do the 30 minutes earlier time, otherwise this would conflict with the spec sig.

@dashpole I guess that's fine if we overlap with the Specification meeting since the K8s one is bi-weekly? 

From the answers above I think Tuesdays same time is our best bet?

dashpole:
I would rather not miss the spec sig meetings, at least not the first 30 minutes

ChrsMark:
Verified with @jinja2 and @dmitryax about Tuesday 30 mins earlier. We are going to use this time.

PR #3105, opened 2025-10-30
Change K8s SIG meeting schedule to Tuesdays

ChrsMark:
Updated the schedule for the K8s Semantic Conventions SIG to meet every other Tuesday at 07:30 PT.

Fixes https://github.com/open-telemetry/community/issues/3060.

I've updated the calendar event already.


mx-psi (approved):

arminru:
Please update https://github.com/open-telemetry/community/blob/main/sigs.yml which is the source from which the `README.md` is populated (hence the failing https://github.com/open-telemetry/community/actions/runs/18934076786/job/54056659367?pr=3105 check).

sigs.yml
```diff
@@ -115,7 +115,7 @@
     roadmapProjectIDs:
       - 55
   - name: 'Semantic Conventions: K8s'
-    meeting: Every other Wednesday at 08:00 PT
+    meeting: Every other Monday at 07:30 PT
```
jinja2:
```suggestion
    meeting: Every other Tuesday at 07:30 PT
```
I think we all agreed upon Tuesday

ChrsMark: You are right. No idea why I put Monday there 😬

PR #3123, opened 2025-11-04
[chore] update sigs.yml with cicd new meeting time

adrielp:


trask (approved):

scripts/update-sig-tables.py
```diff
@@ -1,10 +1,10 @@
 #!/usr/bin/env python3
-import pip
+import subprocess
 import sys
 
 # in the Makefile we use a unmodified python container to run this script, so we need to install pyyaml if it's not already installed
 if (len(sys.argv) > 1) and (sys.argv[1] == "--install"):
-    pip.main(['install', 'pyyaml'])
+    subprocess.check_call([sys.executable, '-m', 'pip', 'install', 'pyyaml'])
```
arminru, line 7: In case the other reviewers are wondering too - this appears to be the recommended approach as per https://pip.pypa.io/en/stable/user_guide/#using-pip-from-your-program.

Issue #3182, opened 2025-12-05
Migrate all community google docs to the OpenTelemetry.io Google Workspace


trask:
Currently community google docs are split between the (older) cncf-opentelemetry-governance@lists.cncf.io Google Account and the (newer) OpenTelemetry.io Google Workspace.

This has caused confusion on multiple occasions, and it would be nice to only have community docs in one place (the OpenTelemetry.io Google Workspace).

I wrote (or rather guided copilot to write) an automation for this and tested it with the Sampling SIG doc (which was causing problems this week because it was owned by yet another account).

It copies the doc into the OpenTelemetry.io Google Workspace, and then posts a redirect alert to the old doc:

<img width="1151" height="453" alt="Image" src="https://github.com/user-attachments/assets/5ae1a2c7-ce07-435d-ba7b-7d34dfcb7730" />

(and then updates the link in this repo's sigs.yml/README.md)

mx-psi:
To the extent that we can, could we make the original doc read-only/comment-only?

trask:
> To the extent that we can, could we make the original doc read-only/comment-only?

I was thinking to go with @andrzej-stencel's suggestion to delete the content in the original doc: https://github.com/open-telemetry/community/pull/3183#pullrequestreview-3548961001

mx-psi:
That also works!

trask:
Notes:
- do both (delete the original content + make the doc read-only)
- update link in meeting invites

PR #3183, opened 2025-12-06
Migrate some meeting notes to the OpenTelemetry.io Google Workspace

### LINKED ISSUE #1147: SIG documents should be owned by an OpenTelemetry account

trask:
Part of https://github.com/open-telemetry/community/issues/3182

Started with docs which weren't even owned by an OpenTelemetry account.

Closes #1147 (along with #3181 which I sent earlier for the Sampling SIG)

cc @open-telemetry/prometheus-interoperability @open-telemetry/lambda-extension-approvers @open-telemetry/collector-approvers @open-telemetry/collector-contrib-approvers @open-telemetry/cpp-approvers @open-telemetry/cpp-contrib-approvers @open-telemetry/python-approvers


README.md
```diff
@@ -157,8 +157,8 @@ SIG-specific GitHub discussions.
...
-| C++: SDK&nbsp;<a id="sig-cpp" href="#sig-cpp"><sup>🔗</sup></a> | Alternating between Monday at 13:00 PT and Wednesday at 09:00 PT | [Google Doc](https://docs.google.com/document/d/1i1E4-_y4uJ083lCutKGDhkpi3n4_e774SBLi9hPLocw) | [#otel-cpp](https://cloud-native.slack.com/archives/C01N3AT62SJ) | [calendar-cpp](https://groups.google.com/a/opentelemetry.io/g/calendar-cpp) | [Severin Neumann](https://github.com/svrnm) |
+| Collector&nbsp;<a id="sig-collector" href="#sig-collector"><sup>🔗</sup></a> | Alternating between Tuesday at 17:00 PT, Wednesday at 09:00 PT, and Wednesday at 05:00 PT | [Google Doc](https://docs.google.com/document/d/1AeWpwxjMPm8jOAZymCB3BrwwmNiJiKOO7dc4EvVq3x0) | [#otel-collector](https://cloud-native.slack.com/archives/C01N6P7KR6W) | [calendar-collector](https://groups.google.com/a/opentelemetry.io/g/calendar-collector) | [Pablo Baeyens](https://github.com/mx-psi) |
```
andrzej-stencel, line 160: Should the [archive doc](https://docs.google.com/document/d/1un2i9RmyCJ6J5cs7MoBjwi1GXAN2DLMgu3Z-0kViN70) also be moved, is it in the correct place? The archive doc has historical meeting notes from July 2019 to March 2024.

trask: done

trask:
> I wonder if the original docs' contents should be deleted, leaving only the red frame with the move announcement in it? Otherwise I'm afraid ultimately someone will ignore the red frame and update the old doc.

done

codeboten (approved):

arminru:
> Thanks Trask!
> 
> I wonder if the original docs' contents should be deleted, leaving only the red frame with the move announcement in it? Otherwise I'm afraid ultimately someone will ignore the red frame and update the old doc.

@trask I would also suggest making the old docs with just the red frame read-only. If they remain publicly editable, they might be vandalized and stay publicly defaced without us noticing. This already happened in the past with some abandoned document.

trask:
Can you add this comment to the issue so I don't forget? Want to merge this.

PR #3208, opened 2026-01-06
Add a note about tables being auto-generated

cijothomas:
Its easy to miss this when making changes, so trying to make it easier... Not sure if this is considered good practice. Feel free to close, if not desired.

### DIFF
--- a/.cspell.yaml
+++ b/.cspell.yaml
@@ -230,6 +230,7 @@ words:
   - Quesma
   - raesene
   - reiley
+  - readmemd
   - roadmaps
   - rolldice
   - ruech


svrnm (approved):

cijothomas:
If this works, then I want to propose similar to https://github.com/open-telemetry/opentelemetry-specification/blob/main/spec-compliance-matrix.md . That is the other place which is not super obvious that its generated content. Sem-conventions also, but it was relatively easy to spot that (maybe just my observation!)

README.md
```diff
@@ -125,6 +125,9 @@ SIG-specific GitHub discussions.
 > The meeting times in the tables below are given in 24-hour notation.
 > Meetings are either in Pacific Time (PT), with Daylight Saving Time, or UTC+8, without Daylight Saving Time.
 
+> [!NOTE]
+> The tables below are auto-generated. To make changes, see [CONTRIBUTING.md](./CONTRIBUTING.md#updating-sig-information-on-the-readmemd).
+
```
mx-psi: We could even make this an HTML comment, since the only people that care about this are people that are looking at the source code already?
```suggestion
<!-- The tables below are auto-generated. To make changes, see CONTRIBUTING.md#updating-sig-information-on-the-readmemd -->
```

mx-psi (approved):

README.md
svrnm: @cijothomas wdyt?

cijothomas: agreed!

mx-psi (approved):

PR #3215, opened 2026-01-09
Add project ID for Blueprints project

danielgblanco:
This adds OTel Blueprints project ID so that it shows up in the [OpenTelemetry Roadmap](https://github.com/orgs/open-telemetry/projects/158), along with target dates for the project.


danielgblanco:
Looks like errors are unrelated to this change. Possibly transient errors? I cannot re-run the check though.

PR #3241, opened 2026-01-28
Update End-User SIG meeting times to include OTel Blueprints

danielgblanco:
Updates times to include the new agreed meeting slot for discussions specific about OTel Blueprints


danielgblanco:
Looks like there's a 404 on an unrelated resource

PR #3266, opened 2026-02-11
Update GC charter: Goals to Mission and 'How'

svrnm:
@open-telemetry/governance-committee PTAL! This is long overdue as our current charter still states that we have an "initial role". I used our meeting notes + repo content + [k8s steering charter](https://github.com/kubernetes/steering/blob/main/charter.md) as input for this to have a draft created by LLM. I think it is a good starting point, but all of it is open for debate.

### DIFF
--- a/governance-charter.md
+++ b/governance-charter.md
@@ -17,7 +17,7 @@ strategic direction. The GC serves as a steward for the project's
 [mission, vision, and values](./mission-vision-values.md) and defines its
 non-technical governance structures.
 
-## How
+## Approach
 
 - Adapt the role and structure of the Governance Committee as needed to meet the
   evolving needs of the project. Governance processes are never complete;


governance-charter.md
```diff
@@ -1,24 +1,35 @@
...
+## Mission
+
+The Governance Committee is the governing body of the OpenTelemetry project,
+providing decision-making and oversight pertaining to the project's governance
+policies, [project management](./project-management.md), community health, and
+strategic direction. The GC serves as a steward for the project's
+[mission, vision, and values](./mission-vision-values.md) and defines its
+non-technical governance structures.
+
+## How
```
arminru:
```suggestion
## Approach
```
Perhaps?

austinlparker (approved):

PR #3323, opened 2026-03-10
update npm ownership

maryliag:


assets.md
```diff
@@ -156,9 +156,9 @@ Link: https://www.myget.org/feed/Packages/opentelemetry
 Link: https://www.npmjs.com/settings/opentelemetry/packages
 
 - Ask any of the following people if you need access
-- Owner: Mayur Kale @mayurkale22
-- Admin: Daniel Dyla @dyladan
-- Member: Bogdan Drutu @bogdandrutu
+- Owner: OpenTelemetry account `npmjs-account` (GC/TC have access to it)
+- Member: Daniel Dyla @dyladan
+- Member: Marc Pichler @pichlermarc
```
trask, line 161: does this need to be owned by GC/TC, or can it be owned by SIG similar to the other package management accounts?

maryliag: I create a google group (similar to what we did for python), and then created an account on npm with that email.
I can still share the pwd with JS maintainers, but I stored the user/pwd into the 1Password account we have access, so in case they need, we have access to share with them.

trask: why store it in GC 1Password Vault instead of storing it in the JS 1Password Vault?

maryliag: There isn't a JS 1Password vault, because there was never a need for it. 
For npm package management, the `member` role allows everything that maintainers need, so they don't need access to the owner account, they can publish, deprecate, etc, on their own. 
The `owner` account is in case something goes wrong (e.g. accounts being compromised)

trask: thanks for the explanation, btw there is a JS 1Password vault, but if JS folks don't need access to the owner account, I think it's ok to store it in GC account if that's their preference, though I think also ok to store in the JS 1Password account which is what other SIGs with the master keys for their respective package managers

maryliag:
ci error it's unrelated, there was no changes on links in this PR

assets.md
trentm, line 161:
```suggestion
- Member: Marc Pichler @pichlermarc
- Member: Trent Mick @trentm
```

Happy to be on this npmjs org as well, but I don't *need* to be. /cc @pichlermarc

pichlermarc: Sounds good to me :) 
We should review from time to time to avoid having too many people on the list though. 

More people -> larger attack surface. Phishing is a real concern with these things.

PR #3357, opened 2026-04-04
ci: add workflow concurrency to stale action to prevent overlapping runs

hesxo:
#### Description

Add workflow-level concurrency to the stale issue management workflow to prevent overlapping scheduled runs.

This ensures only one stale job runs at a time per repository, which helps avoid duplicate stale comments, conflicting label updates, and unnecessary GitHub Actions usage.

#### Link to tracking issue

N/A

#### Testing

Validated the workflow YAML update.
Confirmed the concurrency block is defined at the workflow level and applies to the scheduled stale job.

### DIFF
--- a/.github/workflows/issue-management-stale-action.yml
+++ b/.github/workflows/issue-management-stale-action.yml
@@ -5,14 +5,18 @@ on:
     # Hourly at minute 23
     - cron: "23 * * * *"
 
+concurrency:
+  group: stale-${{ github.repository }}
+  cancel-in-progress: true
+
 permissions:
   contents: read
 
 jobs:
   stale:
     permissions:
-      issues: write  # for actions/stale to close stale issues
-      pull-requests: write  # for actions/stale to close stale PRs
+      issues: write # for actions/stale to close stale issues
+      pull-requests: write # for actions/stale to close stale PRs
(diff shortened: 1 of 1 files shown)


hesxo:
The failing `markdown-link-check` job appears unrelated to this change.

This PR only updates `.github/workflows/issue-management-stale-action.yml` to add workflow-level concurrency for the stale action.

The reported failures are from existing broken links in:
- `assets.md`
- `projects/resources-and-entities.md`
- `README.md`

Examples from the job log:
- Oracle docs fragment not found
- Google Doc 404
- Google Slides 404
- `open-telemetry/opentelemetry-kotlin/discussions` 404

So this looks like a pre-existing repository-wide link check failure rather than an issue introduced by this PR.

arminru:
Hi @hesxo! Thanks for your PR.

Have parallel runs ever occurred before and if so, is there any example where this has been an issue?

The cron schedule is set to run once an hour and each run only takes a few seconds, so I wonder if there is any potential for this to happen to begin with:

https://github.com/open-telemetry/community/blob/6d44ee3df36df59fdd2a5ce1adb406191d406608/.github/workflows/issue-management-stale-action.yml#L3-L6

PR #3346, opened 2026-03-27
[projects] Add details for Collector MCP project

mx-psi:
Fixes #3330

PR #3361, opened 2026-04-08
Ensure projects are reviewed promptly

jsuereth:
This updates the project review process.

- Adds automation to place project proposals in the `triage:tc-inbox`
- Add a new label to denote when TC review comments have been places: `triage:tc-reviewed`
- Update project process documentation to denote expectations of what TC will provide during this review and its scope.

### DIFF
--- a/.github/workflows/proposal-triage.yml
+++ b/.github/workflows/proposal-triage.yml
@@ -1,4 +1,4 @@
-name: Issue management - Triage Proposals
+name: Issue management - Triage Project Proposals
 
 on:
   pull_request:
--- a/.github/workflows/proposal-triage.yml
+++ b/.github/workflows/proposal-triage.yml
@@ -1,0 +1,39 @@
+name: Issue management - Triage Proposals
+
+on:
+  pull_request:
+    types: [opened, reopened]
+    paths:
+      - 'projects/**'
+
+permissions:
+  contents: read
+
+jobs:
+  triage:
+    permissions:
+      pull-requests: write
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check for reviewed label
+        id: check_label
+        env:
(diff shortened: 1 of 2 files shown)


.github/workflows/proposal-triage.yml
```diff
@@ -0,0 +1,39 @@
+name: Issue management - Triage Proposals
+
+on:
+  pull_request:
+    types: [opened, reopened]
+    paths:
+      - 'projects/**'
```
mx-psi, line 7: This also picks up PRs such as #3346. This may be fine, but you could also check that the PR introduces net new files as a way to tell if the PR is about a new project

trask (approved):

.github/workflows/proposal-triage.yml
mx-psi: Discussed offline; it looks like the logic would be too complex so I'll resolve this (wasn't meant to be blocking but I'd like to make it clear :) )

mx-psi (approved):

jmacd (approved):

dashpole (approved):

.github/workflows/proposal-triage.yml
```diff
@@ -0,0 +1,39 @@
+name: Issue management - Triage Proposals
```
arminru:
```suggestion
name: Issue management - Triage Project Proposals
```

arminru (approved):

maryliag (approved):

.github/workflows/proposal-triage.yml
jsuereth: Discussed this offline - I think that's a wonderful improvement, it was beyond my ability in the "time box" I had for this.  I think it's ok to start with a bit of noise and we can fix that up later.

tedsuo (approved):

svrnm (approved):

PR #3382, opened 2026-04-22
Move to pull_request_target - where we don't use unauthroized code fo…

jsuereth:
…r traige flow, and we can have write permissions.

### DIFF
--- a/.github/workflows/proposal-triage.yml
+++ b/.github/workflows/proposal-triage.yml
@@ -1,7 +1,7 @@
 name: Issue management - Triage Project Proposals
 
 on:
-  pull_request:
+  pull_request_target:
     types: [opened, reopened]
     paths:
       - 'projects/**'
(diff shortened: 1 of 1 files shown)


.github/workflows/proposal-triage.yml
```diff
@@ -1,7 +1,7 @@
 name: Issue management - Triage Project Proposals
 
 on:
-  pull_request:
+  pull_request_target:
```
jack-berg, line 6: I like to leave comments on workflows that use `pull_request_target` to remind future maintainers of the risk associated for secret exfiltration

jsuereth: Good point - I'll add that.

Something like:

```
// This workflow should NOT ever run code from an untrusted source and care needs to be taken crafting this workflow and reviewing it to ensure secrets are not exfiltrated.
```

trask (approved):

jack-berg (approved):

mx-psi (approved):
LGTM modulo adding the comment above

.github/workflows/proposal-triage.yml
jsuereth: Added

jsuereth:
Comment added - I wasn't sure how to make sure it shows up easily when reviewing diffs on the file, but I put it at the top - best option considering.

Issue #2830, opened 2025-06-23
Standardize section listing maintainer, approver, etc, across all repo README files


trask:
Since the list of GitHub team members is not public (for people outside of the GitHub org), it's helpful to have a canonical place where we display those team members publicly (e.g. https://github.com/open-telemetry/community/issues/2814#issuecomment-2997398045).

We could create this list somewhere centrally and sync it regularly from the GitHub team memberships, but we already have something that _almost_ works, since most repos already list the maintainers and approvers at the bottom of their README files.

I'd like to propose standardizing this section across all repos:

The most important part:
- Anchors for `#maintainers`, `#approvers`, `#triagers` (if any)

The less important part, that I propose doing at the same time (since I'll probably attempt to automate this):
- Have these anchors as L3 headings under an L2 "Contributing" heading
- Have them in this order: maintainers, approvers, triagers, emeritus maintainers, emeritus approvers, emeritus triagers
- Don't display the github team name (I think it's unnecessary, and just adds a bit of extra clutter)
- Include the line "For more information about the X role, see ..."
- Have an L3 heading at the end of this section "Thanks to all of our contributors!" with the contributor graph

You can see an example with all the suggestions here:
https://github.com/trask/opentelemetry-python/tree/contributing-section-updates#contributing

pellared:
Here is how it looks in OTel Go: https://github.com/open-telemetry/opentelemetry-go/blob/main/CONTRIBUTING.md#approvers-and-maintainers

Per https://github.com/open-telemetry/opentelemetry-go/issues/6854 we also include the maintainers' GPG keys.

reyang:
> You can see an example with all the suggestions here: https://github.com/trask/opentelemetry-python/tree/contributing-section-updates#contributing

@trask I think the example needs to be fixed - we don't want to keep company information for emeritus. I think we should call out that for emeritus we want to remove company info.

![Image](https://github.com/user-attachments/assets/d2ef524d-a66b-4802-9336-7e097e5fefd1)

PR #3396, opened 2026-04-29
run script in all repos

maryliag:
Test run was [successful](https://github.com/open-telemetry/community/actions/runs/25138755894/job/73683386840) with only a few repos. It was also tested with multiple executions and no new PR was created (as expected).

PRs generated:
https://github.com/open-telemetry/opentelemetry-java/pull/8347
https://github.com/open-telemetry/opentelemetry-js/pull/6649
https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3500

Now this PR remove the list of repos, so it can be executed in all repos under the org `open-telemetry`


trask (approved):

scripts/move-to-emeritus.py
```diff
@@ -1025,7 +1025,7 @@ def _build_pr_body(repo, changes, cutoff, warning=None):
     body += "\n"
 
     body += (
-        "**Note:** After merging, remove the users from the listed team(s) in GitHub and from any relevant private channels on Slack.\n\n"
+        "**IMPORTANT:** After merging, remove the users from the listed team(s) in GitHub and from any relevant private channels on Slack.\n\n"
```
martincostello: Using an Alert would make it "pop" in the UI more: https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts

mx-psi (approved):

scripts/move-to-emeritus.py
maryliag: Nice

thompson-tomo:
@maryliag this run appears to have a couple of edge cases:

- some teams use * for list so markdownlint fails. Should the guidance be to use -?
- some documents are not having names being removed & instead they are just being added to emeritus ie https://github.com/open-telemetry/opentelemetry-cpp-contrib/pull/611

martincostello:
> * some teams use * for list so markdownlint fails. Should the guidance be to use -?

Not every repo has the same lint requirements so I don't think we can make everyone happy - SIGs can fix-up PRs if they want to proceed with merging them.

> some documents are not having names being removed

Everyone's documents are different - again, SIGs can fix-up anything extra.

thompson-tomo:
Yes but there was a completed attempt to standardise the contributing section across all repos https://github.com/open-telemetry/community/issues/2830

martincostello:
So for the sake of one list in one file everyone needs to change the markdown linting rules for their entire repo?

thompson-tomo:
No i didn't suggest that. I am purely talking about the sections which this script is changing/managing.

The most likely scenario I see would be to suppress the violation of that markdownlint rule within a section of the doc. Then within that section have it as a - so that alphabetical sorting etc works.

maryliag:
@thompson-tomo ideally all readmes would use `-`. If the repo wants to update to that, great, otherwise I won't enforce this.
I'm adding an update to the script to deal with both scenarios (`*` and `-`), which is why some nomes did not got removed, because the script didn't recognize the pattern with `*`. If there are open PRs with this issue, feel free to commit any fixes and following iterations should handle properly

thompson-tomo:
Thanks @maryliag just for completness it might be worthwhile to also cater for the + symbol which is a valid list character -> https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md004---unordered-list-style

PR #3398, opened 2026-04-30
Add Slack member ID to people.yml

reyang:
This PR adds Slack member ID to all the existing members listed in the `people.yml` file.

The member ID can be verified by using `https://cloud-native.slack.com/team/<id>`, for example, this is my profile https://cloud-native.slack.com/team/U01NA9DUCLT.

This change is based on a recent discussion in the Technical Committee meeting. The TC member who takes the weekly security on-call rotation might find it hard to reach out to SIG maintainers on Slack because:
1. The maintainer might use a strange/different name on Slack.
2. The maintainer might have several Slack accounts, and the TC member doesn't know which is the right one.
3. The same name might be used by many folks who are not maintainers, it could be really bad if the TC member involved the wrong person and caused security leak.

So in this PR I am trying to hammer out "how to map a GC / TC / Spec Approvers member to Slack". Once this is merged, we can take the same approach and cover all the SIG maintainers. The eventual goal is that we can automate the security incident/advisory flow on both GitHub and Slack.

Tagging @open-telemetry/technical-committee @open-telemetry/sig-security-approvers

### DIFF
--- a/people.yml
+++ b/people.yml
@@ -47,93 +47,121 @@ people:
   MrAlias:
     name: Tyler Yahn
     company: Splunk
+    slackMemberId: U01N9F13ERG
   Oberon00:
     name: Christian Neumüller
     company: Dynatrace
+    slackMemberId: U01NM39DZFB
   alolita:
     name: Alolita Sharma
     company: Apple
+    slackMemberId: U01J8KALC4F
     termEnd: October 2026
   arminru:
     name: Armin Ruech
     company: Dynatrace
+    slackMemberId: UM283PDT7
   austinlparker:
     name: Austin Parker
     company: Honeycomb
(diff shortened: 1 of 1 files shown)


mx-psi (approved):
This looks good to me. However, for moving forward with adding SIG maintainers I think we *need* a way to sync this file with Github teams, otherwise it becomes unworkable.

I think @trask looked into that at some point for the admin repo, I do not know how feasible it is (the Terraform integration seems to support it at least)

carlosalberto (approved):

trask (approved):

trask:
> I think @trask looked into that at some point for the admin repo, I do not know how feasible it is (the Terraform integration seems to support it at least)

https://github.com/open-telemetry/admin/issues/58

PR #3337, opened 2026-03-18
Data model and reporting for OpenTelemetry workstreams

jack-berg:
At the last GC/TC meeting, I volunteered to work on reporting to make it more clear what the TC is currently sponsoring to evaluate how new project proposals will impact attention. While starting to work on this, it became clear that we rely heavily on unstructured conventions. We have [sigs.yml](https://github.com/open-telemetry/community/blob/main/sigs.yml) encoding some information about SIGs, but missing key information about TC sponsorship, projects that we didn't classify as a SIG, and relationships between SIGs.

To execute on this, I found it necessarily build a data model to codify how we think about workstreams. 

## Key Pieces

The data model is defined in JSON schema at [workstreams.schema.yml](https://github.com/jack-berg/community/blob/workstream-data-model/scripts/schema/workstreams.schema.yml).

[workstreams.yml](https://github.com/jack-berg/community/blob/workstream-data-model/workstreams.yml) is a lightweight database representing the current state of our workstreams. It supersedes `sigs.yml`.

[validate-workstreams.py](https://github.com/jack-berg/community/blob/workstream-data-model/scripts/validate-workstreams.py) verifies the that `workstreams.yml` conforms to the schema, and enforces various logical validation criteria.

The SIG table from the [README.md](https://github.com/jack-berg/community/blob/workstream-data-model/README.md#special-interest-groups) is derived from `workstreams.yml`, as it was previously derived from `sigs.yml`. 

A new [workstreams.md](https://github.com/jack-berg/community/blob/workstream-data-model/workstreams.md) file is generated by [generate-workstream-report.py](https://github.com/jack-berg/community/blob/workstream-data-model/scripts/generate-workstream-report.py). For now, it includes a mermaid chart which visualizes the workstreams, their relationships, and other meta data, and a pivot table summarizing TC members and sponsorship. The data in this report is subject to gamification and misinterpretation, and there's a [preamble warning](https://github.com/jack-berg/community/blob/workstream-data-model/workstreams.md#workstream-report) which aims to address / discourage.

## About the data model

The data model attempts to encode information which is sometimes written down but unstructured, and other times tribal knowledge we talk about and rely on but never codified. Its mean to be a starting place I hope we iterate on over time.

* A workstream is a unit of work that we impose some process around and report on. Currently, there are two types of workstreams:
  * SIG: a permanent operational group with recurring meetings, a GC liaison, TC sponsor.
[...]

PR #3260, opened 2026-02-07
Project proposal: bootstrap Zig SIG

inge4pres:
Related to #2514 

Propose a new project to form a Zig Special Interest Group.

There is growing interest in Zig, and while there is a working SDK, it would be great to have it officially supported by the OpenTelemetry community.

PR #3411, opened 2026-05-04
[chore] Add initiatives to workstreams.yaml

mx-psi:
Another subset of #3337.

This one adds `initiative`, which is the rename I went with from this discussion https://github.com/open-telemetry/community/pull/3337#discussion_r2959382348, from the original PR.

The first commit tries to be as close as possible to the original content of #3337. The second commit


scripts/validate-workstreams.py
```diff
@@ -33,10 +33,12 @@
 
 KIND_REQUIRED_ROLES = {
     "sig":           {"gcLiaison", "tcSponsor"},
+    "initiative":    {"gcLiaison", "tcSponsor", "lead"},
```
jack-berg: I propose:

- that we define all initiatives as children of existing SIGs
- That the parent SIG is solely responsible for managing initiative lifecycle. I.e. accepting / rejecting initiatives, assigning leads, managing sponsorship and laying out their own sponsorship requirements. 
- Because initiative lifecycle is managed by parent SIG, there is no GC liaison or TC sponsor requirement. Those are inherited through the parent SIG.

danielgblanco: I agree with this. This was the intention of https://github.com/open-telemetry/community/blob/main/project-management.md#project-sponsorship

> If a project proposal is led by an existing SIG, TC sponsor and GC liaison for that SIG are assumed to be the sponsor and liaison for the new project. However, before approving the project, TC sponsorship level must be reviewed by the current TC sponsor. This ensures that the existing SIG is sufficiently supported if its scope is extended.

I remember we discussed this at some point between GC and TC. I can see how a new initiative may require the TC sponsorship level to change, so before it's approved the GC/TC should ensure that it has the right level of support.

If the initiative is to create a new SIG (and have some short-term deliverables for that SIG) then a new SIG is created with that initiative as the only child.

mx-psi: Alright, I modified it to satisfy those constraints. I am a bit skeptical about this since there may be initiatives that are cross-SIG, but maybe those should always be scoped to a cross-cutting SIG.

>If the initiative is to create a new SIG (and have some short-term deliverables for that SIG) then a new SIG is created with that initiative as the only child.

I did this for the Zig SIG, could you take a look and see if that's what you had in mind? It feels a bit awkward to have the Zig SDK entry with no information

workstreams.yml
```diff
@@ -1156,3 +1166,61 @@
...
+  - lead: dmathieu
+  resources:
+  - meeting:
+      schedule: Every other Thursday at 10:00 PT (alternating with general End-User SIG meetings)
+      gDocNotes: 1e-UNZA3Tuno9b53RQbe--whUcO0VIXF3P81oXsrBK6g
+  - slack:
+      name: '#otel-blueprints'
+      id: C0A844D6ZCH
+- id: zig-sig-bootstrap
+  kind: initiative
```
jack-berg, line 1191: This looks weird because it should actually be just a regular SIG, not an initiative, I believe

mx-psi: I think your comment is missing something :) 

I did it this way since the project is named "bootstrap Zig SIG", I could just put it as a SIG and remove this initiative, no strong opinion about it

jack-berg: I think the project is misnamed. I don't see anything in the Zig project proposal that differentiates it from other languages implementation SIGs: https://github.com/open-telemetry/community/pull/3260/changes

I think it should just be "Zig: SDK"

TC sponsor Is @jmacd 

If it is in fact a normal sig and not an initiative, then this problem goes away:

> It feels a bit awkward to have the Zig SDK entry with no information

danielgblanco: IIRC we ask folks to create a project/initiative proposal to create a new SIG, so GC can approve its creation and TC can agree level of sponsorship. Plus, it's a good way to advertise a new SIG creation and get the community to put their names down to contribute and be part of the SIG.

That's why the initiative is "Bootstrap Zig SIG". I've not had a look at the proposal but ideally it has a list of short-term deliverables. After those are done the SIG would move to BAU (i.e. what's "done" is the bootstrap, not the SIG).

I personally think it's helpful to have both an initiative and a new SIG with the bootstrap initiative when new SIGs are created. It gives the SIG an initial purpose.

workstreams.yml
```diff
@@ -1156,3 +1166,61 @@
...
+- id: otel-blueprints
+  kind: initiative
+  name: OTel Blueprints
+  parent: end-user
+  people:
+  - lead: danielgblanco
+  - lead: dmathieu
+  resources:
+  - meeting:
+      schedule: Every other Thursday at 10:00 PT (alternating with general End-User SIG meetings)
```
danielgblanco, line 1185: For an initiative, would meetings not be inherited from SIG meetings too? Just thinking in terms of presenting this information on the `community` README. Not sure what the plans are there.

workstreams.yml
danielgblanco: An example of this was the CI/CD SIG. The SIG was created with an initial set of deliverables. They completed them, moved on to a phase two but the rest of structures (e.g. GC liaison, TC sponsor, meetings, Slack, etc) remained the same. They just started a new initiative within that SIG.

README.md
```diff
@@ -174,6 +174,7 @@ SIG-specific GitHub discussions.
...
 | Rust: SDK&nbsp;<a id="sig-rust" href="#sig-rust"><sup>🔗</sup></a> | Alternating between Tuesday at 09:00 AM PT and Wednesday at 8:00 AM PT | [Google Doc](https://docs.google.com/document/d/12upOzNk8c3SFTjsL6IRohCWMgzLKoknSCOOdMakbWo4) | [#otel-rust](https://cloud-native.slack.com/archives/C03GDP0H023) | [calendar-rust](https://groups.google.com/a/opentelemetry.io/g/calendar-rust) | [Ted Young](https://github.com/tedsuo) |
 | Swift: SDK&nbsp;<a id="sig-swift" href="#sig-swift"><sup>🔗</sup></a> | Thursday at 09:00 PT | [Google Doc](https://docs.google.com/document/d/1LugL8r4bAkbTxZ1Gq_6j_8SDv1LuZaNeUdR-QLN4d48) | [#otel-swift](https://cloud-native.slack.com/archives/C01NCHR19SB) | [calendar-swift](https://groups.google.com/a/opentelemetry.io/g/calendar-swift) | [Alolita Sharma](https://github.com/alolita) |
+| Zig: SDK |  |  |  |  | [Alolita Sharma](https://github.com/alolita) |
```
danielgblanco, line 177: I guess here if we don't have meetings or notes created we could simply render it as TBD?

danielgblanco: Or have them as required fields in the schema, and set them as TBD in the workstreams file.

workstreams.yml
jack-berg: What's the distinction between Zig and Kotlin? Both are new language implementations, with goals of developing API / SDK and reaching 1.0 stable release, but Zig is an initiative and Kotlin a SIG. 

The Zig project's goals look an awful lot like BAU for a typical language implementation SIG: https://github.com/open-telemetry/community/blame/main/projects/zig-sig-bootstrap.md#L76-L89

danielgblanco: Kotlin should've had an initiative created to bootstrap the SIG, IMO. The intention of having these "bootstrap" initiatives was to create a set of initial deliverables that ensure the SIG is set up for success.

Let's say I'm an end-user and I see that there's new Kotlin SIG. How do I know what they're aiming to achieve in the next few months after creation? Without something like that I guess I'd have to join a meeting and ask?

jack-berg: This is squarely GC business so feel free to proceed however you see fit.

However, characterizing the early phase of a SIG as an initiative is in tension with [the idea that all initiatves have a parent SIG](https://github.com/open-telemetry/community/pull/3411#discussion_r3184718794) because then you need to decide which SIG is the parent for it.

Seems simpler to just call it a SIG, since that's what it will become, and be clear in the project document about what activities are going to be worked on in the early phase.

danielgblanco: I think we're probably referring to the same thing. What I'd propose is that, when a new SIG is proposed (as an initiative/project), a new SIG is created with a single initiative under it, which is the bootstrap one as approved by GC and TC.

This just avoids a new SIG being created with no initial initiative to work on.

PR #3445, opened 2026-05-17
Document specification deviation exceptions

### LINKED ISSUE #2097: How to handle explicit deviations to the spec?
During the discussion of https://github.com/open-telemetry/oteps/pull/232, a previous version of that proposal stated that language SIGs would only be considered stable once they implemented the stable parts of the API and SDK specs. The dotnet maintainers brought up the point that not everything in the spec makes sense for dotnet, and it shouldn't prevent the dotnet SDK/API from being called stable.

While the dotnet-specific parts are interesting, I think we have a more generic matter to discuss: if a SIG determines that a required part of the spec shouldn't be implemented, should the deliverable still be considered stable? Who should make this call? Should a TC waiver be provided?

henriquejsza:
Closes #2097.

## Summary

- Document that GC-approved specification exceptions may apply during stability review.
- Update the stabilization review template so requesters disclose approved deviations and user-facing documentation.

## Testing

- `npx cspell --config .cspell.yaml guides/contributor/processes.md .github/ISSUE_TEMPLATE/stabilization-review.md`
- `git diff --check`
- `docker run --rm --mount \"type=bind,source=$(pwd),target=/home/repo\" -e GITHUB_TOKEN lycheeverse/lychee:sha-8222559@sha256:6f49010cc46543af3b765f19d5319c0cdd4e8415d7596e1b401d5b4cec29c799 --config home/repo/.lychee.toml --root-dir /home/repo --verbose home/repo/guides/contributor/processes.md home/repo/.github/ISSUE_TEMPLATE/stabilization-review.md`

### DIFF
--- a/guides/contributor/processes.md
+++ b/guides/contributor/processes.md
@@ -21,6 +21,8 @@ help, and how changes are merged and released.
   - [GitHub workflow](#github-workflow)
   - [Open a Pull Request](#open-a-pull-request)
   - [Code Review](#code-review)
+  - [TC Stability Review](#tc-stability-review)
+    - [Specification deviations](#specification-deviations)
   - [Code attribution and licensing](#code-attribution-and-licensing)
     - [Copyright Notices](#copyright-notices)
   - [Specification Changes](#specification-changes)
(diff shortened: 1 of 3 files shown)


guides/contributor/processes.md
```diff
@@ -205,6 +207,24 @@ responsive to any requests for assistance and to issues opened. In other words,
 a TC stability review is a dialogue requiring engagement from both the TC and
 language SDK maintainers.
 
+### Specification deviations
```
cijothomas, line 210: I don't know if this is the right place to document this requirement. There is no requirement that a language implementation must request a TC review itself. And it feels a bit odd that TC is the one reviewing, but GC is the one approving the exemptions if any.

henriquejsza: Good point. The GC decision in #2097 says the policy should live in this repository, but placing it under TC stability review can imply that implementations must request TC review and blur GC ownership.

I propose moving the general exception policy to a standalone top-level section of `guides/contributor/processes.md`, outside the TC review section, while keeping the stabilization template limited to disclosing an already approved exception. Would that address your concern? Once the location is agreed, I'll also rebase and drop the now-obsolete community-link commit.

PR #3450, opened 2026-05-18
Add steps for promotion for triager and approver

bogdan-st:
Proposed fix for https://github.com/open-telemetry/sig-contributor-experience/issues/13
This file has gone a long way since the issue has been opened so some of it has been resolved organically. This is just formalising the process for promoting people to triager or approver, which inpractice has been pretty similar to the one for maintainers. 
This should add some clarity on what maintainers should expect from a nominated person in order to approve the promotion, and also for contributors to know when they should start a discussion in that direction. 

I ve made the decision (which is what i have observed and also follows other OSS projects) that for triagers and approvers we should only count maintainer votes. I'd be happy to change it to include approvers if anyone thinks it should be that way. 

Im still a bit confused about the approver-codeowner thing since from what i've seen in collector contrib at least they are different concepts, but that may be a different issue or just me understanding things wrong.

### DIFF
--- a/guides/contributor/membership.md
+++ b/guides/contributor/membership.md
@@ -125,14 +125,47 @@ elsewhere on GitHub.
 
 ### Requirements
 
-- Nominated by a maintainer, with no objections from other maintainers.
-- Consistently interact with the community and help with issues for at least 1 month.
+- Has been performing the role's responsibilities (see below), within the
+  access they have, for at least 1 month.
 
 ### Responsibilities and privileges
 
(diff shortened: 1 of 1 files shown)


atoulme (approved):

trask:
thanks @bogdan-st!

@open-telemetry/sig-contributor-experience-approvers would be great to get your reviews, thanks.

amy-super (approved):
Looks good to me. Thanks @bogdan-st !

maryliag (approved):

trask (approved):

PR #3452, opened 2026-05-19
Post-Graduation Roadmap for OpenTelemetry

tedsuo:
This meta-project is a re-organization of the work presented in the [Stable-by-default OTEP](https://github.com/open-telemetry/opentelemetry-specification/pull/4813/). Since almost none of the work required relates to changes in the spec, I'm moving it from an OTEP to the community repo, since this is where we do the rest of our project planning.

The goal of this PR is to identify the remaining workstreams that are needed to complete delivery of the initial set of goals for OpenTelemetry: a telemetry system that delivers tracing, metrics, and logs from the most common software libraries and cloud infrastructures.

The terms "stable by default" and "general availability" seemed to be a little confusing to some people. So we are trying a different term: "post-graduation roadmap." Third time is a charm.

This document is intended to be a high-level roadmap. If we agree that this scope of work is correct, we can merge this PR and move on to defining each individual workstream in more detail, as separate documents. Please evaluate this document from that perspective.


projects/otel-ga.md
```diff
@@ -0,0 +1,161 @@
+# OpenTelemetry GA: Completing our initial scope of work
+
+This project file identifies the remaining workstreams needed to complete delivery of the
+initial scope of work for OpenTelemetry: A telemetry system that delivers tracing,
```
cijothomas: The title "OpenTelemetry GA" is ambiguous and likely to confuse the community. Many language SDKs and instrumentations have been declared v1.0/stable for years. Do we meant OpenTelemetry as-a-whole here? Lets clarify to avoid this confusion.

projects/otel-ga.md
```diff
@@ -0,0 +1,161 @@
...
+
+## Current challenges
+
+### De facto stable
+
+Today, many components are "de facto" stable, meaning that they are versioned as 0.X and are
+technically still in beta, but are recommended to be used in production. This is confusing, as
+OpenTelemetry also has components marked 0.X that are genuinely experimental and should not be
+used in production. Additionally, some end user organizations have rules that prohibit them
+from deploying 0.X software to production.
```
cijothomas: lets also ack that there are many components that are declared 1.0 stable.

projects/otel-graduation.md
```diff
@@ -0,0 +1,161 @@
...
+Managed by Collector SIG, the OpenTelemetry Collector needs to complete its roadmap for v1.0.
+
+### Stability: SemConv Tooling for managing instrumentation
+
+In order to lower the cost of manage instrumentation at scale, and to better support native
+instrumentation efforts, we need tooling that improves correctness while reducing the cost of
+maintaining instrumentation.  The SemConv Tooling SIG is in charge of this project.
+
+* Weaver
+* AI coding
```
cijothomas, line 97: AI coding - bit vague - could we add a one-liner to clarify what we mean?

projects/otel-graduation.md
```diff
@@ -0,0 +1,161 @@
...
+* Test harnesses
+
+### Stability: Instrumentation
+
+The biggest barrier to general availability is unstable instrumentation.
+
+* Move away from the “community contrib” model for critical instrumentation packages.
+* Deploy the new SemConv tooling across all language ecosystems.
+* Badges and other forms of recognition
+* Native instrumentation push to move instrumentation out of OpenTelemetry.
```
cijothomas, line 108: Native instrumentation - do we mean libraries/frameworks natively picking a dependency on opentelemetry api, and isntrumenting themselves, without the need for us to make a instrumentation library package?

projects/otel-ga.md
```diff
@@ -0,0 +1,161 @@
...
+## Out of scope
+
+There is additional work that we see as necessary for the success of the project. However, in
+order to stay focused as a community, we must limit the number of large projects that we take
+on at any one time – we cannot ask maintainers to have five or six "number one priorities."
+
+The following projects are seen as important for the long term success of OpenTelemetry, but
+not actually necessary to deliver stable components that are deployable at scale.
+
+### Performance / Benchmarking
```
cijothomas: Performance is one of the key things we list in our [mission](https://opentelemetry.io/community/mission/#we-value-performance), so listing it as out-of-scope reads a bit oddly. If this is a bandwidth call rather than a position change, could we say so explicitly?

projects/otel-ga.md
```diff
@@ -0,0 +1,161 @@
...
+components, such as APIs and plugin architectures. Once we have reached this milestone, we
+can revisit our compatibility requirements and long term support guarantees.
+
+### Better designs and architecture
+
+The core specification for tracing, metrics, and logs have all been completed. However, we
+are always learning. It's reasonable that we may want to revisit these designs, either to
+incorporate new developments within the industry, or to address fundamental performance issues.
+
+Let's finish shipping v1.0 first, before distracting ourselves with v2.0.
```
cijothomas: similar to my comment in the beginning line - this also assumes OTel is not v1.0 yet.
https://opentelemetry.io/status/#language-apis--sdks paints a different picture. Lets clarify we meant OTel as whole, not pieces of it to avoid this confusion.

projects/otel-ga.md
cijothomas: @martincostello and I had discussed some work to help with this effort, hopefully we can help solve bandwidth problem to a certain extent.

projects/otel-ga.md
```diff
@@ -0,0 +1,161 @@
...
+instrumentation efforts, we need tooling that improves correctness while reducing the cost of
+maintaining instrumentation.  The SemConv Tooling SIG is in charge of this project.
+
+* Weaver
+* AI coding
+* Test harnesses
+
+### Stability: Instrumentation
+
+The biggest barrier to general availability is unstable instrumentation.
```
cijothomas: lets also list who owns this piece? I think we are moving to a model where the maintainers pick a set of key instrumentations and treat it like the core api/sdk? That was my understanding from the wording "move away from community contrib" model!
(💯 agree with the change, comment is about making the ownership clear)

projects/otel-ga.md
```diff
@@ -0,0 +1,170 @@
...
+NOTE: This is a meta-project. It describes a set of workstreams at a high level, so that we can
+agree upon the overall scope of work needed for OpenTelemetry to be considered GA or "generally
+available." For that reason, it is missing some sections that would normally be in a project
+file. In the future we should develop better road-mapping tools, but this is what we have today.
+
+## Background and description
+
+As part of the [due diligence](https://github.com/cncf/toc/blob/main/projects/open-telemetry/otel-graduation-dd.md)
+for OpenTelemetry's graduation, a scope of work was identified as required in order
+for OpenTelemetry to be considered "stable" or "generally available."
```
tigrannajaryan: How do we define "stable" for this project?

Does “stable” mean “doesn’t crash at runtime” or “doesn’t introduce breaking changes between releases”? Or is it both?

It would be nice to explicitly define what sort of stability goals we have.

projects/otel-ga.md
```diff
@@ -0,0 +1,170 @@
...
+* Stability: Instrumentation
+* Deployment: OpAMP v1.0
+* Deployment: Packaging v1.0
+* Deployment: Kubernetes Operator v1.0
+* Security
+* Roadmaps & Project Management
+
+### Stability: Collector v1.0
+
+Managed by Collector SIG, the OpenTelemetry Collector needs to complete its roadmap for v1.0.
```
TylerHelmuth: Many collector component 1.0 has historically be blocked by stable semantic conventions, so that no breaking changes to names occur once the component is tagged 1.0. I recall recently that there was discussion about making it easier to update semconv in 1.0 components. Is that still in effect for the GA concept this document describes? I would be good to include a link to that decision somewhere since it affects a lot of components (collector or instrumentation).

projects/otel-ga.md
```diff
@@ -0,0 +1,170 @@
...
+* Declarative configuration for managing instrumentation and stability
+
+### Deployment: Kubernetes Operator v1.0
+
+The Kubernetes Operator needs several features in order to make OpenTelemetry generally
+available on Kubernetes. Operator SIG is in charge of this project.
+
+* The need for pod attribution and other manual configuration requirements that interfere
+  with deploying OpenTelemetry at scale.
+* All major languages supported.
```
TylerHelmuth: Is this bullet specifically referring to auto-instrumentation? If so we should be explicit.

projects/otel-ga.md
```diff
@@ -0,0 +1,170 @@
...
+### Deployment: Packaging v1.0
+
+The Packaging SIG is in charge of this project.
+
+* Official packages for Debian and RHEL
+* Cross language definitions for distributions and versioning
+* Language distributions for SDKs, plugins, and instrumentation
+* Declarative configuration for managing instrumentation and stability
+
+### Deployment: Kubernetes Operator v1.0
```
TylerHelmuth: The Operator is the go-to k8s solution for many users, but many other users prefer to use the helm charts to install OpenTelemetry in Kubernetes. How do they fit into the GA picture?

projects/otel-ga.md
tedsuo: My intention is that stable means both. Once something is v1.0:
* it is ready for production, meaning it doesn't crash or cause harm.
* it is supported, meaning we will fix bugs and issue security patches without having them mixed in with breaking changes.

projects/otel-ga.md
tedsuo: Yes, we decided that we were being overly cautious by including data stability with v1.0. While it should be a breaking change if we changed the data, there's nothing wrong with breaking data changes being issued as a v2.0. Combining the two together concepts together as a requirement for v1.0 left us with no way of indicating that the code is safe to run in production.

But you're right, I don't think that this decision has been recorded in the spec. I believe this is the doc that needs to be updated: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/telemetry-stability.md

projects/otel-ga.md
tedsuo: It is, the intention is that all languages that have an auto-instrumentation mechanism can be installed via both the Operator and via system packaging.

projects/otel-ga.md
tedsuo: My understanding is that our current approach is for the OTel Helm Charts to just install the Operator. Am I off base in that assumption? Do we need a better solution than that?

projects/otel-ga.md
tedsuo: Clarify how? Can you suggest a new opening sentence?

projects/otel-graduation.md
tedsuo: Definitely need more details here, but I want the SemConv Tooling SIG to take a look at it before I put too many words in their mouth 🙂

projects/otel-graduation.md
tedsuo: Yes. There are two directions a community/contrib instrumentation package could gain more support. One is that a trusted group within OpenTelemetry maintains the package. The second is that the library itself includes the instrumentation natively, so there is no need for OpenTelemetry to maintain a separate package.

In both cases, we've identified a lack of tooling as a barrier. It's difficult to write instrumentation that matches the semantic conventions without making mistakes. So, once we have better tooling, we have an opportunity to try to upstream instrumentation. This would be preferable to maintaining a separate instrumentation package.

projects/otel-ga.md
TylerHelmuth:
```suggestion
* Auto-instrumentation supported for all major languages.
```

Do we have an existing definition for `major language` or is that distinction left up to the operator/packaging sig?

projects/otel-ga.md
tedsuo: Yes I'm only putting it as out of scope because I feel like we can't ask maintainers to focus on ten things at once. We want maintainers to focus on instrumentation, packaging, and declarative config because those things are necessary for OpenTelemetry to be stable and manageable at scale.

Improving performance is important, but it is not necessary for stability or deployability. So I want to say that at this time, improving performance is something optional that individual SIGs can work on, but not something we are trying to standardize across all implementations by producing a set of standard benchmarks or something like that.

If, in the same way that the SemConv tooling SIG has been workign on tools for managing instrumentation, a group could possibly be working on performance benchmarks they could propose to the community. In the past, this effort has always died because we have not found a useful way to define universal performance benchmarks, and SIGs have instead made better progress working on performance on a case-by-case basis, based on user complaints.

projects/otel-ga.md
TylerHelmuth: The `opentelemetry-operator` chart is one way to install the operator, and its 1.0 could be molded into this effort. But separately there is the `opentelemetry-collector` chart which installs the collector directly.

We have purposefully never given a stance like `you should use the opentelemetry operator to install a collector in kubernetes` because its not accurate to always suggest the operator.

TylerHelmuth: I think it would make sense to handle the helm charts via the respective efforts tho:

- `opentelemetry-operator` 1.0 with the Operator effort
- `opentelemetry-collector` 1.0 with the Collector effort

But I think it is dangerous to claim for GA that only the Operator needs 1.0, as that would be taking a strong stance that the Operator is the Official OpenTelemetry Way to install a collector on Kubernetes.

If the GC/TC wants to take that stance I think its worth discussing further.

projects/otel-ga.md
tedsuo: OTel as a whole is not "v1.0," that chart is too limited as it does not include the work listed in this document: instrumentation stability, packaging, deployment, and management at scale are still not completed.

While it is possible to install an SDK that is v1.0, you cannot reasonably instrument a real world application using only stable components. Today, if we shipped a "stable" distribution of OpenTelemetry in any language, that contained only components that are v1.0 or greater, it would contain no instrumentation and thus be completely useless in a real world scenario. A stable SDK simply isn't enough.

Furthermore, there's currently no way for an organization to deploy that stable distribution at scale, except for a couple of languages in a single environment: Kubernetes. These two hurdles – stability and deployability – are huge gaps, which is why they were flagged as part of the due diligence process done for OpenTelemetry's graduation

projects/otel-ga.md
tedsuo: I completely agree. Actually, figuring out who owns instrumentation is the biggest challenge in all of this work, in my opinion. Currently, no one owns it. And we have no one available to own it: it's unfair to assume that the current SDK maintainers can also take on the work of maintaining instrumentation, even with better tooling.

So, an important part of this workstream is figuring out a new model that offers some kind of reward for organizations that put in the work needed to maintain all of this instrumentation.

projects/otel-ga.md
tedsuo: We usually mean "Java, .NET, NodeJS, Python, Ruby, PHP" as those are the popular languages that also have an auto-installation path. Maybe there's a better name for it than "major languages" that could be seen as rude. Go could also be on the list if OBI-based auto-instrumentation gets to a good spot. I'm not sure what is required to other languages such as Erlang.

projects/otel-ga.md
```diff
@@ -0,0 +1,170 @@
+# OpenTelemetry GA: Completing our initial scope of work
+
+This project file identifies the remaining workstreams needed to complete delivery of the
+initial scope of work for OpenTelemetry: A telemetry system that delivers tracing,
+metrics, and logs from the most common software libraries and cloud infrastructures.
+
+NOTE: This is a meta-project. It describes a set of workstreams at a high level, so that we can
+agree upon the overall scope of work needed for OpenTelemetry to be considered GA or "generally
+available." For that reason, it is missing some sections that would normally be in a project
+file. In the future we should develop better road-mapping tools, but this is what we have today.
```
thompson-tomo:
```suggestion
> [!NOTE]
> This is a meta-project. It describes a set of workstreams at a high level, so that we can
> agree upon the overall scope of work needed for OpenTelemetry to be considered GA or "generally
> available." For that reason, it is missing some sections that would normally be in a project
> file. In the future we should develop better road-mapping tools, but this is what we have today.
```

projects/otel-graduation.md
cijothomas: I am not sure on framing tooling (or lack of it) as the main barrier. If I'm a library owner deciding whether to take a direct dependency on the OTel API, my decision tree looks roughly like this:

 1. API stability & long-term support — Is the OTel API in my language stable, and is there a guarantee of support for at least ~3 years? I can't take a dependency on something that might force me to react to breaking changes every year.
 2. Performance — Does depending on the OTel API regress my library on the **no-op path**? Some cost when the SDK is enabled is acceptable; cost when telemetry is disabled is not, because it directly slows down every user of my library, whether they use OTel or not.
 3. Semantic convention stability — Is the semantic convention I'm being asked to emit stable? If it churns, I either ship breaking changes downstream or carry compatibility layers/opt-in-out flags forever.
 4. Tooling / validation — Then — is there tooling to validate that what I produce matches what I'm supposed to produce?

The current wording treats (4) as the primary blocker, but (1)–(3) come first IMHO for any library owner. I also don't recall library authors citing tooling as their blocker in prior discussions — the concerns I've consistently heard are stability and performance. If the SemConv Tooling SIG has data showing otherwise, it would be great to link it. (I only have anecdotal evidence only, so happy to correct my position once I learn more)

For this workstream to actually deliver native instrumentation at scale, I think we need explicit commitments on (1)–(3) — API LTS guarantees, a no-op/hot-path performance commitment, and semconv stability — alongside the tooling work.

projects/otel-ga.md
cijothomas: > Actually, figuring out who owns instrumentation is the biggest challenge in all of this work, in my opinion. Currently, no one owns it. 

Sharing my observations:

In my opinion, it is fair to ask the language's core repo maintainers/approvers to find a list of libraries that they deem important for their ecosystem, and make themselves as owners for them. For years, OTel .NET maintained instrumentation libraries in the core repo itself. It actually helps with validating things e2e also.
I am currently doing it in OTel Rust by owning instrumentation library for the most important library ourselves (of course there is other owners as well)

Rewarding organizations are good idea, but we need a concrete plan.

projects/otel-ga.md
```diff
@@ -0,0 +1,170 @@
...
+a roadmap to GA that can be presented to the community.
+
+* Stability: Collector v1.0
+* Stability: SemConv Tooling for Instrumentation
+* Stability: Instrumentation
+* Deployment: OpAMP v1.0
+* Deployment: Packaging v1.0
+* Deployment: Kubernetes Operator v1.0
+* Security
+* Roadmaps & Project Management
```
cijothomas: Question: should OpenTelemetry's own self-observability be an explicit workstream in this GA plan?

If we're calling OTel GA — production-ready and deployable at scale — operators need to be able to answer basic questions like is my sdk/exporter dropping data? or is my collector silently failing to export? Today, in most languages and components, you can't easily tell:

 - The semantic conventions for OTel's self-instrumentation are still experimental.
 - Very few SDKs/components implement them end-to-end.
 - Silent data loss has come up repeatedly in issues, SIG discussions, and customer complaints
 - Users often discover missing telemetry days later with no signal from OTel itself.

To operate OTel at scale, we need the system to tell you when it's unhealthy. Is this intentionally out of scope, covered implicitly under another workstream (Collector v1.0? Instrumentation?), or is it a gap worth calling out explicitly? 

A concrete example I have always noticed:
OTel SDK's batch processor (default with OTLP exporter) *drops* telemetry when exporter cannot keep up. And there is no *standard* way for an operator to know about this. The fix is 
1. Have semantic convention for internal telemetry/self-instrumentation stable
2. Make sure all sdks implement it. (I think only Java and Go implements this. I opened a PR to add it to Rust recently.)

projects/otel-ga.md
cijothomas: Some wording suggestion:

This document identifies the remaining workstreams needed for the OpenTelemetry project as a whole to be considered generally available — i.e. an end-to-end platform where users can install, deploy, and operate tracing, metrics, and logs at scale using stable components.
 
 Note: many individual components (language APIs, SDKs, and a growing set of instrumentation libraries) are already at v1.0 today. "Project GA" here is a higher-level milestone about the platform as a whole, not a per-component status.

reyang:
I have a couple meta comments:

* I feel the use of "General Availability" is confusing and misleading. Take Kubernetes as an example, there is no such thing as "Kubernetes as a project has reached General Availability", the GA term is used for specific features https://kubernetes.io/search/?q=GA.
* There are many components which are already at v1.0 or even v2.0, if we put something like "this means finalizing the v1.0 roadmap for every component..." it would surprise many users.
* The term "GA" has been used several times:
  * https://opentelemetry.io/search/?q=GA#gsc.tab=0&gsc.q=GA&gsc.page=1
  * In the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/document-status.md#maturity-levels), we already said:
  
image

projects/otel-ga.md
mx-psi:
```suggestion
Managed by Collector SIG, the OpenTelemetry Collector needs to complete [its roadmap for v1.0](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/ga-roadmap.md). This includes marking core APIs as well as a minimal OTLP distro as 1.x.

The opentelemetry-collector-contrib repository contains over 200 hundred components. [Initial adopter interview findings](https://docs.google.com/document/d/1SQMdfYpCiBfpxtWDwASXVIl-PIzD9X4vdDPXYUphAF0/edit?tab=t.0) revealed that, although many of these are considered not 'core' and are instead community-supported, end-users rely on them. General availability therefore will include [the stability of additional priority components identified in 'phase 1'](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44130).

Marking further components or distros as 1.x is explicitly out of scope.
```

projects/otel-ga.md
tedsuo: Well, if the `opentelemetry-collector` helm chart only installs the Collector, that's pretty limiting, isn't it? I'm definitely not opposed to listing it. But I'd like to get a better understanding about whether or not we should be expanding our helm offerings beyond these two options.

projects/otel-ga.md
tedsuo: I think that's reasonable. OpenTelemetry definitely isn't finished if we are still missing critical forms of self-observability. Possibly related is OpAMP management and health reporting for the SDKs.

tedsuo:
@reyang I think you're correct about watching our language about v1.0 vs v2.0, I use v2.0 casually in one place and it isn't appropriate. I'll add some clarification.

As far as the term "General Availability" vs "Stable by Default" or something else... I don't know how much I want to bikeshed that 😅. But I will say, we have always used the term "General Availability" to mean exactly this: a component is stable, v1.0, ready for production and supported. "Stable" in the spec matches this meaning. The issue with most of the components in this roadmap is that they are "de facto" stable, meaning we tell people to run them in production but we haven't marked them as stable or issued a v1.0 for them.

So, given that no term is perfect, and given that we have always used GA in this manner, I'd like to stick with it. If people want to bikeshed and come up with a better term and get buy in from the community, maybe slack is a better place to do that. I promise to change it if it looks like there's consensus, but if you don't mind, I'd like to keep the comment threads here focused on the content.

projects/otel-ga.md
atoulme: Go has special meaning attached to https://go.dev/doc/modules/version-numbers where moving from 0 to 1 bears stability requirements, and it will take some alignment to move up past to v2, etc. 

> A module developer should increment this number past v1 only when necessary because the version upgrade represents significant disruption for developers whose code uses function in the upgraded module. This disruption includes backward-incompatible changes to the public API, as well as the need for developers using the module to update the package path wherever they import packages from the module. 

So the problem for the collector is we conflate GA and stability with API stability, which is a lot of small details.

projects/otel-ga.md
cijothomas: https://github.com/open-telemetry/opentelemetry-specification/pull/5118 Opened an OTEP to get cross-implementation performance tracked centrally. It shows prototypes as well.

projects/otel-ga.md
carlosalberto: IMHO this should be an _immediate_ followup to the Collector 1.0 stream (which includes completing self-observability). The reasons are 1) A lot can be learn and adopted from the Collector and 2) Making this part of the SDKs would have to involve all SIGs and potentially work with the Collector to align the telemetry output.

projects/otel-graduation.md
carlosalberto: > Native instrumentation push to move instrumentation out of OpenTelemetry.

What do you actually expect from this at this time? Creating documentation and (stable) tooling that gets us ready to **start** upstreaming instrumentation?

projects/otel-graduation.md
```diff
@@ -0,0 +1,174 @@
...
+The OpAMP SIG is in charge of this project.
+
+### Deployment: Packaging v1.0
+
+The Packaging SIG is in charge of this project.
+
+* Official packages for Debian and RHEL
+* Cross language definitions for distributions and versioning
+* Language distributions for SDKs, plugins, and instrumentation
+* Declarative configuration for managing instrumentation and stability
```
carlosalberto, line 132: Maybe @jack-berg has an opinion (or maybe not, but pinging him just in case).

projects/otel-graduation.md
```diff
@@ -0,0 +1,174 @@
...
+* All major languages supported.
+* Works with the same distributions and configuration options as developed by the package
+  management SIG, so that end users only need to learn a single set of configuration patterns.
+
+### Security
+
+While project-wide security protocols can always be improved. The following tasks have been
+identified as high priority. The Security SIG is in charge of this project.
+
+* Better staffing of the Security SIG.
```
carlosalberto, line 164: Any _initial_ idea on how to accomplish this? Some SIGs are (or feel?) understaffed and the only way to improve things is to get companies onboard.

tedsuo: Well, I think that security pulls from a different part of the organization than staffing other parts of OpenTelemetry. Right now, there is a heightened awareness around security, and many companies (end users, cloud providers, and vendors alike) are aware to the fact that OpenTelemetry is an important part of their technology stack. If OpenTelemetry is vulnerable, then they are vulnerable. 

So I think that companies could be convinced to contribute time from their security professional in two different ways: one-off projects that improve security for OpenTelemetry. These projects would need to have limited scope and clear benefits, but I think that it's reasonable to assume we could get them staffed if the value is high enough. For longer-term support, if we can set clear expectations for what the monthly time commitment is, the value is that these security professionals would have early access to CVE reports. I think that some companies with a serious commitments to OpenTelemetry would rest easier with that knowledge, enough so that the security investment would seem worth the cost of having to triage that backlog and work with other maintainers to ensure that any problems are dealt with quickly and effectively.

projects/otel-ga.md
tedsuo: I added self-observability as a workstream, I'm in complete agreement that this should be a priority.

projects/otel-ga.md
tedsuo: Hi Thompson. Generally, I don't think we use this alerting syntax, or github markdown extensions, though certainly I don't think we have a rule about it. But regardless, I ended up resolving this by re-wording the section to not look like a note; upon reflection, there's no need to treat this information like an aside, or to call it out for special treatment.

projects/otel-ga.md
tedsuo: @TylerHelmuth  I added a section to cover helm charts, with your suggestion.

projects/otel-ga.md
tedsuo: In Go, isn't there a difference between rules for module versioning vs the versioning for the overall service?

projects/otel-ga.md
tedsuo: Sounds good. Updated to include this new language. :)

projects/otel-ga.md
```diff
@@ -0,0 +1,218 @@
+# OpenTelemetry General Availability: Completing our initial scope of work
+
+This document identifies the remaining workstreams needed for the OpenTelemetry project as a whole
+to be considered generally available — i.e. an end-to-end platform where users can install, deploy,
+and operate tracing, metrics, and logs at scale using stable components.
+
+Please note that many individual components (language APIs, SDKs, and a growing set of instrumentation libraries)
+are already at v1.0 today. The term "General Availability" as used here is intended to represent a
+higher-level milestone about the platform as a whole, not a comment on the status or version number of
+any particular component.
```
tigrannajaryan: If we rename to avoid the term "GA" then this paragraph is probably no longer necessary.

projects/otel-graduation.md
```diff
@@ -0,0 +1,218 @@
...
+
+### De facto stable
+
+Today, many components are "de facto" stable, meaning that they are versioned as 0.X and are
+technically still in beta, but are recommended to be used in production. This is confusing, as
+OpenTelemetry also has components marked 0.X that are genuinely experimental and should not be
+used in production. Additionally, some end user organizations have rules that prohibit them
+from deploying 0.X software to production.
+
+As part of graduation, it was requested that we provide a mechanism for indicating to end
```
tigrannajaryan, line 35: I think it would be great if we can come up with a consistent labeling mechanism for Otel docs. 

Different SDKs and parts of OTel specs use different labels currently. OTLP in particular labels individual features in a way that I like (for example [here](https://github.com/open-telemetry/opentelemetry-proto/blob/c54a8ea010b7322ba6cc457adaabe5fa46c5c2c0/opentelemetry/proto/profiles/v1development/profiles.proto#L124)), similarly OpAMP (see example [here](https://github.com/open-telemetry/opamp-spec/blob/b1905ce23df84497cd55067fdce7b71a9696571f/specification.md?plain=1#L1599)). This approach is incidentally similar to how Kubernetes docs use stability labels (for example [here](https://kubernetes.io/docs/concepts/containers/runtime-class/)).

I would love to have this consistent `[stability-level]` labeling in all out docs.

projects/otel-graduation.md
```diff
@@ -0,0 +1,218 @@
...
+users which OpenTelemetry components are "production ready." Actually, we already have a
+mechanism for indicating this – the version number of the component. Going forward, we need to
```
tigrannajaryan, line 37: The version numbers typically apply to whole components (like the entire language SDK). However any actively evolving component often ends up having features at different stability levels. It would be useful to label individual features in addition to the version number of the component.

I also like how Kubernetes includes both the stability level of the feature and the version number that stability level was attained by the feature. See for example [this](https://kubernetes.io/docs/concepts/architecture/leases/#api-server-identity), saying `FEATURE STATE: Kubernetes v1.26 [beta](enabled by default)`.

tedsuo:
Several major updates to this PR:

* We're renaming the project to "Post-Graduation Roadmap." Let's see if it sticks. :)
* Added benchmarking, self-observability, and LTS as workstreams that are in scope.
* Generally tried to clean up the language for clarity.

Even though this is intended to be a high-level roadmap, some workstreams could still use more detail. If you see anything missing in the descriptions, please make suggestions.

projects/otel-graduation.md
```diff
@@ -0,0 +1,214 @@
...
+and alerting needed to respond to common problems. In other words, OpenTelemetry itself should
+be a model citizen when it comes to being an observable system. As part of encouraging native
+instrumentation, we want to implement our own best practices in order to encourage others to
+improve the observability of their systems.
+
+The Specification SIG is in charge of this project.
+
+### Redefine Long term support
+
+The primary goal of this initiative is to get all of the necessary components to v1.0.
```
tigrannajaryan, line 208: This probably need to be rewritten to match the new title.

projects/otel-graduation.md
```diff
@@ -0,0 +1,214 @@
...
+
+### Performance & Benchmarking
+
+To facilitate performance improvements, SDK maintainers have created a number of benchmarks for
+use within their individual SIGs. Based on a review of this work, we would like to create a
+set of standard benchmarks to be applied across all SDKs. This is both to help with performance
+improvements, and to help give end users a reasonable set of expectations when considering the use
+of our SDKs.
+
+The Specification SIG is in charge of this project.
```
cijothomas, line 192: nit: Link to https://github.com/open-telemetry/community/pull/3495 if that gets approved first.

projects/otel-graduation.md
```diff
@@ -0,0 +1,214 @@
...
+
+OpenTelemetry is itself a high-throughput distributed sub-system. This means it needs to emit
+sufficient telemetry about itself in order to be observable, along with a set of playbooks
+to explain how to interpret this telemetry, and advise operators on how to implement basic dashboards
+and alerting needed to respond to common problems. In other words, OpenTelemetry itself should
+be a model citizen when it comes to being an observable system. As part of encouraging native
+instrumentation, we want to implement our own best practices in order to encourage others to
+improve the observability of their systems.
+
+The Specification SIG is in charge of this project.
```
cijothomas, line 204: Nit: More like a join effort with SemanticConventions group, as the self-observability semantics are going to be defined in the semantic conventions.

projects/otel-graduation.md
```diff
@@ -0,0 +1,214 @@
...
+use within their individual SIGs. Based on a review of this work, we would like to create a
+set of standard benchmarks to be applied across all SDKs. This is both to help with performance
+improvements, and to help give end users a reasonable set of expectations when considering the use
+of our SDKs.
+
+The Specification SIG is in charge of this project.
+
+### Self-Observability
+
+OpenTelemetry is itself a high-throughput distributed sub-system. This means it needs to emit
```
cijothomas, line 196: this may require sub-sections covering collector, operator, language sdks.
If required to create a new sub-project for this, I'll be happy to bootstrap that after https://github.com/open-telemetry/opentelemetry-specification/pull/5087 is merged

projects/otel-graduation.md
```diff
@@ -0,0 +1,214 @@
+# Post-Graduation Roadmap: Completing our initial scope of work
```
cijothomas, line 1: 👍 this is very good! It also makes sense to do a blog after this PR is approved to let the world know what is OTel doing after graduation.

tigrannajaryan: +1 to the new name.

reyang:
During the TC discussion, our feedback is that we expect each area owner to review and agree/approve the proposal.

projects/otel-graduation.md
```diff
@@ -0,0 +1,214 @@
...
+
+The Governance Committee is in charge of this project, but everyone's input is needed.
+
+* More agency and responsibility for the roadmap in the hands of the maintainers.
+* Better ways to socialize the projects that need input.
+* Better ways to visualize the long term roadmap.
+
+### Performance & Benchmarking
+
+To facilitate performance improvements, SDK maintainers have created a number of benchmarks for
```
cijothomas, line 186: thanks for adding this. it was also part of our original OTel GA plan :)

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md#opentelemetry-ga

tigrannajaryan:
@tedsuo we were having a discussion in the TC about what it means for us to approve and merge this PR and it was not entirely clear to us. The following is my opinion.

- Approving this PR signals the agreement with the roadmap.
- Once merged @tedsuo with help of others will prepare and publish a roadmap at opentelemtry.io.
- The purpose of the published roadmap is to signal how we intend to evolve Otel post graduation.

To reiterate - in my opinion accepting and merging this PR means we agree to a roadmap. It does not necessarily mean we have identified the staffing for all items listed in the roadmap and immediately begin execution on all projects. That work will happen later, over time, with the published roadmap helping us to understand which future projects the GC and TC will support.

Is this also what you expect or you have a different expectation?

projects/otel-graduation.md
```diff
@@ -0,0 +1,214 @@
...
+end-users rely on them. v1.0 therefore will include
+[the stability of additional priority components identified in 'phase 1'](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44130).
+
+Marking further components or distros as 1.x is explicitly out of scope.
+
+### Tooling for Instrumentation
+
+In order to lower the cost of manage instrumentation at scale, and to better support native
+instrumentation efforts, we need tooling that improves correctness while reducing the cost of
+maintaining instrumentation.  The SemConv Tooling SIG is in charge of this project.
```
pellared, line 94: nit 😉 

```suggestion
maintaining instrumentation. The SemConv Tooling SIG is in charge of this project.
```

projects/otel-graduation.md
```diff
@@ -0,0 +1,214 @@
...
+* Test harnesses
+
+### Instrumentation Stability & Management
+
+Currently, the biggest barrier to stability is instrumentation. The Specification SIG is
+in charge of this project, but it requires significant input from maintainers and the community.
+
+* Move away from the “community contrib” model for critical instrumentation packages.
+* Deploy the new SemConv tooling across all language ecosystems.
+* Badges and other forms of recognition.
```
pellared, line 107: What about https://opentelemetry.io/ecosystem/registry/?

projects/otel-graduation.md
```diff
@@ -0,0 +1,214 @@
...
+* Weaver
+* AI coding
+* Test harnesses
+
+### Instrumentation Stability & Management
+
+Currently, the biggest barrier to stability is instrumentation. The Specification SIG is
+in charge of this project, but it requires significant input from maintainers and the community.
+
+* Move away from the “community contrib” model for critical instrumentation packages.
```
pellared, line 105: From my experience, I would be against moving away from the community/contrib model for "critical" instrumentation packages.

Instrumentation is not one-size-fits-all. Even with semantic conventions, there are still many valid choices an instrumentation package has to make: which spans, metrics, attributes, and events to emit; which features should be enabled by default; how much configuration should be exposed; how to handle high-cardinality data; and how much runtime overhead is acceptable. Different users often make different trade-offs here.

Because of that, I think it is healthy to allow multiple instrumentation packages for the same library or framework, as long as they follow semantic conventions and are discoverable. A single “official” package can easily become a bottleneck or end up optimizing for one set of users while being unsuitable for another.

For example, in OTel Go, `database/sql` is a good illustration of this. There is no single canonical instrumentation that fits every use case, and users choose between different packages depending on their requirements and preferences, such as:

- `github.com/XSAM/otelsql`
- `github.com/uptrace/opentelemetry-go-extra/otelsql`
- `github.com/signalfx/splunk-otel-go/instrumentation/database/sql/splunksql`

Instead of replacing the community/contrib model, I would prefer improving the quality bar around it: clearer stability signals, better discoverability, documented compatibility, semantic convention compliance, and guidance on which packages are recommended for which use cases. That would preserve the flexibility and innovation of the ecosystem while still helping users make safe choices for critical instrumentation.

This is the main reason I am not approving this PR as I am against this bullet 😉

projects/otel-graduation.md
```diff
@@ -0,0 +1,214 @@
...
+
+The remaining helm charts either need a roadmap for v1.0, or a warning that they are still experimental
+and not meant to be run in production. The Kubernetes SIG is in charge of this project.
+
+## Improvements
+
+### Security
+
+While project-wide security protocols can always be improved. The following tasks have been
+identified as high priority. The Security SIG is in charge of this project.
```
pellared, line 162: What about mitigating supply chain attacks? E.g. applying immutable releases, build attestations, artifact signing for all projects. Take https://github.com/open-telemetry/opentelemetry-java-instrumentation/releases/tag/v2.29.0 as an example that could be improved.

I think this is an area which is overlooked.

projects/otel-graduation.md
```diff
@@ -0,0 +1,214 @@
...
+
+* Better staffing of the Security SIG.
+* Triage that can keep up with AI-powered CVE reporting noise.
+* Oversight to ensure that security response protocols are being followed.
+* OTel scanning tools to counter noisy and inaccurate scanning tools used by end users.
```
pellared, line 167: Shouldn't we also do our own scanning? 

I think we should also add `THREAT_MODEL.md` to our repositores as it can not only act as documentation, but also can be respected by the security scanners.

References:
- https://github.com/OWASP/www-project-threat-modeling
- https://github.com/OWASP/www-project-lets-threat-model

pellared, line 167: Should we also recommend fuzzing? Currently we have some fuzz tests in OTel Collector and OTel Go

tigrannajaryan (approved):
@tedsuo confirmed we have the same understanding of this project in the spec call today. I am approving based on that.

projects/otel-graduation.md
dashpole: I read this as "move critical instrumentation from contrib into core". I don't _think_ this is proposing moving away from the community/contrib model generally.  @tedsuo maybe you can clarify what you had in mind here?

jack-berg (approved):

dashpole (approved):

carlosalberto:
@tedsuo We have already a few approvals from the TC, but we need to get some updates from your side (nothing polemical I think). Also some GC reviews would be great.

jmacd (approved):

projects/otel-graduation.md
tedsuo: I like the idea of following Kubernetes lead on this. Three possible values `[alpha, beta, stable]` plus listing the version number would work both for the spec and for implementations.

I'll add this to the doc. I should note, it's slightly different than the challenge this section is trying to address, namely that we have whole components that are listed as `v0.X` but are expected to be used in production.

projects/otel-graduation.md
tedsuo: Closing this comment to move the discussion here: https://github.com/open-telemetry/community/pull/3452/changes#r3342546984

projects/otel-graduation.md
tedsuo: So, there is a security challenge we are facing in OSS in general, and OpenTelemetry in particular. AI is creating the opportunity for supply chain attacks to be executed at scale. OpenTelemetry is a cross-cutting concern that is injected across an entire organization. This makes us a very attractive target.

Users want vendor-neutral telemetry system, meaning that they want us to provide everything they need. At the margins, users may choose to install different instrumentation packages, and of course OpenTelemetry will always provide a mechanism for allowing this. But that does not mean we can excuse instrumentation libraries from our standards, or leave them to the community as poorly vetted contrib components. SDKs and the Collector are both part of core, but they are also optional, users can provide their own SDK if they want, and they can use an alternative to the Collector, or no Collector at all if they so choose.

But, just as providing an SDK is a requirement for the project, providing instrumentation is also a requirement. The only way to use OpenTelemetry effectively is to install the instrumentation libraries that we provide. It is not realistic to ask large-scale organizations to go through our instrumentation offerings with a fine-toothed comb to determine which ones are safe to run and which ones are unacceptable – either from a quality standpoint, or a security/provenance standpoint.

So, that's the issue. Saying "instrumentation is something the community should manage, we don't make any guarantees about contrib from a security standpoint, and if you don't like the instrumentation we provide you should go write your own" is basically a non-starter for OpenTelemetry as a project. We need to ensure that we are providing enough instrumentation coverage to make OpenTelemetry useful for most organizations. The newfound security concerns make it even more critical that we have a way of managing this instrumentation without relying on a wide array of contrib maintainers.

pellared: For me what you describe feels for me to be in conflict with "Native instrumentation push to move instrumentation out of OpenTelemetry" and https://github.com/open-telemetry/community/pull/3452/changes#r3268803928:

> In both cases, we've identified a lack of tooling as a barrier. It's difficult to write instrumentation that matches the semantic conventions without making mistakes. So, once we have better tooling, we have an opportunity to try to upstream instrumentation. This would be preferable to maintaining a separate instrumentation package.

This is why my feedback stays:

> Instead of replacing the community/contrib model, I would prefer improving the quality bar around it: 

> it requires significant input from maintainers

This is my feedback as a maintainer 😉

Issue #3542, opened 2026-06-24
Proposal: OpenTelemetry User Groups


svrnm:
This is a follow up to today's GC meeting: while talking about the proposal #3430 we got into a conversation about the fact, that we have not outlined how groups of people that share an interest in the wider ecosystem and community of OpenTelemetry can meet and organize. This is not about yet-another-process, this is about giving guidance how those groups can self-organize, that we are happy to have them and that we can provide a certain level of support to them.

The proposal is to create a document (e.g. in contributor guidelines) that outlines best practices(!) for such a group, i.e.:

- have a slack channel `#otel-<name>` for your interest group (like #otel-industrial, #otel-collector-and-wasm, ...) and add the OpenTelemetry Admin account to it as channel manager. 
- have them listed on the website or community README as inofficial, "not blessed but welcomed" OpenTelemetry User Groups for discovery
- suggest them to have their own github org like `otel-<name>`
- suggest them to put their code if any under Apache-2.0 license to simplify potential later donations
- point them to the marketing and naming guidelines of OpenTelemetry and CNCF
- allow them to use OpenTelemetry Zoom Meetings if they want to meet on a regular basis
- recommend them to follow the CNCF Code Of Conduct (although this is mandatory for the channel+zoom)
- ...

I am happy to write down this guide, but wanted to collect feedback on this first. Again, this would not be a process or something we enforce, but a recommendation to communities that they can follow. 

These communites can be the source of future SIGs or project proposals.

cc @open-telemetry/governance-committee

trask:
I think this sounds like a nice way to expand the community, while keeping SIGs more formal.

I'd suggest not recommending them to create a github org, since that could look official (even if it isn't), and probably isn't needed in most cases where personal repo(s) would work fine and be simpler anyways for early stage work like this.

mx-psi:
I support this and agree on the github org comment. 

We could also have some template to include on repositories that makes it clear what the user group is and isn't to avoid any confusion

maryliag:
I support this proposal

svrnm:
Ok, I'll write out a doc and share it in a PR, thanks all!

alolita:
Great outcome from the GC discussions - look forward to the doc @svrnm

svrnm:
Thank you all! 

https://github.com/open-telemetry/community/pull/3564 is a draft proposal for this.

danielgblanco:
[...]

PR #3564, opened 2026-07-01
Add document for User Groups

svrnm:
Addresses #3542, this is an initial proposal for guidelines around "OpenTelemetry User Groups"


guides/contributor/user-groups.md
arminru, line 1: Don't forget to add it to the ToC here:
https://github.com/open-telemetry/community/tree/main/guides#opentelemetry-guidebooks

guides/contributor/user-groups.md
```diff
@@ -0,0 +1,167 @@
...
+  **"OpenTelemetry User Group for Industrial Applications"** over "OpenTelemetry
+  Industrial". The "User Group for …" construction makes clear that this is a
+  community using OpenTelemetry, which is exactly what it is. If you're unsure
+  whether a name is acceptable, ask the project, or reach out to
+  [trademarks@linuxfoundation.org](mailto:trademarks@linuxfoundation.org).
+- **Slack channel:** `#otel-<name>` (e.g. `#otel-telco`). Do not name it `#opentelemetry-<name>`
+- **GitHub:** if your group wants a place for code or documents, use a **personal
+  repository** — there's no need for (and we'd rather you not
+  create) a separate GitHub organization. You may prefix the repository name with
+  `otel-` (e.g. `otel-telco-examples`) to make it discoverable.
```
mx-psi, line 130: I'd add here some sort of "please add the disclaimer in the appendix" which can be something like:

```md
> [!NOTE]
>
> This is an [OpenTelemetry User Group](link to this doc). It is not an official body within the OpenTelemetry project. It is subject to the [CNCF Code of Conduct](link to CoC).
```

guides/contributor/user-groups.md
```diff
@@ -0,0 +1,167 @@
...
+### Spaces the project can offer
+
+OpenTelemetry lives in the **CNCF Slack**. If you're not there yet, sign up at
+[slack.cncf.io](https://slack.cncf.io/) and join the `#opentelemetry` channel —
+that's the project's main channel and the best place to start a conversation.
+
+For a User Group, the project can offer:
+
+- **A dedicated Slack channel** named `#otel-<name>` (e.g. `#otel-healthcare`),
+  with the OpenTelemetry Admin user as a channel manager. Ask in `#opentelemetry` and
```
mx-psi, line 82: We could also add a Slack automation for first-time joiners to the channel that says "this is part of an OpenTelemetry User Group"

guides/contributor/user-groups.md
```diff
@@ -0,0 +1,167 @@
...
+  `otel-` (e.g. `otel-telco-examples`) to make it discoverable.
+- **Licensing:** license any code or content you publish under **Apache-2.0**.
+  Beyond being permissive and well-understood, it keeps the door open for a clean
+  donation to the OpenTelemetry project later, should your group's work ever grow
+  into something the project wants to adopt.
+
+When in doubt about a name or logo usage, ask — it's much easier to adjust early
+than after a group has momentum.
+
+## Getting your group listed
```
mx-psi, line 139: Where would this list be?

svrnm: that's TBD, we can have it on opentelemetry.io or somewhere on the community README or both

guides/contributor/user-groups.md
svrnm: done!

guides/contributor/user-groups.md
```diff
@@ -0,0 +1,181 @@
...
+
+### Spaces the project can offer
+
+OpenTelemetry lives in the **CNCF Slack**. If you're not there yet, sign up at
+[slack.cncf.io](https://slack.cncf.io/) and join the `#opentelemetry` channel —
+that's the project's main channel and the best place to start a conversation.
+
+For a User Group, the project can offer:
+
+- **A dedicated Slack channel** named `#otel-<name>` (e.g. `#otel-healthcare`),
```
thompson-tomo: Would it make sense to prefix the channel with #otel_ug instead that way it is clearly a user group. Scenario is if we had a user group for wearables #otel_wearable if would look just like #otel_browsers with the later being a sig/project.

guides/contributor/user-groups.md
```diff
@@ -0,0 +1,181 @@
...
+  problems brings people together.
+
+If people are curious about a topic and want to explore it together, that's
+enough. There are no size or seriousness thresholds.
+
+User Groups are **not**:
+
+- **A replacement for existing project work.** Areas the project already
+  organizes around already have a home, i.e. the language implementations (.NET, Java, Go, …), the
+  Collector, semantic conventions, and so on. Use the
```
thompson-tomo: Would it make sense to tighten scoping of semantic conventions
```suggestion
  Collector, semantic convention domains, and so on. Use the
```

Reason is i see a user group as being the natural owner of refinements to a core domain definition. Ie the grpc user group would own the grpc refinement of the rpc conventions & any additional attributes specific to their refinement.

svrnm:
@open-telemetry/governance-committee any more feedback?

guides/contributor/user-groups.md
```diff
@@ -0,0 +1,181 @@
...
+That said, groups have a way of surfacing good ideas, and sometimes an idea
+wants a more official home. When that happens, a User Group can be fertile ground
+for it, and the usual project paths are open:
+
+- A group might put forward a **project proposal**, often in collaboration with
+  an existing SIG. See the
+  [project management guidelines](https://github.com/open-telemetry/community/blob/main/project-management.md).
+- A group might turn out to be **adjacent to an existing SIG** — for instance the
+  [End-User SIG](https://github.com/open-telemetry/community/blob/main/community-members.md) —
+  and the natural move is to bring the work there.
```
kamphaus, line 165: Should there be an additional path for a User Group to be come a dedicated SIG?
Ie. if it turns out that interest is strong and there are regular contributions to core OpenTelemetry repos, then it might be worth making the group a SIG to reduce friction.

svrnm: Yes, this is kind of hidden in the first point that they could put forward a project proposal.... the additonal sentence "in collaboration with" might make this less obvious.

What I tried to avoid stating explcitly that this is what's going to happen, most User Groups should never become a SIG

kamphaus: > most User Groups should never become a SIG

Agreed, I'm fine with not calling this out as an option.
In any case whenever this would occur the existing SIGs and maintainers can give guidance on the best path forward...

reese-lee:
Is this something we want to advertise (let people know about), or only point them to it if someone asks about such an option? At the moment, I don't feel strongly about either option; I'm mainly interested in what people think about this.

guides/contributor/user-groups.md
svrnm: Exactly, it should come naturally. The thing we don't want is people form a User Group, run it for 6-12 months and then expect to be "promoted" to a SIG. My preference would be that there are LOTS of user groups, and SIGs remain the special (...) case.

svrnm:
> Is this something we want to advertise (let people know about), or only point them to it if someone asks about such an option? At the moment, I don't feel strongly about either option; I'm mainly interested in what people think about this.

Personally I think we should advertise this, and let people know, that if they want to do things opentelemetry, here's some RECOMMENDATIONS by the project how to do it. I think especially the question around having a slack channel or spaces to meet are helpful for people to just go ahead and have #otel-whatever-they-want-to-talk-about

kamphaus (approved):

guides/contributor/user-groups.md
```diff
@@ -0,0 +1,181 @@
...
+OpenTelemetry is used across an enormous range of industries, domains, and
+environments — far more than the project's official structures can or should
+track directly.
+
+Making something *official* in OpenTelemetry, like forming a
+[SIG](https://github.com/open-telemetry/community/blob/main/community-members.md),
+starting a new project, or donating code under the
+[project management](https://github.com/open-telemetry/community/blob/main/project-management.md)
+and [donation](https://github.com/open-telemetry/community/blob/main/guides/contributor/donations.md)
+guidelines, is deliberately a heavy process with real requirements. That
```
danielgblanco:
```suggestion
guidelines, is deliberately a heavy process with well defined requirements. That
```

guides/contributor/user-groups.md
```diff
@@ -0,0 +1,181 @@
...
+[slack.cncf.io](https://slack.cncf.io/) and join the `#opentelemetry` channel —
+that's the project's main channel and the best place to start a conversation.
+
+For a User Group, the project can offer:
+
+- **A dedicated Slack channel** named `#otel-<name>` (e.g. `#otel-healthcare`),
+  with the OpenTelemetry Admin user as a channel manager. Ask in `#opentelemetry` and
+  an admin can create it.
+- **Access to the OpenTelemetry Zoom** for regular meetings, if your group wants
+  to meet synchronously.
```
danielgblanco: I think this is something that https://ocgroups.dev/ could offer as the recommended CNCF solution for community groups. Group organizers can create their own Zoom links and recording is available as option (I've not tried it but I presume recording will go to the hosts email to handle later, potentially guests too).

This would allow groups to manage their own recordings. There are certain limitations (e.g. they have to set a max of attendees, and it can only be repeated 12 times) but I think those are pretty minor.

Image

guides/contributor/user-groups.md
```diff
@@ -0,0 +1,181 @@
...
+
+That said, groups have a way of surfacing good ideas, and sometimes an idea
+wants a more official home. When that happens, a User Group can be fertile ground
+for it, and the usual project paths are open:
+
+- A group might put forward a **project proposal**, often in collaboration with
+  an existing SIG. See the
+  [project management guidelines](https://github.com/open-telemetry/community/blob/main/project-management.md).
+- A group might turn out to be **adjacent to an existing SIG** — for instance the
+  [End-User SIG](https://github.com/open-telemetry/community/blob/main/community-members.md) —
```
danielgblanco: Is this link intended to point to the community members page? I'm not sure I understand the relation.

guides/contributor/user-groups.md
reese-lee: well-defined* ?

guides/contributor/user-groups.md
reese-lee: Ooo I think this is a good consideration. I'm good with `#otel-ug-*`.

guides/contributor/user-groups.md
danielgblanco: Fair point, I never know when to use a hyphen and when not to be honest 😅

guides/contributor/user-groups.md
svrnm: good point!

guides/contributor/user-groups.md
svrnm:
```suggestion
guidelines, is deliberately a heavy process with well-defined requirements. That
```

guides/contributor/user-groups.md
svrnm: Oh wow, that's really cool! Let me add this in.

guides/contributor/user-groups.md
svrnm: no... that's the wrong link indeed. Would https://opentelemetry.io/community/end-user/ work?

guides/contributor/user-groups.md
svrnm: works for me

guides/contributor/user-groups.md
svrnm: To own the rpc conventions you need to go through the project propsal process and by that the group of people working on that are a sub-group of Semantic Conventions SIG and or their own SIG (t.b.h. I am not sure how we distinguish this today, but that's not part of this proposal).

So what would work is that a group of people comes together as a "RPC SemConv OTel User Group", they talk and make plans, and eventually they stand up a project proposal and say "hey, this is what we want ot do", from there _technically_ two entities exist, the one within the core project and the user group which may or may not do additional work as well.

The goal of this proposal is to create a recommendation not a process, so if we say "a user group owns" or "you must form a user group to do x", this adds overhead I rather want to avoid. I'd rather say, forming a user group before submitting a proposal is a route to success.

guides/contributor/user-groups.md
danielgblanco: Yes, I think that's the go-to resource :)

guides/contributor/user-groups.md
thompson-tomo: I am not talking about a user group owning the conventions defined for the rpc domain, but rather just the refinements/adjustments to the rpc conventions specific to grpc. The grpc conventions take the rpc conventions as a dependency.

Another example is db, the oracle db refinements may be owned by an oracle db user group which is what the case currently is.

Hence my change just makes it clearer that user groups are not to own the semantic convention domains.

guides/contributor/user-groups.md
danielgblanco: I cannot resolve the thread (no write permissions) but feel free to do so as per the change below.

guides/contributor/user-groups.md
danielgblanco: ✅

danielgblanco (approved):

guides/contributor/user-groups.md
svrnm: fixed!

guides/contributor/user-groups.md
svrnm: fixed!

mx-psi (approved):

tedsuo (approved):

maryliag (approved):

trask (approved):

PR #3569, opened 2026-07-02
initial policy project proposal

jaronoff97:
This is the proposal to establish a project around the recently merged [OTEP](https://github.com/open-telemetry/opentelemetry-specification/pull/4738). The goal is to create alpha components, a specification, and more implementations and recommendations.

### DIFF
--- a/projects/policies.md
+++ b/projects/policies.md
@@ -132,6 +132,7 @@ See [Project Staffing](/project-management.md#project-staffing)
 - **@jaronoff97** (Tero) – Maintainer
 - **@jsuereth** (Google) – Maintainer
 - **@dashpole** (Google) – Maintainer
+- **@menderico** (Google) – Maintainer
 - **@jackshirazi** (Elastic) – Maintainer
 
 #### Other Staffing
--- a/projects/policies.md
+++ b/projects/policies.md
@@ -93,7 +93,10 @@ foundation the rest of the ecosystem can build on.
 - **Fail-open:** an invalid policy becomes inert rather than breaking valid
   policies or dropping telemetry.
 - **Atomic and self-contained:** one matcher set, one action set, no references
-  to other policies, no ordering dependency.
+  to other policies, no ordering dependency. This means no policy depends on
+  another policy having already been applied — it does not mean policies
+  cannot conflict or overwrite each other's effects; precedence rules (see
+  Deliverables) govern that.
 - **Portable:** field references use the OpenTelemetry data model so a policy
   runs identically across conformant runtimes.
 
@@ -207,7 +210,7 @@ Relative to project start:
 The SIG will give regular (roughly monthly) updates to the Maintainer and
 Specification SIGs. After the project starts, specific target dates will be
 tracked via GitHub project updates (see
-[Project Lifecycle](project-management.md#project-lifecycle)).
+[Project Lifecycle](/project-management.md#project-lifecycle)).
 
 ## Labels (Optional)
 
@@ -227,6 +230,6 @@ here.
 
 Meeting times, Slack channel, and meeting notes will be posted on the
 [OpenTelemetry public calendar](https://github.com/open-telemetry/community#calendar)
-and added to the community SIG tables via [workstreams.yml](./workstreams.yml)
+and added to the community SIG tables via [workstreams.yml](/workstreams.yml)
 (running `make generate`), including the GitHub project ID as a `roadmapProject`
 entry so the project appears in the OpenTelemetry Roadmap.
--- a/projects/policies.md
+++ b/projects/policies.md
@@ -67,8 +67,9 @@ foundation the rest of the ecosystem can build on.
 **Goals:**
 
 - Deliver a small scoped set of stable policies — **sampling, filtering, and
-  transformation** — that work consistently across OpenTelemetry. Additional
-  policy types can be added iteratively once these are successful.
+  scoped transformations** — that work consistently across OpenTelemetry. More
+  transform types and additional policy types can be added iteratively once
+  these are successful.
 - Split and stabilize the specification into two documents: a **data model**
   specification (the policy schema, matchers, and actions) and an
   **enforcement** specification (how a component evaluates policies, including
@@ -118,9 +119,8 @@ that can be enforced anywhere in the pipeline.
   across logs, metrics, and traces (not a translation into existing Collector
   configuration), including fail-open handling and match tracking.
 - **Experimental policies** for the initial scope: **sampling, filtering, and
-  attribute sanitization**. These are the crisp deliverables that let us prove
-  the model quickly; transformation and other policy types are deferred and
-  added iteratively.
+  scoped transformations**. Additional transform types for things like
+  datapoints and other policy types are deferred and added iteratively.
 - **Applied-policy reporting.** Scope and formalize how a component reports
   which policies are actually in effect — for example an OpAMP status tagged by
   policy version/UUID, or self-observability signals for SDKs — so operators can
--- a/projects/policies.md
+++ b/projects/policies.md
@@ -10,16 +10,25 @@ telemetry components process data. A Policy is an atomic, self-contained rule
 that declares a single intent — "match specific telemetry and apply an action" —
 and can be evaluated identically regardless of where it runs.
 
-Because Policies are self-contained, independent, and can be evaluated in
-parallel and scaled to thousands of rules without degradation. The same policy
-behaves the same way whether it executes inside an SDK, a Collector, or any
-other specification-compliant component, and policies are designed to be defined
+Policies are self-contained, independent, and can be evaluated in parallel and
+scaled to thousands of rules without degradation. The same policy behaves the
+same way whether it executes inside an SDK, a Collector, or any other
+specification-compliant component, and policies are designed to be defined
 centrally and distributed to running systems.
 
 This project coordinates the initial work needed to move Policies from an
-accepted OTEP to an accepted specification: defining the specification, defining
-the canonical Protobuf schemas, and building an initial prototype in the
-Collector for validation.
+accepted OTEP into stable, usable policies people can rely on across
+OpenTelemetry. Concretely, that means defining the data model and enforcement
+specifications, defining the canonical Protobuf schemas, and shipping an **alpha
+OpenTelemetry Collector component** that natively interprets and enforces
+Policies. There are existing policy implementations already found in
+[java-contrib](https://github.com/open-telemetry/opentelemetry-java-contrib) and
+Tero's Policy work ([1](https://github.com/usetero/policy),
+[2](https://github.com/usetero/policy-go),
+[3](https://github.com/usetero/policy-zig)) that this project can build on as a
+second implementation. "Done" for this project is a small set of stable policies
+you can interact with; the components and features we need fall out from making
+those policies successful.
 
 ### Current challenges
 
@@ -57,17 +66,26 @@ foundation the rest of the ecosystem can build on.
 
 **Goals:**
 
-- Refine and stabilize the Policy specification so that it is general, minimal,
-  and implementable consistently across logs, metrics, and traces.
+- Deliver a small scoped set of stable policies — **sampling, filtering, and
+  transformation** — that work consistently across OpenTelemetry. Additional
+  policy types can be added iteratively once these are successful.
+- Split and stabilize the specification into two documents: a **data model**
+  specification (the policy schema, matchers, and actions) and an
+  **enforcement** specification (how a component evaluates policies, including
+  the two-stage match then keep & transform model and the core guarantees).
+  Enforcment should also describe and require a set of conformance tests to
+  ensure that implementations are consistent.
 - Define the canonical Protobuf schema for Policies, with a human-authorable
   YAML representation and bidirectional conversion (including shorthand forms).
-- Build an initial prototype in the OpenTelemetry Collector that evaluates
-  policies through the two-stage model (match, then keep & transform) and honors
-  the core guarantees: atomicity, fail-open behavior, idempotence, and
-  data-model-based field references for portability.
-- Establish match tracking (hits/misses) so operators can observe policy
-  effectiveness.
-- Validate that the same policy semantics can later be implemented in SDKs.
+- Ship an alpha OpenTelemetry Collector component that **natively interprets and
+  enforces** policies (rather than translating them into existing Collector
+  configuration), honoring the core guarantees: atomicity, fail-open behavior,
+  idempotence, and data-model-based field references for portability.
+- Use the host component's existing self-observability (e.g. Collector processor
+  self-obs metrics) to expose policy match tracking (hits/misses) so operators
+  can observe policy effectiveness, and scope out how a component reports
+  _which_ policies are applied (e.g. an OpAMP status tagged by policy
+  version/UUID, or self-observability signals for SDKs).
 
 **Requirements the model must preserve:**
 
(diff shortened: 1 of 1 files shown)


projects/policies.md
```diff
@@ -0,0 +1,200 @@
...
+
+With the recent merge of the
+[Policies OTEP](https://github.com/open-telemetry/opentelemetry-specification/pull/4738),
+there is significant interest in implementing Policies across the OpenTelemetry
+ecosystem. Policies introduce an intent-based specification for configuring how
+telemetry components process data. A Policy is an atomic, self-contained rule
+that declares a single intent — "match specific telemetry and apply an action" —
+and can be evaluated identically regardless of where it runs.
+
+Because Policies are self-contained, independent, and can be evaluated in
```
jackshirazi:
```suggestion
Policies are self-contained, independent, and can be evaluated in
```

projects/policies.md
```diff
@@ -0,0 +1,200 @@
...
+parallel and scaled to thousands of rules without degradation. The same policy
+behaves the same way whether it executes inside an SDK, a Collector, or any
+other specification-compliant component, and policies are designed to be defined
+centrally and distributed to running systems.
+
+This project coordinates the initial work needed to move Policies from an
+accepted OTEP to an accepted specification: defining the specification, defining
+the canonical Protobuf schemas, and building an initial prototype in the
```
dashpole: Maybe this is your intent, but I hope we have more than a prototype.  I think an alpha collector component is a reasonable goal to have.

projects/policies.md
```diff
@@ -0,0 +1,200 @@
...
+
+**Why now:** the OTEP has just merged and there is broad interest across SIGs.
+Coordinating the specification, schema, and a reference prototype now delivers
+immediately on the goals of the OTEP. Delivering this gives OpenTelemetry a
+common, governable way to express telemetry-processing intent that can be
+enforced anywhere in the pipeline.
+
+## Deliverables
+
+- **Policy specification refinements.** Iterate on the merged OTEP into
```
dashpole: To me, this sounds like two specification documents: A data model document, and a document specifying how policies are enforced.

projects/policies.md
```diff
@@ -0,0 +1,200 @@
...
+
+- Refine and stabilize the Policy specification so that it is general, minimal,
+  and implementable consistently across logs, metrics, and traces.
+- Define the canonical Protobuf schema for Policies, with a human-authorable
+  YAML representation and bidirectional conversion (including shorthand forms).
+- Build an initial prototype in the OpenTelemetry Collector that evaluates
+  policies through the two-stage model (match, then keep & transform) and honors
+  the core guarantees: atomicity, fail-open behavior, idempotence, and
+  data-model-based field references for portability.
+- Establish match tracking (hits/misses) so operators can observe policy
```
dashpole: It is an interesting question as to whether self-observability signals for policy enforcers should follow the conventions of the component they are embedded in (i.e. collector processor self-obs metrics and SDK self-obs metrics), or whether they should consistently use the same self-observability signals across components. But I would probably start with just using the component's existing self-obs for now.

projects/policies.md
```diff
@@ -0,0 +1,200 @@
...
+  transform), the matcher system (field selectors, match types, options,
+  `AttributePath` nested traversal), keep/sampling semantics for logs, metrics,
+  and traces, transform operations (remove, redact, rename, add), precedence
+  (most-restrictive-wins), and the error-handling taxonomy (parse, compilation,
+  runtime).
+- **Canonical Protobuf schema** for Policies, plus the YAML authoring format and
+  documented bidirectional conversion including shorthand forms.
+- **Collector prototype** implementing policy evaluation across logs, metrics,
+  and traces, including fail-open handling and match tracking.
+- **Example policies** demonstrating transformation, sampling, and attribute
```
dashpole: I think we should deliver **experimental** (not example) policies for at least sampling, filtering, and redaction. Transformation is a bit more complex, and i'm on the fence about including it right away. I think it is more important to focus on landing a few core policies quickly and making them successful across OTel.  Adding more policies is easy to do iteratively.

projects/policies.md
jsuereth: Agreed -

I think our scope should be alpha collector component initially.  We *also* have the java-contrib component that exists today we could include here if we pull in that group of folks.

If we take prinicples of otel specification today - that means we need these alpha components before we formalize a spec.  However I think "done" for this project is "a set of stable policies you can interact with".  The rest of what we need falls out from the features/components we need to build to be confident in those policies.

projects/policies.md
jsuereth: +1

projects/policies.md
jsuereth: I'd actually like to be a bit more aggressive since this i project management.

Let's pick specific policies we'd like to deliver so we have a very crisp/clear scope.

I'd like to commit to sampling, filtering and attribute sanitization - but we can discuss that *scope* here - this is the right place to do so.

Agreeing to initial scope should help us make rapid progress on prototypes and communicate in this proposal what we're really tackling

projects/policies.md
```diff
@@ -0,0 +1,200 @@
...
+### SIG
+
+A new Policies SIG will lead this work. Initial coordination and the reference
+prototype will happen in collaboration with the Collector SIG.
+
+### Required staffing
+
+See [Project Staffing](/project-management.md#project-staffing)
+
+#### Project Leads(s)
```
jsuereth, line 167: We should probably call out the areas of otel we contribute to to help here.

projects/policies.md
```diff
@@ -0,0 +1,200 @@
...
+
+- **@jaronoff97** (Tero) – Maintainer
+- **@jsuereth** (Google) – Maintainer
+- **@dashpole** (Google) – Maintainer
+- **@menderico** (Google) – Maintainer
+- **@jackshirazi** (Elastic) – Maintainer
```
jsuereth: Jack owns/contributed the policy work in java-contrib - we should call that out.

projects/policies.md
```diff
@@ -0,0 +1,200 @@
...
+
+## GitHub Project (Post-Approval)
+
+To be created from the
+[GitHub Project template](https://github.com/orgs/open-telemetry/projects/140)
+once the project is approved, pre-populated with issues covering the
+deliverables above and organized by the timeline milestones. Link to be added
+here.
+
+## SIG Meetings, Roadmap, and Other Info (Post-Approval)
```
jsuereth, line 240: We shoudl call out we plan to do regular (perhaps monthly?) updates in the Maintainer + Specification SIG.

projects/policies.md
```diff
@@ -0,0 +1,200 @@
...
+- **Policy specification refinements.** Iterate on the merged OTEP into
+  specification text covering the two-stage execution model (match; keep &
+  transform), the matcher system (field selectors, match types, options,
+  `AttributePath` nested traversal), keep/sampling semantics for logs, metrics,
+  and traces, transform operations (remove, redact, rename, add), precedence
+  (most-restrictive-wins), and the error-handling taxonomy (parse, compilation,
+  runtime).
+- **Canonical Protobuf schema** for Policies, plus the YAML authoring format and
+  documented bidirectional conversion including shorthand forms.
+- **Collector prototype** implementing policy evaluation across logs, metrics,
```
XuechunHou: Could you expand on policy evaluation? Are we translating Telemetry policies into an OTel YAML configuration before evaluating? Or are we extending the OTel collectors so they natively interpret Telemetry policies?

projects/policies.md
```diff
@@ -0,0 +1,200 @@
...
+  and traces, including fail-open handling and match tracking.
```
XuechunHou: How would users know what Telemetry policies are actually applied to an otel collector instance? Is that exposed by 'match tracking'?

jsuereth: That's something we need to build out / scope out.

Something we've been talking about is using, e.g. OpAMP status to report policy's applied by a version/UUID tag *or* self-o11y metrics for SDKs, e.g.  I expect us to have this as a work item for the project were we can formalize the mechanisms for reporting that.

jaronoff97:
@jsuereth @dashpole @jackshirazi I've updated based on your feedback. PTAL! :bow:

projects/policies.md
```diff
@@ -0,0 +1,232 @@
...
+  can observe policy effectiveness, and scope out how a component reports
+  _which_ policies are applied (e.g. an OpAMP status tagged by policy
+  version/UUID, or self-observability signals for SDKs).
+
+**Requirements the model must preserve:**
+
+- **Fail-open:** an invalid policy becomes inert rather than breaking valid
+  policies or dropping telemetry.
+- **Atomic and self-contained:** one matcher set, one action set, no references
+  to other policies, no ordering dependency.
```
jackshirazi: Somewhere we need to note that `no ordering dependency` means that any single policy is not dependent on any previous policy being applied, however policies can still conflict or overwrite each other, ie `no ordering dependency` does NOT mean no race conditions

projects/policies.md
```diff
@@ -0,0 +1,235 @@
...
+  taxonomy (parse, compilation, runtime).
+- **Canonical Protobuf schema** for Policies, plus the YAML authoring format and
+  documented bidirectional conversion including shorthand forms.
+- **Alpha Collector component** that natively interprets and enforces policies
+  across logs, metrics, and traces (not a translation into existing Collector
+  configuration), including fail-open handling and match tracking.
+- **Experimental policies** for the initial scope: **sampling, filtering, and
+  scoped transformations**. Additional transform types for things like
+  datapoints and other policy types are deferred and added iteratively.
+- **Applied-policy reporting.** Scope and formalize how a component reports
```
XuechunHou, line 135: How do we know which policies were applied by the collector if the collector process crashed before it could even report those metrics?

dashpole: General health and availability of the collector is already something that can be established by existing means (e.g. prometheus `up` metric, k8s liveness probes, or other forms of uptime monitoring). This will allow operators to diagnose issues with a policy that isn't applied (e.g. because it is invalid) or issues delivering policies, or will allow operators to tell how long it takes between when a policy is defined by the user, and when a policy is actually being enforced.

projects/policies.md
```diff
@@ -0,0 +1,235 @@
...
+way to express telemetry-processing intent.
+
+### Goals, objectives, and requirements
+
+The aim of this project is to deliver an initial, working implementation of
+Policies along with the specification and schemas that support it, proving the
+intent-based model end to end in at least one component (the Collector) as a
+foundation the rest of the ecosystem can build on.
+
+**Goals:**
```
dashpole, line 67: Just to confirm: We don't need any changes to OpAmp to be able to use it to deliver policies to components, right?

projects/policies.md
```diff
@@ -0,0 +1,235 @@
...
+- **Applied-policy reporting.** Scope and formalize how a component reports
+  which policies are actually in effect — for example an OpAMP status tagged by
+  policy version/UUID, or self-observability signals for SDKs — so operators can
+  see what is applied to a given instance.
+- **Integration direction** documenting how Policies relate to distribution
+  mechanisms (e.g. OpAMP, the OpenTelemetry Operator) and to policy composition
+  ("merger") — with dynamic distribution intentionally kept separate from core
+  policy semantics so both can iterate independently.
+
+Following OpenTelemetry specification practice, the alpha component(s) — the
```
dashpole, line 144: I think we should add an OBI prototype (not anything merged, just with maintainer feedback) on the list of deliverables here, just to provide a bit more diversity for the ways policies can be distributed and enforced.

projects/policies.md
```diff
@@ -0,0 +1,235 @@
...
+A new Policies SIG will lead this work. Initial coordination and the alpha
+Collector component will happen in collaboration with the Collector SIG.
+
+### Required staffing
+
+See [Project Staffing](/project-management.md#project-staffing)
+
+#### Project Leads(s)
+
+- **@jaronoff97** (Tero) – Maintainer; Operator, Helm, Injector
```
dashpole: There was some confusion when we reviewed this in the TC meeting. Consider removing "maintainer", since it could be read as "jaraonoff is a maintainer of the Injector", rather than "... is a maintainer of the policies project".

projects/policies.md
```diff
@@ -0,0 +1,235 @@
...
+
+- **Data model specification.** The policy schema itself: the matcher system
+  (field selectors, match types, options, `AttributePath` nested traversal),
+  actions, keep/sampling semantics for logs, metrics, and traces, and precedence
+  (most-restrictive-wins).
+- **Enforcement specification.** How a component evaluates policies: the
+  two-stage execution model (match; then keep & transform), the core guarantees
+  (atomicity, fail-open, idempotence, portability), and the error-handling
+  taxonomy (parse, compilation, runtime).
+- **Canonical Protobuf schema** for Policies, plus the YAML authoring format and
```
dashpole, line 119: IMO it would be good to decide where the schema will live in the project proposal (since it would mean additional work for the existing maintainers of the project this will live in).

I discussed this briefly with @jack-berg at the TC meeting, and the opentelemetry-configuration repo seemed like a possible place for it to live. It wouldn't be incorporated directly into the existing SDK configuration (since this is distinct by design), but it would be helpful if they adhered to similar design patterns (e.g. enabled vs disabled in config). Alternatively, we could make this its own repository.

projects/policies.md
jaronoff97: Nope, none required. there will be a few improvements that we may want to make on top of OpAMP (biggest would be minimal diff sets). No required changes though, just QOL improvements.

projects/policies.md
jaronoff97: Were we to do the opentelemetry-configuration repository, would we generate the protos from there? I lean towards putting it there, but would want to ensure that the SIG there would be comfortable with us making changes there for Policies. I think it would make sense to put it there given in the future we could conceivably set a list of policies in declarative configuration.

One of the things I have in my repositories is a list of conformance tests to ensure that implementations adhere to the specification. Were we to put the policy configuration in that repository, would we put the conformance tests there? Maybe we can chat about this on slack and bring the decision back here.

jack-berg: We're in solution territory, but opentelemetry-configuration defines its schemas using JSON schema, not proto. I would not be a proponent of defining proto message definitions in opentelemetry-configuration. IMO, protobuf shines for easy, compact binary encoding. Great for things like OTLP. JSON schema is great for describing schemas of YAML or JSON content. 

My intuition says that policies don't have high enough throughput or bandwidth considerations that would warrant protobof schema def. And the benefit of putting in opentelemetry-configuration in JSON schema is that it would allow any policy message definitions to be directly referencable by [opentelemetry-configuration types](https://opentelemetry.io/docs/specs/otel-config/types/). That will not be possible if they're defined using protobuf.

jaronoff97: One of the reasons I like proto beyond the binary encoding is the built in typing. This is helpful for cases like transferring large lists of policies over OpAMP where the native configuration data type is raw bytes. We've also already had conversations about bandwidth considerations for organizations transferring thousands of policies over the wire via OpAMP.

Either way, we will certainly have a JSON schema for a human-readable format and something referenceable by the declarative configuration types. I don't think having a proto definition and a json definition are mutually exclusive.

dashpole: Is there agreement that this should live within the opentelemetry-configuration repo?  I think thats all we need to decide right now.

dashpole: Actually, re-reading @jack-berg's comment, it sounds like you doesn't want protobufs in opentelemetry-configuration. Does potentially having both forms (proto and json schema) alleviate that concern? Or do you only think it is a good idea for them to live in opentelemetry-configuration if it only includes the json schema form?

jack-berg: I don't know what it would mean to have both JSON schema and protobuf representations. Seems like you should choose one and deal with the tradeoffs:

1. Protobuf: should probably live in opentelemetry-proto. Smaller / binary encoded representation available. Not directly referenceable by type in declarative config JSON schema types (i.e. references would have to be a string or some sort of encoded binary type with the expectation that the content is parsed to a policy protobuf message).
2. JSON schema: should probably live in opentelemetry-configuration. Larger / json encoded representation. Directly referenceable by declarative config types. 

My gut reaction to protobuf message definitions in opentelemetry-configuration is no it doesn't make sense. But make a proposal and we'll evaluate it.

jaronoff97: I think for now let's put it in opentelemetry-proto. Once we are happy with the proto definition we will also look into contributing a canonical JSON Schema which will live in the opentelemetry-configuration repository.

dashpole (approved):

projects/policies.md
```diff
@@ -0,0 +1,243 @@
...
+
+**Requirements the model must preserve:**
+
+- **Fail-open:** an invalid policy becomes inert rather than breaking valid
+  policies or dropping telemetry.
+- **Atomic and self-contained:** one matcher set, one action set, no references
+  to other policies, no ordering dependency. This means no policy depends on
+  another policy having already been applied — it does not mean policies cannot
+  conflict or overwrite each other's effects; precedence rules (see
+  Deliverables) govern that.
```
jackshirazi, line 99: One of the points I'm discussing in [this clarification](https://github.com/open-telemetry/opentelemetry-specification/issues/5218) points out that while policies shouldn't depend on other policies, they can depend on previous applications of the same policy. Specifically a policy that has been applied and then removed from the set of applied policies implies resetting the policy to it's default values. I'm not sure if this needs a specific callout in this doc

jsuereth (approved):

projects/policies.md
jaronoff97: yeah i think we'll go over this when we begin the work on the specification

dineshg13:
Thanks @jaronoff97 for driving policy work. I have been exploring similar concept within Datadog, i would be interested to contribute to this initiate.

jackshirazi (approved):

dashpole:
@open-telemetry/governance-committee this project needs a GC Liaison. Is anyone interested?

XuechunHou (approved):

mx-psi:
> @open-telemetry/governance-committee this project needs a GC Liaison. Is anyone interested?

@dashpole I'd be happy to be listed as the GC liaison (I am speaking here for myself, this does not mean the project is approved by the GC yet)

projects/policies.md
```diff
@@ -0,0 +1,245 @@
...
+
+- @dashpole (Google)
+
+#### Delegated TC Sponsor (Optional)
+
+N/A
+
+#### GC Liaison
+
+TBD
```
mx-psi:
```suggestion
- @mx-psi (Datadog)
```

trask (approved):

mx-psi (approved):

svrnm (approved):

Logiraptor:
@jaronoff97 I work on related components within Grafana Cloud, I'd be happy to contribute as well from our end if possible

jaronoff97:
@Logiraptor added you in, thank you very much :bow: mind ✅ on this PR when you get a chance?

Logiraptor (approved):

xrmx (approved):

tedsuo (approved):

dmitryax (approved):

mtwo (approved):

alolita (approved):
Lgtm.

tigrannajaryan (approved):

mx-psi:
This has a majority GC approval, and a TC sponsor. The approvals from Dmitrii and Tigran should satisfy the requirement from @jpkrohling here https://cloud-native.slack.com/archives/C01NJ7V1KRC/p1787760076044559, so I am merging!

PR #3582, opened 2026-07-09
chore: update link for langtrace and add exclusion for query

maryliag:
The existing link is returning 404 and causing error on CI. Update to a valid link.

GitHub Projects URLs with query params return 400 when accessed without a browser session


lmolkova (approved):
TY!

trask (approved):

PR #3584, opened 2026-07-13
Document FOSSA support contact

trask:
Document Robert Kielty as the first point of contact before filing FOSSA support requests.


trask:
Replaced with a fork-based draft because repository rules prevented updates to the same-repository branch.

Issue #3550, opened 2026-06-26
Migrate existing meetings to LFPCC


mx-psi:
Our current Zoom accounts are documented [here](https://github.com/open-telemetry/community/blob/main/assets.md#zoom-accounts).

We should also update Google Calendar events to reflect the new Zoom links.

## Table of recurring Zoom meetings from Calendar

> [!NOTE] 
> This is AI-generated and may have errors.

- [x] Android SIG (Zoom account `#1`)
- [x] Browser SIG (Zoom account `#4`)
- [x] C/C++ SIG (Zoom account `#1`)
- [x] CI/CD SemConv SIG (Zoom account `#1`)
- [x] Collector SIG (APAC/PT) (Zoom account `#3`)
- [x] Collector SIG (EU/ET) (Zoom account `#3`)
- [x] Collector SIG (NA) (Zoom account `#3`)
- [x] Communications SIG (Zoom account `#0`)
- [x] Community Demo App SIG (Zoom account `#2`)
- [x] Contributor Experience SIG (Zoom account unknown)
- [x] eBPF Instrumentation (Zoom account `#2`)
- [x] End-User SIG (APAC) (Zoom account `#1`)
- [x] End-User SIG: OTel Blueprints (Zoom account `#1`)
- [x] End-User SIG (Zoom account `#1`)
- [x] Entities SIG (Zoom account unknown)
- [x] es-ES Localization (SIG Communications) (Zoom account `#3`)
- [x] GC internal meeting (Zoom account unknown)
- [x] Go Compile Time Instrumentation SIG (Zoom account `#1`)
- [x] ja-JA Localization (SIG Communications) (Zoom account unknown)
- [x] JavaScript SIG (Zoom account unknown)
- [x] Kubernetes Operator SIG (Zoom account `#2`)
- [x] .NET SDK SIG (Zoom account `#2`)
- [x] OpenTelemetry PHP SIG (Zoom account `#0`)
- [x] OpenTelemetry Specification SIG + Maintainers Sync (Zoom account `#0`)
- [x] Project triage - EU-friendly (Zoom account unknown)
- [x] Prometheus WG (Zoom account unknown)
- [x] Python SIG (Zoom account `#4`)
- [x] pt-BR Localization (SIG Communications) (Zoom account unknown)
- [x] Ruby SIG (Zoom account `#2`)
- [x] Sampling SIG (Zoom account `#2`)
- [x] Semantic Convention Tooling (Zoom account `#3`)
[...]

PR #3610, opened 2026-07-31
Add LFPCC meeting links for FAAS WG, DevEx SIG, and Erlang/Elixir SIG

austinlparker:
## Description

Migrates meeting links from legacy Zoom to the Linux Foundation Project Control Center (LFPCC) for:

- **Functions as a Service (FAAS) WG** — added LFPCC meeting link
- **Developer Experience SIG** — added separate LFPCC links for NA (11:00 PT) and EU (17:00 UTC+8) meetings
- **Erlang/Elixir SDK SIG** — added LFPCC meeting link

Each entry adds a `Meeting link` field alongside the existing meeting time, notes, and invite group fields.

Part of the LFPCC migration tracked in #3550.


austinlparker:
huzzah.

svrnm (approved):
I think that's a great idea we should have that across all meetings, right?:)

scripts/update-sig-tables.py
arminru, line 1: Looks like you haven't pushed the actual changes to this file (just the chmod), thus the new properties don't make the way into the https://github.com/open-telemetry/community/blob/main/sigs.md generated from the YAML.

PR #3662, opened 2026-08-25
add saved replies templates

maryliag:
Provide guidance on common cases that need a reply.

Unfortunately, this is not possible to do org wise (https://github.com/orgs/community/discussions/4974), so providing the templates so people can add to their own account.

### DIFF
--- a/docs/replies/genai-policy-violation.md
+++ b/docs/replies/genai-policy-violation.md
@@ -1,0 +1,30 @@
+# Reply: Generative AI policy violation — closing PR
+
+**When to use:** A pull request appears to be generated primarily by an LLM
+without the contributor's own review, understanding, or engagement and you
+are closing it.
+
+Before using this reply, make sure you have read the contribution and are
+confident it violates the policy. If you are unsure, ask the contributor to
+clarify their process instead of closing.
+
+---
+
+## Reply body
+
+> Thanks for opening this pull request. Unfortunately, this contribution
+> appears to have been generated primarily by a generative AI tool without
+> sufficient human review or engagement, which is not in line with our
+> [Generative AI Contribution Policy](https://github.com/open-telemetry/community/blob/main/policies/genai.md).
+>
+> Maintainer time is limited, and we rely on contributors to be active,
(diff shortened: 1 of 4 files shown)


martincostello:
This is one of mine that might be useful to expand upon. Boilerplate for when an issue doesn't have any actionable details to investigate further.

> Needs repro
>
> Please provide a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) as a public GitHub repository that demonstrates the issue you're experiencing so we can look into this further.

trask (approved):

docs/replies/genai-policy-violation.md
```diff
@@ -0,0 +1,30 @@
...
+> appears to have been generated primarily by a generative AI tool without
+> sufficient human review or engagement, which is not in line with our
+> [Generative AI Contribution Policy](https://github.com/open-telemetry/community/blob/main/policies/genai.md).
+>
+> Maintainer time is limited, and we rely on contributors to be active,
+> willing participants in the contribution process, understanding the code
+> they submit, being able to discuss and iterate on it, and ensuring it fits
+> the current state of the project. LLM output alone does not meet that bar.
+>
+> I'm closing this PR for now. You are very welcome to come back with a
```
lmolkova, line 25: nit:

In semconv-genai repo we have quite a few AI-generated *issues* and quite extensive (but decipherable) discussions on them between what appears to be bots. Sometimes I want to close the whole issue based on the AI policy. I *think* the policy applies to issues too. 

It'd be nice to have a version of it tailored to issues, we can also wait and see if other repos have this problem.

lmolkova (approved):
Thanks a lot, this is really helpful!

docs/replies/genai-policy-violation.md
maryliag: That's a good point. I can't think how to phrase that for the reply, because the issue could be valid, but it would be closed because of the policy, so we might want to encourage for people to re-open but more succinct?
I'll go ahead with the merge of this PR and we can add more on top of that

